PR #82, opened 2018-07-09
Groupresults

Destidom:
Autograde is now able to show a result set of group submissions to a teacher, and the submission results for students belonging to a group.

### DIFF
--- a/public/src/components/teacher/GroupResults.tsx
+++ b/public/src/components/teacher/GroupResults.tsx
@@ -83,7 +83,8 @@ class GroupResults extends React.Component<IResultsProp, IResultsState> {
     }
 
     private getGroupResultSelector(group: IGroupCourseWithGroup): Array<string | JSX.Element> {
-        let selector: Array<string | JSX.Element> = [group.group.name, "5"];
+        const slipdayPlaceholder = "5";
+        let selector: Array<string | JSX.Element> = [group.group.name, slipdayPlaceholder];
         selector = selector.concat(group.course.assignments.filter((e, i) => e.assignment.isgrouplab).map((e, i) => 
         <a className="lab-result-cell"
             onClick={() => this.handleOnclick(e)}
--- a/public/src/components/teacher/GroupResults.tsx
+++ b/public/src/components/teacher/GroupResults.tsx
@@ -80,13 +80,14 @@ class GroupResults extends React.Component<IResultsProp, IResultsState> {
 
     private getResultHeader(): string[] {
         let headers: string[] = ["Name", "Slipdays"];
-        headers = headers.concat(this.props.labs.map((e) => e.name));
+        headers = headers.concat(this.props.labs.filter((e) => e.isgrouplab).map((e) => e.name));
         return headers;
     }
 
     private getGroupResultSelector(group: IGroupCourseWithGroup): Array<string | JSX.Element> {
         let selector: Array<string | JSX.Element> = [group.group.name, "5"];
-        selector = selector.concat(group.course.assignments.map((e, i) => <a className="lab-result-cell"
+        selector = selector.concat(group.course.assignments.filter((e, i) => e.assignment.isgrouplab).map((e, i) => 
+        <a className="lab-result-cell"
             onClick={() => this.handleOnclick(e)}
             href="#">
             {e.latest ? (e.latest.score + "%") : "N/A"}</a>));
--- a/public/src/components/teacher/GroupResults.tsx
+++ b/public/src/components/teacher/GroupResults.tsx
@@ -17,10 +17,11 @@ class GroupResults extends React.Component<IResultsProp, IResultsState> {
     constructor(props: IResultsProp) {
         super(props);
 
-        
-        if (this.props.groups[0] && this.props.groups[0].course.assignments[0]) {
+        let currentGroup = this.props.groups.length > 0 ? this.props.groups[0] : null;
+        if (currentGroup && currentGroup.course.assignments.length > 0) {
             this.state = {
-                assignment: this.props.groups[0].course.assignments[0],
+                // Only using the first group to fetch assignments.
+                assignment: currentGroup.course.assignments[0],
                 groups: this.props.groups,
                 
             };
@@ -34,10 +35,8 @@ class GroupResults extends React.Component<IResultsProp, IResultsState> {
 
     public render() {
         let groupLab: JSX.Element | null = null;
-        let tmp: JSX.Element | null = null;
-        
-        if (this.props.groups 
-            && this.props.groups.length > 0 
+        let currentGroup = this.props.groups.length > 0 ? this.props.groups : null;
+        if (currentGroup 
             && this.state.assignment
             && this.state.assignment.assignment.isgrouplab) 
             {
@@ -68,7 +67,6 @@ class GroupResults extends React.Component<IResultsProp, IResultsState> {
                             data={this.state.groups}
                             selector={(item: IGroupCourseWithGroup) => this.getGroupResultSelector(item)} 
                         />
-                        {tmp}
                     </div>
                     <div className="col-lg-6 col-md-6 col-sm-12">
                         {groupLab}
@@ -104,8 +102,7 @@ class GroupResults extends React.Component<IResultsProp, IResultsState> {
         query = query.toLowerCase();
         const filteredData: IGroupCourseWithGroup[] = [];
         this.props.groups.forEach((std) => {
-            if (std.group.name.toLowerCase().indexOf(query) !== -1
-            ) {
+            if (std.group.name.toLowerCase().indexOf(query) !== -1) {
                 filteredData.push(std);
             }
         });
--- a/public/src/components/teacher/GroupResults.tsx
+++ b/public/src/components/teacher/GroupResults.tsx
@@ -1,42 +1,47 @@
 import * as React from "react";
-import { IAssignment, ICourse, IStudentSubmission, IUser, IUserCourseWithUser } from "../../models";
+import { IAssignment, ICourse, IStudentSubmission, IUser, IUserCourseWithUser, IGroupCourseWithGroup } from "../../models";
 
 import { DynamicTable, Row, Search, StudentLab } from "../../components";
 
 interface IResultsProp {
     course: ICourse;
-    students: IUserCourseWithUser[];
+    groups: IGroupCourseWithGroup[];
     labs: IAssignment[];
     onApproveClick: (submissionID:number) => void;
 }
 interface IResultsState {
     assignment?: IStudentSubmission;
-    students: IUserCourseWithUser[];
+    groups: IGroupCourseWithGroup[];
 }
-class Results extends React.Component<IResultsProp, IResultsState> {
+class GroupResults extends React.Component<IResultsProp, IResultsState> {
     constructor(props: IResultsProp) {
         super(props);
 
-        if (this.props.students[0] && this.props.students[0].course.assignments[0]) {
+        
+        if (this.props.groups[0] && this.props.groups[0].course.assignments[0]) {
             this.state = {
-                assignment: this.props.students[0].course.assignments[0],
-                students: this.props.students,                
+                assignment: this.props.groups[0].course.assignments[0],
+                groups: this.props.groups,
+                
             };
         } else {
             this.state = {
                 assignment: undefined,
-                students: this.props.students,
+                groups: this.props.groups,
             };
         }
     }
 
     public render() {
-        let studentLab: JSX.Element | null = null;
-        if (this.props.students.length > 0 
-            && this.state.assignment 
-            && !this.state.assignment.assignment.isgrouplab
-        ){
-            studentLab = <StudentLab
+        let groupLab: JSX.Element | null = null;
+        let tmp: JSX.Element | null = null;
+        
+        if (this.props.groups 
+            && this.props.groups.length > 0 
+            && this.state.assignment
+            && this.state.assignment.assignment.isgrouplab) 
+            {
+                groupLab = <StudentLab
                 course={this.props.course}
                 assignment={this.state.assignment}
                 showApprove={true}
@@ -47,6 +52,7 @@ class Results extends React.Component<IResultsProp, IResultsState> {
                     }
                 }}
             />;
+
         }
 
         return (
@@ -55,16 +61,17 @@ class Results extends React.Component<IResultsProp, IResultsState> {
                 <Row>
                     <div className="col-lg6 col-md-6 col-sm-12">
                         <Search className="input-group"
-                            placeholder="Search for students"
+                            placeholder="Search for groups"
                             onChange={(query) => this.handleOnchange(query)}
                         />
                         <DynamicTable header={this.getResultHeader()}
-                            data={this.state.students}
-                            selector={(item: IUserCourseWithUser) => this.getResultSelector(item)}
+                            data={this.state.groups}
+                            selector={(item: IGroupCourseWithGroup) => this.getGroupResultSelector(item)} 
                         />
+                        {tmp}
                     </div>
                     <div className="col-lg-6 col-md-6 col-sm-12">
-                        {studentLab}
+                        {groupLab}
                     </div>
                 </Row>
             </div>
@@ -77,9 +84,9 @@ class Results extends React.Component<IResultsProp, IResultsState> {
         return headers;
     }
 
-    private getResultSelector(student: IUserCourseWithUser): Array<string | JSX.Element> {
-        let selector: Array<string | JSX.Element> = [student.user.name, "5"];
-        selector = selector.concat(student.course.assignments.map((e, i) => <a className="lab-result-cell"
+    private getGroupResultSelector(group: IGroupCourseWithGroup): Array<string | JSX.Element> {
+        let selector: Array<string | JSX.Element> = [group.group.name, "5"];
+        selector = selector.concat(group.course.assignments.map((e, i) => <a className="lab-result-cell"
             onClick={() => this.handleOnclick(e)}
             href="#">
             {e.latest ? (e.latest.score + "%") : "N/A"}</a>));
@@ -94,19 +101,18 @@ class Results extends React.Component<IResultsProp, IResultsState> {
 
     private handleOnchange(query: string): void {
         query = query.toLowerCase();
-        const filteredData: IUserCourseWithUser[] = [];
-        this.props.students.forEach((std) => {
-            if (std.user.name.toLowerCase().indexOf(query) !== -1
-                || std.user.email.toLowerCase().indexOf(query) !== -1
+        const filteredData: IGroupCourseWithGroup[] = [];
+        this.props.groups.forEach((std) => {
+            if (std.group.name.toLowerCase().indexOf(query) !== -1
             ) {
                 filteredData.push(std);
             }
         });
 
         this.setState({
-            students: filteredData,
+            groups: filteredData,
         });
     }
 
 }
-export { Results };
+export { GroupResults };
--- a/public/src/components/teacher/Results.tsx
+++ b/public/src/components/teacher/Results.tsx
@@ -16,10 +16,12 @@ interface IResultsState {
 class Results extends React.Component<IResultsProp, IResultsState> {
     constructor(props: IResultsProp) {
         super(props);
-
-        if (this.props.students[0] && this.props.students[0].course.assignments[0]) {
+        
+        let currentStudent = this.props.students.length > 0 ? this.props.students[0] : null;
+        if (currentStudent && currentStudent.course.assignments.length > 0 && currentStudent.course.assignments[0]) {
             this.state = {
-                assignment: this.props.students[0].course.assignments[0],
+                // Only using the first student to fetch assignments.
+                assignment: currentStudent.course.assignments[0],
                 students: this.props.students,                
             };
         } else {
@@ -32,7 +34,8 @@ class Results extends React.Component<IResultsProp, IResultsState> {
 
     public render() {
         let studentLab: JSX.Element | null = null;
-        if (this.props.students.length > 0 
+        let currentStudents = this.props.students.length > 0 ? this.props.students : null;
+        if (currentStudents
             && this.state.assignment 
             && !this.state.assignment.assignment.isgrouplab
         ){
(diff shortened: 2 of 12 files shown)


public/src/components/teacher/GroupResults.tsx
```diff
@@ -0,0 +1,119 @@
...
+            assignment: item,
+        });
+    }
+
+    private handleOnchange(query: string): void {
+        query = query.toLowerCase();
+        const filteredData: IGroupCourseWithGroup[] = [];
+        this.props.groups.forEach((std) => {
+            if (std.group.name.toLowerCase().indexOf(query) !== -1
+            ) {
```
meling: Formatting should follow standard `) {` on the line with the `if` statement, unless we need multiple conditionals.

public/src/components/teacher/GroupResults.tsx
```diff
@@ -0,0 +1,119 @@
...
+interface IResultsState {
+    assignment?: IStudentSubmission;
+    groups: IGroupCourseWithGroup[];
+}
+class GroupResults extends React.Component<IResultsProp, IResultsState> {
+    constructor(props: IResultsProp) {
+        super(props);
+
+        
+        if (this.props.groups[0] && this.props.groups[0].course.assignments[0]) {
```
meling: Are all these `0` indices safe? That is, could it happen that `groups[0]` is actually empty (`nil`)? I'm asking because elsewhere (e.g. line 40) you are checking `this.props.groups.length > 0`.

public/src/components/teacher/GroupResults.tsx
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    private getResultHeader(): string[] {
+        let headers: string[] = ["Name", "Slipdays"];
+        headers = headers.concat(this.props.labs.filter((e) => e.isgrouplab).map((e) => e.name));
+        return headers;
+    }
+
+    private getGroupResultSelector(group: IGroupCourseWithGroup): Array<string | JSX.Element> {
+        let selector: Array<string | JSX.Element> = [group.group.name, "5"];
```
meling: What does `5` mean?

public/src/components/teacher/Results.tsx
```diff
@@ -20,7 +20,7 @@ class Results extends React.Component<IResultsProp, IResultsState> {
         if (this.props.students[0] && this.props.students[0].course.assignments[0]) {
```
meling: Is it safe to use `0` here?

meling: Wouldn't it be better to replace common phrases such as `this.props.student[0]` with a new variable `student` or `stud` or `currentStudent` or `currStud`?

public/src/components/teacher/GroupResults.tsx
meling: Perhaps replace `this.props.groups[0]` with `currentGroup` or just `group` (following the same logic as for the proposed `student` variable in Results.tsx.)

public/src/components/teacher/GroupResults.tsx
Destidom: Ok, will fix this.

public/src/components/teacher/GroupResults.tsx
Destidom: As I have understood typescript, if there is not a question mark when declaring the variable it can not be null, since I'm not used to this I have an old habit of not trusting it. 
I will double check this, and implement the proposed changes.

public/src/components/teacher/GroupResults.tsx
Destidom: This is how slip days are posted to the user, right now they are hardcoded, both for Results and GroupResults, so need to implement some slip day logic for front-end and backend.

public/src/components/teacher/GroupResults.tsx
Destidom: You are correct, still need to check for length of array, since it is possible to pass down a empty array. So I will correct this.

public/src/components/teacher/GroupResults.tsx
meling: Ah, ok. Suggest to make it into a variable (or temporary constant) so that it is easier to recall, when we get time to implement the necessary logic.

public/src/managers/ServerProvider.ts
```diff
@@ -243,6 +243,67 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
         return JSON.parse(JSON.stringify(result.statusCode)) as IStatusCode;
     }
 
+    // getAllGroupLabInfos 
+    public async getAllGroupLabInfos(courseId: number, groupID: number): Promise<IMap<ISubmission>> { 
+        const result = await this.helper.get<ISubmission[]>(
+            ("courses/" + courseId.toString() + "/groups/" + groupID + "/submissions"),
```
meling: Could this type of string concatenation be replaced with [template strings](https://basarat.gitbooks.io/typescript/docs/template-strings.html).

Alternatively, we could use an `Array` and `join("/")` as shown [here](https://www.tutorialspoint.com/typescript/typescript_array_join.htm). But this approach may not be appropriate in this particular case, unless you move the concatenation before the `get<>` call. 

Another alternative that may be the best, would be create a URL joiner function that takes a variadic list of strings and does a `join("/")`. This function could then be reused everywhere that we want to concat a URL.

Another concern I have with these elements being strings is that it is quite easy to misspell one of these path elements, and so, perhaps it would be useful to create a set of constants that we import everywhere, and use as replacement for strings.

public/src/managers/ServerProvider.ts
```diff
@@ -243,6 +243,67 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
...
+        );
+        if (result.statusCode === 200 && result.data === undefined) {
```
meling: Shouldn't these http codes be predefined constants that are readable?

public/src/managers/ServerProvider.ts
```diff
@@ -243,6 +243,67 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
...
+            e.passedTests = 0;
+            if (e.testCases == null) e.testCases = [];
+            e.testCases.forEach(x => {
+                if (x.points !== x.score) {
+                    e.failedTests++;
+                } else {
+                    e.passedTests++;
+                }
+            })
+
```
meling: Unnecessary spaces.

public/src/managers/ServerProvider.ts
```diff
@@ -243,6 +243,67 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
...
+        );
+        if (result.statusCode === 200 && result.data === undefined) {
+            return {};
+        }
+        if (!result.data) {
+            this.handleError(result, "getAllLabInfos");
+            return {};
+        }
+        return mapify(result.data, (e) => {
+            let a = "{\"builddate\": \"2017-07-28\", \"buildid\": 1, \"buildlog\": \"This is cool\", \"execTime\": 1}";
```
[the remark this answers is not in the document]
meling: But then again, looking onwards in the code, I'm not quite sure why these `a` and `b` variables needs to be initialized like this. I think they could be just empty strings as well, since they will be overwritten in all cases.

public/src/managers/ServerProvider.ts
```diff
@@ -243,6 +243,67 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
...
+            let b = "[{\"name\": \"Test 1\", \"score\": 3, \"points\": 4, \"weight\": 100}]";
+            if ((e as any).buildinfo && ((e as any).buildinfo as string).trim().length > 2) {
```
meling: First, I would prefer if all variables were camel cased consistently: `buildInfo` and below `scoreObjects`. PS: Do you mean `scoreObject` here (without the `s`)??

Moreover, this code is not entirely obvious to me; I think it is extracting the `buildinfo` and `scoreobjects` fields from object`e`. I'm sure it is probably doing the right thing, but it looks like this is a pattern (repeated also below), and I wonder if this could be made into a function with a descriptive name, or if the type casting to `string` were done separately. The casting seems to be done three times per object. Alternatively, add a brief comment about what is going on here. But I'm also ok with keeping this code as is, if this is the "appropriate" typescript pattern for doing this.

public/src/managers/ServerProvider.ts
```diff
@@ -243,6 +243,67 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
...
+                a = (e as any).buildinfo as string;
+            }
+            if ((e as any).scoreobjects && ((e as any).scoreobjects as string).trim().length > 2) {
+                b = (e as any).scoreobjects;
+            }
+            let tempInfo: IBuildInfo;
```
meling: Replace `tempInfo` with `buildInfo`??

public/src/managers/ServerProvider.ts
```diff
@@ -243,6 +243,67 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
...
+            let scoreObj: ITestCases[];
+            try {
+                tempInfo = JSON.parse(a);
+            } catch (e) {
+                tempInfo = JSON.parse(
```
meling: `JSON.stringify()` with more appropriate information (failure info).
Perhaps this exception should also log a warning message or something. 
(The same applies to the `scoreObj` below.)

public/src/managers/ServerProvider.ts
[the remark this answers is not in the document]
meling: Also, since `a` and `b` are used several times below and somewhat far apart from this definition, they should probably have more descriptive names. General rule: If you are only using the variables very close to where they are defined, it is sometimes ok to use single letter variable names. The typical example is loop index variables such as `i` and `j`, as long as they are used close to where they are defined.

public/src/managers/ServerProvider.ts
```diff
@@ -243,6 +243,67 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
...
+            ("courses/" + courseId.toString() + "/groups/" + groupID + "/submissions"),
+        );
+        if (result.statusCode === 200 && result.data === undefined) {
+            return {};
+        }
+        if (!result.data) {
+            this.handleError(result, "getAllLabInfos");
+            return {};
+        }
+        return mapify(result.data, (e) => {
```
meling: The `e` loop variable should probably be given a more descriptive name, since it is being used further down in this return clause, and also since this `mapify()` (closure?) function does not describe its content.

public/src/managers/ServerProvider.ts
```diff
@@ -243,6 +243,67 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
...
+                    e.failedTests++;
+                } else {
+                    e.passedTests++;
+                }
+            })
+            
+            
+            
+            
+            return e.id;
```
meling: It is unclear what this is returning? Is `e.id` just a field of `e` or is it the key to the map that can be used to access the result data initialized above. Suggest to add a comment to explain what is being returned. (Also as part of the function documentation.)

public/src/managers/CourseManager.ts
```diff
@@ -309,6 +312,29 @@ export class CourseManager {
...
+     * @param group The group the information should be retrived from
+     * @param course The course the data should be loaded for
+     */
+    public async getGroupCourse(group: ICourseGroup, course: ICourse): Promise<IGroupCourse | null> {
+        // Fetching group enrollment status
+        const groupEnrollment = await this.courseProvider.getGroup(group.id)
+        
+        if (groupEnrollment != null && groupEnrollment.id === group.id) {
+            if (group.courseid === course.id) {
+                const returnTemp: IGroupCourse = {
```
meling: `returnTemp` -->  `groupCourse`

public/src/managers/CourseManager.ts
```diff
@@ -362,6 +388,27 @@ export class CourseManager {
...
+    /**
+     * Add IStudentSubmissions to an IUserCourse
+     * @param group The group
+     * @param groupCourse The group course
+     */
+    private async fillLinksGroup(group: ICourseGroup, groupCourse: IGroupCourse): Promise<void> {
+        if (!groupCourse.link) {
+            return;
+        }
+        const assigns = await this.getAssignments(groupCourse.course.id);
```
meling: `assigns` --> `assignments`

public/src/managers/CourseManager.ts
```diff
@@ -362,6 +388,27 @@ export class CourseManager {
...
+        if (assigns.length > 0) {
+            const submissions = MapHelper.toArray(
+                await this.courseProvider.getAllGroupLabInfos(groupCourse.course.id, group.id));
+
+            for (const a of assigns) {
+                const temp = submissions.find((sub) => sub.assignmentid === a.id);
```
meling: `temp` --> `submission` ?? (I think that is what it is.)

meling: `assignmentid` --> `assignmentID` (see other comment below about ID vs Id vs id)

(Not sure if `a.id` should be replaced with `a.ID`, but I think it should be ok in Typescript, since capitalization won't export the field, unless I'm mistaken.)

public/src/managers/CourseManager.ts
```diff
@@ -362,6 +388,27 @@ export class CourseManager {
         }
     }
 
+    /**
+     * Add IStudentSubmissions to an IUserCourse
```
meling: Documentation incorrect.

public/src/managers/CourseManager.ts
```diff
@@ -362,6 +388,27 @@ export class CourseManager {
...
+     * @param group The group
+     * @param groupCourse The group course
+     */
+    private async fillLinksGroup(group: ICourseGroup, groupCourse: IGroupCourse): Promise<void> {
```
meling: The first argument `group` is of type `ICourseGroup`. This isn't clear in the documentation and in the naming of the argument.

public/src/managers/CourseManager.ts
```diff
@@ -309,6 +312,29 @@ export class CourseManager {
         return await this.courseProvider.getCourseGroups(courseid);
     }
 
+    /**
+     * Load an IGroupCourse object for a single group and a single course
+     * @param group The group the information should be retrived from
+     * @param course The course the data should be loaded for
+     */
+    public async getGroupCourse(group: ICourseGroup, course: ICourse): Promise<IGroupCourse | null> {
```
meling: Naming of the first input argument: `ICourseGroup` is misleading (just `group`)?

Also, it is a bit confusing to have both `ICourseGroup` and `IGroupCourse`. I assume this is useful and there is a reason for it.

public/src/pages/StudentPage.tsx
```diff
@@ -208,28 +235,59 @@ export class StudentPage extends ViewPage {
...
                     CourseUserState.student,
                     CourseUserState.teacher,
                 ]);
-            this.activeCourses = this.onlyActiveCourses(this.courses);
+            this.activeUserCourses = this.onlyActiveCourses(this.userCourses as IUserCourse[]);
+
+            // preloading groupdata.
+            this.GroupUserCourses = [];
+
+            for(var i=0; i < this.activeUserCourses.length; i++) {
```
meling: Seems like a better for-loop would be:
```typescript
for (const course of activeUserCourses) {
...
}
```

public/src/pages/TeacherPage.tsx
```diff
@@ -154,6 +157,37 @@ export class TeacherPage extends ViewPage {
...
+                    labs={labs} 
+                    groups={linkedGroups}
+                    onApproveClick={async (submissionID:number) => {
+                        await this.courseMan.approveSubmission(submissionID);
+                        this.navMan.refresh();
+                    }}
+                    >
+                </GroupResults>;
+        }
+        return <div>404 Page not found</div>;
```
meling: Suggest to make a `PageNotFound` a global constant and use everywhere where such `<div>`s are being used. Or perhaps better make it a function that takes a specific error message as input and produces a debuggable error message. That is, care should be taken to avoid leaking sensitive information through error messages.

public/src/pages/TeacherPage.tsx
```diff
@@ -154,6 +157,37 @@ export class TeacherPage extends ViewPage {
         return <div>404 Page not found</div>;
     }
 
+    public async groupresults(info: INavInfo<{ course: string }>): View {
+        const courseId = parseInt(info.params.course, 10);
+        const course = await this.courseMan.getCourse(courseId);
+        if (course) {
```
meling: These three lines appear in all functions in `TeacherPage.tsx`, except `editGroup`, which has a slight variation so should be handled separately for now at least. We can simply make a closure `courseFunc` that executes the code inside the `if (course)` block on the provided `courseParam: string`.
```typescript
    private async courseFunc(courseParam: string, fn: (course: ICourse) => View): View {
        const courseId = parseInt(courseParam, 10);
        const course = await this.courseMan.getCourse(courseId);
        if (course) {
            return fn(course)
        }
        return <div>404 Page not found</div>;
    }
```
And you would call it like this:
```typescript
        return this.courseFunc(info.params.course, async (course) => {
            // HERE GOES THE DIFFERENT FUNCTIONS; REPLACE LINES BELOW.
            const groups = await this.courseMan.getCourseGroups(course.id);
            ...
            return <CourseGroup ...;
        })
```
Note that we now would need to replace `courseId` with `course.id`.

Reason to do this: It helps keep functions related to parsing ints from strings in one place. Later we could probably make this closure function generic, so that it could also work on other types, such as `IUser` and `IGroup` if necessary.

_Note that I'm not a big fan of all these `async` and `await` annotations. It is not at all clear to me if they are needed everywhere. My understanding is that they are useful when interacting with the backend, but it feels like much of this stuff does not. I have not investigated._

meling: Pushed an example to the branch.

public/src/managers/ServerProvider.ts
Destidom: Yes, I agree they should be defined by the framework that is talking to the backend. I'll fix this.

public/src/managers/CourseManager.ts
```diff
@@ -362,6 +388,27 @@ export class CourseManager {
...
+                await this.courseProvider.getAllGroupLabInfos(groupCourse.course.id, group.id));
+
+            for (const a of assigns) {
+                const temp = submissions.find((sub) => sub.assignmentid === a.id);
+                groupCourse.assignments.push({ assignment: a, latest: temp });
+            }
+        }
+    }
+
     public async approveSubmission(submissionid: number): Promise<void> {
```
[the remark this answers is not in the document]
Destidom: Ok, I will fix this, but much of the code has ID, id or Id throughout the code. I tried a quick refactoring, but something broke. I will have to spend some time renaming and making sure nothing breaks.
I also have to look into the "capitalization won't export the field" as you mentioned above.

public/src/managers/CourseManager.ts
Destidom: I agree that it is confusing (have a TODO for renaming that interface) but lack of a better name, and I saw that for group the interface is named ICourseGroup instead of just IGroup, which would be similar to the IUser naming.
I'm thinking of refactoring IGroupCourse to IGroup, and let ICourseGroup stay the same, which is the same naming as ICourseUser.

public/src/managers/CourseManager.ts
Destidom: Yes, this is the same problem as the one mentioned above, I think a refactor is necessary to avoid confusing and to be more clear.

public/src/pages/TeacherPage.tsx
Destidom: Yes, debuggable error messages or for production a more suitable error message would be preferable, I do not know how much time this will take to implement, so in the meantime I can make a global constant.

public/src/managers/ServerProvider.ts
Destidom: I'll look into it, and make it more understandable or document it better.

public/src/managers/ServerProvider.ts
Destidom: It seems like there is no standard definition for HTTP-codes in typescript/javascript, XHtmlRequester which is used has something called ReadyState, but this is not the same thing. 
I looked around and the norm seems to be either use a library which has these definitions or define them ourselves.
Do you have any suggestions on how to solve this?

public/src/managers/ServerProvider.ts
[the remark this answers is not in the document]
Destidom: I tested removing the initialized text for both strings a and b, seems like it would work.
The problem removing the text is if any of the two following if statements fail, the corresponding output would be empty when shown to the user.  But this is taken care of by the try/catch on the parsing, since parsing a empty string raises an error.

So i'll remove the initialized strings and rename them

public/src/managers/ServerProvider.ts
Destidom: I think it could be replaced with template strings, but we still need to differentiate the different placeholders.  Maybe end up with something like

```
public async getAllGroupLabInfos(courseId: number, groupID: number): Promise<IMap<ISubmission>> { 
    var  url = concatenate "courses/${courseID}/groups/${groupID}/submissions";
           const result = await this.helper.get<ISubmission[]>(
            (url),);
}
```

I like the idea of having a set of constants for the different endpoints on the backend. Using them with the string templates maybe?
Where should these constants be defined? Inside an own file for a better overview or inside the ServerProvider? Maybe they should be defined inside ServerProvider as this component can be changed out if the network layer changes.

Destidom: Then if a endpoint name changes, we only have to rename it once, instead of looking through many functions to rename them, and possibly miss some of them.

meling: LGTM. Although I would keep each endpoint in the "map" should also be given its own line.

Feel free to commit and push changes to the branch, and I'll review and do edits or suggest changes.

public/src/managers/ServerProvider.ts
meling: I'm surprised this isn't part of the std.lib or something. But it seems a lot of other people make their own thing, and I found [this one](https://gist.github.com/scokmen/f813c904ef79022e84ab2409574d1b45) which looks pretty reasonable.

public/src/managers/CourseManager.ts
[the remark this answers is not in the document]
meling: Let's be pragmatic for now, and rename things when we come across things. I'm ok with using `id` in the interface definitions where all the other fields are also lower case, at least for now (we can decide later). But we should make sure to use variable names like `courseID` and `userID` consistently.

PR #87, opened 2018-07-25
Privrepocreation

Destidom:
This branch makes it possible for the system to try and create private repositories and if it fails it will fall back to public repositories. A teacher may then after sorting payment click refresh course, and the system will loop through the repositories that are supposed to be private to become private.


Destidom:
Closed it while the pullrequest test failed.

PR #114, opened 2019-02-24
Remoteid and group fixes

meling:
This PR contains a bundle of commits to refactor poorly designed code and to squash a few bugs. 

The main thing that is fixed is the group creation bug due to teacher changing the group.
Also fixes some issues with handling remote ids. 
And now cmd/scm can be used with a personal access token, instead of retrieving such from the database.


meling:
Closed because already merged as part of PR #115.

PR #115, opened 2019-02-28
CI and hooks, assignment refresh and parser refactoring

meling:
See the last commit message for most details...


meling:
No, I believe this is unintended. I'm not quite sure what you mean by "owners without access". As I recall, the process should be:
- Students sign up for a group -- this should only add entries in the database
- A teacher approve a group -- this should create repo and team
- A teacher reject a group -- this should remove (or flag) the group in the database, but do nothing on a github.

0xf8f8ff:
Currently, attempts to approve or reject a group from the frontend are both handled by the _PatchGroup_ method of groups.go. In both cases (approval and rejection) _PatchGroup_ calls _createGroupRepoAndTeam_, which, in turn, calls an scm method _CreateRepoAndTeam_ with repository options where Private=true. _CreateRepoAndTeam_ method than attempts to create a private repo. 
If GitHub Organization plan for that course does not allow private repos, an error would be returned and no action taken. As a result, pending groups cannot be rejected (or approved) for courses based on GitHub Organizations with free plans.

0xf8f8ff:
This Pull Request does not fix some of the group updating issues as discussed above. Should it be necessary to fix it with Echo framework, we will provide a fix in the grpc implementation that can possibly be backported.

PR #121, opened 2019-05-24
Grpc web merge

0xf8f8ff:
Replacing Echo API with gRPC


web/grpcservice/autograder_service.go
```diff
@@ -118,6 +119,10 @@ func (s *AutograderService) GetCourses(ctx context.Context, in *pb.Void) (*pb.Co
 
 // GetCoursesWithEnrollment returns all courses for the user where user enrollment is of a certain type
 func (s *AutograderService) GetCoursesWithEnrollment(ctx context.Context, in *pb.RecordRequest) (*pb.Courses, error) {
+	if in.ID < 1 {
```
meling: Would it make sense to add a validation function for the `RecordRequest` type in validation.go?
Also note that there is benefit to using the `GetID()` (and similar for all other field accessor functions), since allows the `in` (or request) object itself to be `nil` and the function would still work; it would return `nil` or `0` depending on the return type.

meling: Commented before checking the next commit, which does exactly the first thing with the validation function... (so ignore that part of the comment...)

0xf8f8ff: Then it is probably safer to use getters instead of dot notation everywhere in the code?

meling: Yes, it is safer. But it is a tradeoff between safety vs readability. I mention it mainly so that you are aware that the methods does this `nil` check, so that, if you also find the need to do a `nil` check yourself, then that would be unnecessary... and you should use the provide accessor methods. (I'm not suggesting that we should change every instance of using the field variables directly... at least not yet.)

public/src/components/student/GroupInfo.tsx
```diff
@@ -1,22 +1,22 @@
...
-    group: ICourseGroup;
+    group: Group;
     course: ICourse;
 }
 class GroupInfo extends React.Component<IGroupPro, any> {
     public render() {
         const groupMembers: JSX.Element[] = [];
-        for (let i: number = 0; i < this.props.group.users.length; i++) {
-            groupMembers.push(<li key={i} className="list-group-item">{this.props.group.users[i].getName()}</li>);
+        for (let i: number = 0; i < this.props.group.getUsersList().length; i++) {
```
meling: Here I suggest to call `this.props.group.getUsersList()` before the for loop and assign it to a variable `users` and instead use `users.length` and `users[i].getName()` in the for loop.

0xf8f8ff:
> Consider if this commit should be rolled back; these methods are used by `cmd/scm`, but not elsewhere. Are they still useful in `cmd/scm`?

Yes, they are useful.

public/proto/ag_pb.js
```diff
@@ -602,7 +602,7 @@ proto.User.prototype.toObject = function(opt_includeInstance) {
  * @suppress {unusedLocalVariables} f is only used for nested messages
  */
 proto.User.toObject = function(includeInstance, msg) {
-  var obj = {
+  var f, obj = {
```
meling: @0xf8f8ff when running `make proto` on macOS, with:
```
% protoc --version
libprotoc 3.7.1
% tsc --version
Version 3.5.2
```
This file get these minor differences.

public/src/managers/ServerProvider.ts
```diff
@@ -84,9 +84,6 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
     public async getCourses(): Promise<Course[]> {
         const result = await this.grpcHelper.getCourses();
         if (result.statusCode !== 0 || !result.data) {
-            if (result.message) {
```
meling: Nice update to avoid these. I notice that there are some variations in that some methods check only the status code and others also check if there is `result.data`. It is not clear to me if these variations are required or if it is a bug. If it is a bug, I suggest to add a check function `hasResult` that returns a boolean that can be used everywhere. Maybe the result objects themselves can have such a function. That is, only if the same check function can be used for all result types, as otherwise we would just move the problem elsewhere.

On the other hand, I wonder if the return of empty list early is actually needed, since the iteration over the result list would be empty and so unless `result` is `null`, the `map` and other code may just return the empty list also, making the check unnecessary. (This would be a nice instance of defining errors out of existence ;-)

public/src/managers/GRPCManager.ts
```diff
@@ -287,4 +290,10 @@ export class GrpcManager {
         });
         return grpcPromise;
     }
+    // if gRPC call returns with error, inform user (currently just logs to console)
+    private informUser(response: IGrpcResponse<any>, sender: string): void {
```
meling: I wonder if by use of reflection, you could find the calling method (`sender`) and thereby avoiding to pass the `sender` parameter. It would make it easier to use...

public/src/managers/ServerProvider.ts
0xf8f8ff: Not a bug. Some grpc methods are only supposed to return errors/status codes. In such methods `result.data` will be `null` or of type Void (kinda `null` but sugarcoated for protobuf needs) and we only care about status code (thats where error notification for user is coming from).

public/src/managers/GRPCManager.ts
0xf8f8ff: Done.

meling: Nice

0xf8f8ff:
I guess so. There is no use for separate scripts now that everything is in Makefile.

web/access_control.go
```diff
@@ -108,3 +109,16 @@ func (s *AutograderService) getUserAndSCM2(ctx context.Context, courseID uint64,
...
 	return s.getUserAndSCM(ctx, crs.GetProvider(), mustBeAdmin)
 }
+
+// teacherScopes defines scopes that must be enabled for a teacher token to be valid.
+var teacherScopes = []string{"admin:org", "delete_repo", "repo", "user"}
+
+// hasTeacherScopes checks whether current user has upgraded scopes on provided scm client.
+func (s *AutograderService) hasTeacherScopes(ctx context.Context, sc scm.SCM) bool {
+	auth := sc.GetUserScopes(ctx)
+	if !cmp.Equal(auth.Scopes, teacherScopes) {
```
meling: @0xf8f8ff This `cmp.Equal` I suppose is sensitive to the specific order of the entries coming from github and so, if the order changes in the future, this may fail, or?

web/autograder_service.go
```diff
@@ -117,9 +117,9 @@ func (s *AutograderService) IsAuthorizedTeacher(ctx context.Context, in *pb.Void
 	if err != nil {
```
meling: @0xf8f8ff please check this question about admin? I forget the rules for teacher vs admin. It's on the readme, but don't have time to check. But you may have a different intent with this functionality (I think I rewrote it using `getUserAndSCM()` following directly previous semantics.

web/access_control.go
0xf8f8ff: Yes, it is order sensitive. I'll add sort before compare, probably as a separate function.

web/autograder_service.go
0xf8f8ff: Yes, user must be admin. Updated boolean in method call.

0xf8f8ff: About passing course ID: there is no course connection here, scopes are granted by github on the user token issued by Autograder. It happens once per user (unless we revoke all user tokens) and has nothing to do with a specific course.

web/organizations.go
```diff
@@ -30,7 +30,6 @@ func (s *AutograderService) getAvailableOrganizations(ctx context.Context, sc sc
 		course, err := s.db.GetCourseByOrganizationID(org.ID)
 		if err != nil {
 			s.logger.Errorf("couldn't fetch course from database: %v", err)
-			continue
```
meling: Sorry about this. It is probably because the error from the database is `gorm.ErrRecordNotFound`, which is what we want. So we could instead check for this error type and only continue if `err` is different from record not found. (We should only log an error if a different error message than record not found is discovered.)

meling:
> Ok, the `Enrolled` field is actually being set by a gorm method, was a bit confusing as the value never comes to the database. Then the only thing we lose is enrollment ID, which does not seem to be used in UI anyway. Returning the old method.

Okay sounds good. But bear in mind that I may not have the complete picture of how this is being used on the frontend. I just thought it was a bit worrying to do multiple RPCs like that...

web/hooks.go
```diff
@@ -171,12 +172,13 @@ func runTests(logger *zap.Logger, db database.Database, runner ci.Runner, repo *
 		logger.Error("Failed to extract results from log", zap.Error(err))
 		return
 	}
+	log.Println("hooks: RunTests got result from CI with score: ", result.TotalScore())
```
meling: If these log statements are meant to be useful in production, we can use `logger.Debug()` as used elsewhere in this function.

ci/results.go
```diff
@@ -34,6 +34,7 @@ func ExtractResult(out, secret string, execTime time.Duration) (*Result, error)
 		// check if line has expected JSON score string
 		if score.HasPrefix(line) {
 			sc, err := score.Parse(line, secret)
+			log.Println("ci got score parsing ci results, line: ", sc)
```
meling: It seems `ExtractResult()` is only used from `web/hooks.go:runTests()` which also has access to the logger, and so we can also pass that logger to `ExtractResults()` as the first argument. Thus, we can replace these `log.Println` statements with appropriate `logger.Debug` statements.

web/hooks.go
0xf8f8ff: No, they are absolutely not meant for production. I will remove them when I am ready with understanding what is wrong with submission score parsing. Hopefully tomorrow.

scm/github.go
```diff
@@ -111,10 +111,40 @@ func (s *GithubSCM) CreateRepository(ctx context.Context, opt *CreateRepositoryO
...
+	var err error
+	// if ID is set, get by ID
+	if opt.ID > 0 {
+		gitRepo, _, err = s.client.Repositories.GetByID(ctx, opt.ID)
+	} else {
+		// otherwise get by repo name and owner (usually owner = organization name)
+		gitRepo, _, err = s.client.Repositories.Get(ctx, opt.Owner, opt.Path)
+	}
+	if err != nil {
+		return nil, err
```
meling: Shouldn't we log err here?

0xf8f8ff: I tried to move most of the logging to API level to avoid spamming with the same error from several packages, only left cases when this is necessary. I'll recheck this one.

PR #128, opened 2019-08-20
Update java8.tmpl

vinaysetty:


vinaysetty:
switching to gradle:5.6-jdk12

PR #134, opened 2019-09-16
Fix for rebuilding of approved labs

0xf8f8ff:
Previously approved labs would be marked as unapproved if rebuild of that lab was triggered by push event. Both push event and rebuild button event will now ignore already approved labs.


0xf8f8ff:
No. Do we want them to be able to update approved labs score? It can be done.

meling:
There seems to be a conflict that prevents merging these changes.

PR #142, opened 2019-10-29
Group submission access

0xf8f8ff:
When accessing group submissions, there will no longer be enough to provide any group ID > 0, the current user must also be a group member (or course teacher).

### DIFF
--- a/web/autograder_service.go
+++ b/web/autograder_service.go
@@ -413,10 +413,15 @@ func (s *AutograderService) GetSubmissions(ctx context.Context, in *pb.Submissio
 		s.logger.Errorf("GetSubmissions failed: authentication error: %w", err)
 		return nil, ErrInvalidUserInfo
 	}
+
+	// even if there is no group ID in request and the grp value is nil, checking whether it contains the current user will just return false
+	grp, _ := s.getGroup(&pb.RecordRequest{ID: in.GetGroupID()})
+	s.logger.Debugf("Checking group user: %v+", grp.Contains(usr))
+
 	// ensure that current user is teacher or the current user is owner of the submission request
 	if !s.hasCourseAccess(usr.GetID(), in.GetCourseID(), func(e *pb.Enrollment) bool {
-		return e.Status == pb.Enrollment_TEACHER || usr.GetIsAdmin() || // TODO: remove isadmin after testing stage
-			(e.Status == pb.Enrollment_STUDENT && (usr.IsOwner(in.GetUserID()) || in.GetGroupID() > 0)) //TODO(meling): The Group check should be more precise; check that usr is in group.
+		return e.Status == pb.Enrollment_TEACHER || usr.GetIsAdmin() ||
+			(e.Status == pb.Enrollment_STUDENT && (usr.IsOwner(in.GetUserID()) || grp.Contains(usr)))
 	}) {
 		s.logger.Error("GetSubmissions failed: user is not teacher or submission author")
 		return nil, status.Errorf(codes.PermissionDenied, "only owner and teachers can get submissions")
--- a/web/autograder_service.go
+++ b/web/autograder_service.go
@@ -406,7 +406,7 @@ func (s *AutograderService) DeleteGroup(ctx context.Context, in *pb.RecordReques
 }
 
 // GetSubmissions returns the submissions matching the query encoded in the action request.
-// Access policy: Current User if Owner of submission, Teacher of CourseID.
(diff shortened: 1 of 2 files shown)


web/autograder_service.go
```diff
@@ -406,17 +406,21 @@ func (s *AutograderService) DeleteGroup(ctx context.Context, in *pb.RecordReques
 }
 
 // GetSubmissions returns the submissions matching the query encoded in the action request.
-// Access policy: Current User if Owner of submission, Teacher of CourseID.
+// Access policy: Current User if Owner of submission (or member of the group in case of group submission), Teacher of CourseID.
```
meling: We use comma separate list for disjoint parts of the access policy.
But since this is a long line, I suggest to move it to separate lines, e.g.:
```
// Access policy:
// Current User if Owner of submission,
// Current User is member of group for group submission,
// Teacher of CourseID.
```

web/autograder_service.go
```diff
@@ -406,17 +406,21 @@ func (s *AutograderService) DeleteGroup(ctx context.Context, in *pb.RecordReques
...
 	}
+
+	// even if there is no group ID in request and the grp value is nil, checking whether it contains the current user will just return false
+	grp, _ := s.getGroup(&pb.RecordRequest{ID: in.GetGroupID()})
+
 	// ensure that current user is teacher or the current user is owner of the submission request
 	if !s.hasCourseAccess(usr.GetID(), in.GetCourseID(), func(e *pb.Enrollment) bool {
-		return e.Status == pb.Enrollment_TEACHER || usr.GetIsAdmin() || // TODO: remove isadmin after testing stage
-			(e.Status == pb.Enrollment_STUDENT && (usr.IsOwner(in.GetUserID()) || in.GetGroupID() > 0)) //TODO(meling): The Group check should be more precise; check that usr is in group.
+		return e.Status == pb.Enrollment_TEACHER || usr.GetIsAdmin() ||
```
meling: Note that admin is not specified in the access policy.

web/autograder_service.go
```diff
@@ -406,17 +406,21 @@ func (s *AutograderService) DeleteGroup(ctx context.Context, in *pb.RecordReques
...
-// Access policy: Current User if Owner of submission, Teacher of CourseID.
+// Access policy: Current User if Owner of submission (or member of the group in case of group submission), Teacher of CourseID.
 func (s *AutograderService) GetSubmissions(ctx context.Context, in *pb.SubmissionRequest) (*pb.Submissions, error) {
 	usr, err := s.getCurrentUser(ctx)
 	if err != nil {
 		s.logger.Errorf("GetSubmissions failed: authentication error: %w", err)
 		return nil, ErrInvalidUserInfo
 	}
+
+	// even if there is no group ID in request and the grp value is nil, checking whether it contains the current user will just return false
```
meling: check to:
```
// grp may be nil if there is no group ID in request; this is fine, since the grp.Contains() returns false in this case.
```

web/autograder_service.go
0xf8f8ff: Well, that's a kind of backdoor that you need if you want an admin to be able to check, debug and fix problem for existing course labs by just adding them to the course as a regular student (without promoting to course teacher).
Otherwise you'd have to promote admins to course teachers, and for now this action cannot be undone.

meling: ok, then we must add Admin to the policy doc

web/submissions_test.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	groupSubmission, err = ags.GetSubmissions(ctx, &pb.SubmissionRequest{CourseID: course.ID, GroupID: 1})
+	if err == nil {
+		t.Error("Expected error: only owner and teachers can get submissions")
+	}
+
+	// the third student (not enrolled in the course) should not be able to access submission even if it belongs to that student
+	ctx = withUserContext(context.Background(), student3)
+	personalSubmission, err = ags.GetSubmissions(ctx, &pb.SubmissionRequest{CourseID: course.ID, UserID: student3.ID})
```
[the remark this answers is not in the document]
0xf8f8ff: If `personalSubmission`'s value is anything but nil, the error we check for on the next line will be nil, so the test will fail. Should we also check explicitly that `personalSubmission` has nil value here?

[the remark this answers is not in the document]
meling: Sorry. I didn’t read this accurately. But it wouldn’t hurt to also check for nil on `personalSubmission`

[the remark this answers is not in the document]
meling: On the other hand it isn’t so important

PR #145, opened 2019-10-31
Attempt to fix "too many open files" error

0xf8f8ff:
- shorter write timeout for http server
- closing the body of echo requests manually


meling:
This could of course be related to the problem but it would be nice if we could write a test case to open many connections in a loop and see if we can reproduce the too many open files problem. We don’t need to run this test always but it would be nice to have a test to try it and that we can rerun later to reproduce and similarly check if the fix actually didn’t cause the problem. 

Also it would be relevant to look over the code for other missing close calls.

meling:
Not that we could run such a test in a separate VM and set the open file count to some reasonable small number.

meling:
Merging now

PR #149, opened 2019-11-07
Remove empty repos

0xf8f8ff:
To be able to remove the group repository when removing a group, it is necessary to ensure that no submissions were committed yet to that repository.


public/src/managers/GRPCManager.ts
```diff
@@ -190,9 +192,11 @@ export class GrpcManager {
         return this.grpcSend<Void>(this.agService.updateGroup, grp);
     }
 
-    public deleteGroup(groupid: number): Promise<IGrpcResponse<Void>> {
-        const request = new RecordRequest();
-        request.setId(groupid);
+    public deleteGroup(groupid: number, courseID: number, withRepo: boolean): Promise<IGrpcResponse<Void>> {
```
meling: Should all methods always have `courseID` first in the argument list.

scm/github.go
```diff
@@ -237,6 +237,24 @@ func (s *GithubSCM) UpdateRepoAccess(ctx context.Context, repo *Repository, user
...
+
+	repo, err := s.GetRepository(ctx, opt)
+	if err != nil {
+		return false
+	}
+
+	// test to check how repo commits look like
+	_, _, err = s.client.Repositories.ListCommits(ctx, repo.Owner, repo.Path, nil)
+	if err != nil {
+		if strings.Contains(err.Error(), "Git Repository is empty") {
```
meling: This is a bit fragile to check for such a string, but I understand why. I googled a bit, and came across this [stackoverflow issue](https://stackoverflow.com/questions/19889899/count-number-of-empty-repositories-on-github). Specifically, this comment:

_That would be an interesting statistic. We don't have a simple way to do that right now. However, you might be able to use the GitHub API to get close. You could look through public repositories and compare "pushed_at" and "created_at" dates to see if there has been any activity. Additionally, you could find repositories with a "size" of 0. There's more information on how to find this information, and much more, right here:_

http://developer.github.com/v3/repos/

Would this give us the same in a safer way?

web/autograder_service.go
```diff
@@ -383,13 +383,15 @@ func (s *AutograderService) UpdateGroup(ctx context.Context, in *pb.Group) (*pb.
 
 // DeleteGroup removes group record from the database.
 // Access policy: Teacher of CourseID.
-func (s *AutograderService) DeleteGroup(ctx context.Context, in *pb.RecordRequest) (*pb.Void, error) {
-	usr, err := s.getCurrentUser(ctx)
+func (s *AutograderService) DeleteGroup(ctx context.Context, in *pb.DeleteGroupRequest) (*pb.Void, error) {
+
+	// TODO(vera): will neen an scm passed to the web method in case we want to delete the group repo too (in.WithRepo == true)
```
meling: I think we should always delete the team, group repo and the group database entry. Does it make sense to keep either of these, and not all?

web/groups.go
```diff
@@ -36,15 +36,37 @@ func (s *AutograderService) getGroupByUserAndCourse(request *pb.GroupRequest) (*
...
+func (s *AutograderService) deleteGroup(ctx context.Context, sc scm.SCM, request *pb.DeleteGroupRequest) error {
+	group, err := s.db.GetGroup(request.GetGroupID())
 	if err != nil {
 		return err
 	}
-	if group.Status > pb.Group_REJECTED {
-		return status.Errorf(codes.Aborted, "accepted group cannot be deleted")
+
+	// if withRepo is true, delete repo and team (needs scm)
+	if request.WithRepo {
```
meling: Do we need this option to only delete group in the database?

public/src/managers/GRPCManager.ts
0xf8f8ff: Should they? More than a half methods in this file have course ID as a second argument, but it is easy to fix.

scm/github.go
0xf8f8ff: Unfortunately, none of these options work.
A repo with a few text files pushed to it still has size 0.
For an empty repo `pushed_at` can somehow differ from `created_at` (Example: Repository  grp1  was created at:  2019-11-14 13:54:43 +0000 UTC  and last pushed to at: 2019-11-14 13:54:44 +0000 UTC).

public/src/managers/GRPCManager.ts
meling: I think we should be consistent throughout the codebase that we always have `courseID` and other fields in the same order, if they are used. This is important when inputs are the same type. But especially important for API methods.
However, I will open a separate issue on this, so that we can do accept this one.

PR #155, opened 2019-11-15
New lab result layout

0xf8f8ff:
image

### DIFF
--- a/public/src/components/lab/LabResult.tsx
+++ b/public/src/components/lab/LabResult.tsx
@@ -5,11 +5,9 @@ interface ILabResult {
     submission_id: number;
     progress: number;
     lab: string;
-    course_name: string;
     authorName?: string;
     showApprove: boolean;
     isApproved: boolean;
-    delivered: string;
     onApproveClick: () => void;
     onRebuildClick: (submissionID: number) => Promise<boolean>;
 }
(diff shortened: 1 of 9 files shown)


public/src/components/lab/LabResult.tsx
```diff
@@ -2,16 +2,19 @@ import * as React from "react";
 import { ProgressBar, Row } from "../../components";
 
 interface ILabResult {
+    submission_id: number;
```
meling: submissionID

meling:
Nice work. I like it. The Rebuild button should be aligned to the right.
I’m not sure we should have the tests in a separate table after all. Looks a bit strange. Let’s add those to the table above as;
```
Tests Passed        5
Tests Failed        0
```
Or
```
Tests Passed / Failed        5 / 0
```
Or
```
Passed / Failed        5 / 0 tests
```
I struggle to make up my mind among these.

0xf8f8ff:
Updated layout:

image

meling:
I like the new layout. A few nits: the date formatting; we probably don’t want to display seconds. And we should use European 24h time format, ie without the am/pm. Maybe we could also skip the year in this particular view since that should be obvious from the context of the course itself. I propose instead writing the day, date, month and time, eg: Tuesday 19 Oct, 18:00

0xf8f8ff:
Updated date format and table rows order:

image

PR #157, opened 2019-11-20
Api updates

0xf8f8ff:
Replaced generic RecordRequest with more specific request messages.


meling:
Great! I like that we now have more clear code for the IDs...

PR #158, opened 2019-11-20
Delete group update

0xf8f8ff:
- removed `Reject` option for groups
- group repository and team will always be deleted after confirmation from teacher
- pending groups can now always be deleted


meling:
So, we no longer need the `Group.GroupStatus.REJECTED` field in the enum. It should be fine to give the enums new numbers, since we control all clients and servers.

0xf8f8ff:
All the status values in the database are based on the current enums, so it does not sound that safe to be honest.

public/src/components/teacher/CourseGroup.tsx
```diff
@@ -181,39 +148,37 @@ export class CourseGroup extends React.Component<ICourseGroupProps, ICourseGroup
...
+            await this.deleteGroup(group) :
+            await this.props.courseMan.updateGroupStatus(group.getId(), status);
         this.props.navMan.refresh();
     }
 
-    private async deleteGroup(groupID: number) {
-        let withRepos = true;
-        let readyToDelete = true;
+    private async deleteGroup(group: Group) {
+        let readyToDelete = false;
```
meling: Could we make the logic clearer in this function by doing the following:
```typescript
let readyToDelete = group.getStatus() === Group.GroupStatus.PENDING
...
if (!readyToDelete && !isEmpty) {
  if (confirm(...))
  ...
```

0xf8f8ff: This way it will never delete a non-pending group with empty repo

0xf8f8ff:
Will no longer ask server (and github) about repo size if the group is pending.

public/src/components/teacher/CourseGroup.tsx
```diff
@@ -157,24 +157,23 @@ export class CourseGroup extends React.Component<ICourseGroupProps, ICourseGroup
...
-                readyToDelete = true;
-            // otherwise the group is alredy approved, ask for confimation
-            } else if (confirm(
-                `Warning! Group repository is not empty!
-                Do you still want to delete group, github team
-                and group repository?`,
-            )) {
+        // if approved group - check if repo is empty
+        if (!readyToDelete) {
+            const isEmpty = await this.props.courseMan.isEmptyRepo(courseID, 0, group.getId());
```
meling: We could also reuse `readyToDelete` instead of `isEmpty`. That way, we can avoid the extra `else` clause. 
```typescript
  readyToDelete = await this.props.courseMan.isEmptyRepo(courseID, 0, group.getId());
  if (!readyToDelete && !confirm(...)) {
     readyToDelete = false;
  }
}
```

Or maybe even better?:
```typescript
  readyToDelete = await this.props.courseMan.isEmptyRepo(courseID, 0, group.getId());
  if (!readyToDelete)
     readyToDelete = confirm(...)
  }
}
```


Another alternative to simplify the logic here is to do:

```typescript
readyToDelete = true;
if (!isEmpty && !confirm(...)) {
  readyToDelete = false;
}
```

0xf8f8ff: The second alternative looks best to me. I'll update after running tests on it just to be safe.

meling: Agree

PR #159, opened 2019-11-21
Getgroup fix

0xf8f8ff:
New request message for GetGroup method, as the old one required course ID to be set at all times.


meling:
In principle, this looks reasonable. But I didn't see where you now avoid to specify the courseID and userID.

0xf8f8ff:
Originally the generic `RecordRequest` was used in this method. I've tried to replace it with `GroupRequest`, but the validation method for `GroupRequest` wants `CourseID` to be always present, which is unreasonable when we want to get a group with known ID.

meling:
I see.

PR #160, opened 2019-11-22
UI updates

0xf8f8ff:
- navigation bar always visible
- no empty `Settings` links
- striped tables
- two-button layout for lists with course members
- adjusted position for `Approve All` button

### DIFF
--- a/public/src/pages/views/MemberView.tsx
+++ b/public/src/pages/views/MemberView.tsx
@@ -87,7 +87,7 @@ export class MemberView extends React.Component<IUserViewerProps, IUserViewerSta
                     if (user.link.getStatus() === Enrollment.UserStatus.TEACHER) {
                         links.push({ name: "Teacher", extra: "light" });
                     } else {
-                        links.push({ name: "Assign as Teacher", uri: "teacher", extra: "primary" });
+                        links.push({ name: "Promote", uri: "teacher", extra: "primary" });
                         links.push({ name: "Remove", uri: "reject", extra: "danger" });
                     }
                     return links;
--- a/public/src/pages/views/MemberView.tsx
+++ b/public/src/pages/views/MemberView.tsx
@@ -45,18 +45,17 @@ export class MemberView extends React.Component<IUserViewerProps, IUserViewerSta
                     onChange={(query) => this.handleSearch(query)}
                 />
             {this.renderPendingView(pendingActions)}
-            {this.props.pendingUsers.length > 0 ? this.approveButton() : null}
             {this.renderUserView()}
             {this.renderRejectedView()}
         </div>;
     }
 
-    public componentDidUpdate(prevProps: IUserViewerProps) {
-        if ((prevProps.acceptedUsers.length !== this.props.acceptedUsers.length)
-         || (prevProps.pendingUsers.length !== this.props.pendingUsers.length)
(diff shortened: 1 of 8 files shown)


public/src/pages/views/MemberView.tsx
```diff
@@ -123,44 +123,48 @@ export class MemberView extends React.Component<IUserViewerProps, IUserViewerSta
         </div>;
     }
 
-    public handleAction(userRel: IUserRelation, link: ILink) {
+    public async handleAction(userRel: IUserRelation, link: ILink) {
```
meling: This code was a bit long. Could these be split into separate methods for the different cases?

public/src/pages/views/MemberView.tsx
```diff
@@ -82,11 +81,11 @@ export class MemberView extends React.Component<IUserViewerProps, IUserViewerSta
...
                 [],
-                ActionType.Menu,
+                ActionType.InRow,
                 (user: IUserRelation) => {
                     const links = [];
                     if (user.link.getStatus() === Enrollment.UserStatus.TEACHER) {
-                        links.push({ name: "This is a teacher", extra: "primary" });
+                        links.push({ name: "Teacher", extra: "light" });
                     } else {
                         links.push({ name: "Make Teacher", uri: "teacher", extra: "primary" });
```
meling: Suggest to replace `Make Teacher` with `Appoint as Teacher`.
Also, perhaps make a little space between the buttons, so they are not touching.

public/src/pages/views/MemberView.tsx
```diff
@@ -82,11 +81,11 @@ export class MemberView extends React.Component<IUserViewerProps, IUserViewerSta
...
                         links.push({ name: "Reject", uri: "reject", extra: "danger" });
```
meling: Suggest to change `Reject` with `Remove` (in the Registered Users list). For the Pending Users we should still keep `Reject`.

public/src/pages/views/MemberView.tsx
```diff
@@ -45,18 +45,17 @@ export class MemberView extends React.Component<IUserViewerProps, IUserViewerSta
...
-       }
+    public componentWillReceiveProps(newProps: IUserViewerProps) {
+        this.setState({
+            acceptedUsers: newProps.acceptedUsers,
+            rejectedUsers: newProps.rejectedUsers,
+            pendingUsers: newProps.pendingUsers,
+        });
     }
 
     public renderRejectedView() {
```
meling: Is this still used? (now that we reject users from a course... and delete the enrollment from the database.)

public/src/pages/views/MemberView.tsx
0xf8f8ff: Done.

public/src/pages/views/MemberView.tsx
0xf8f8ff: Are you sure about that? `Appoint as Teacher` button will take even more space.

public/src/pages/views/MemberView.tsx
0xf8f8ff: Done.

public/src/pages/views/MemberView.tsx
meling: I'm leaning towards changing it to `Promote` only in the button. Then we could have a tool tip on the button?? with `Promote to teacher`, if possible?

PR #161, opened 2019-11-22
Rebuild fix

0xf8f8ff:
Fixed a bug where a wrong repository could sometimes be fetched when rebuilding a student submission.


web/autograder_service.go
```diff
@@ -489,13 +489,10 @@ func (s *AutograderService) ApproveSubmission(ctx context.Context, in *pb.Approv
 
 // RefreshSubmission rebuilds the latest student submission for the logged in user
```
meling: Maybe the doc comment needs to be updated to say rebuild for the user specified in ...?

public/src/pages/views/LabResultView.tsx
```diff
@@ -54,8 +55,8 @@ export class LabResultView extends React.Component<ILabInfoProps> {
     }
 
     private getSubmissionInfo(): string {
-        if (this.props.labInfo.latest) {
-            return this.props.labInfo.latest.approved ? "Approved" : "Not approved";
+        if (this.props.assignment.latest) {
+            return this.props.assignment.latest.approved ? "Approved" : "Not approved";
```
meling: Not Approved (capital A)

0xf8f8ff: The method was not even used anymore, so I've just removed it.

web/autograder_service.go
0xf8f8ff: Done.

PR #162, opened 2019-11-25
Groups and enrollments

0xf8f8ff:
Additional changes related to the new approach to deleting groups and rejecting student enrollments:

- DELETED, REJECTED user and group states are marked as deprecated, no longer used in the codebase
- lists of users/groups based on deprecated states are no longer shown
- when rejecting a student group, repository entry in the database will be deleted as well
- when rejecting an enrollment: 
    - database entries for user repository and course enrollment will be deleted,
    - student repository on github will be deleted (asked for confirmation if not empty)
    - user will be removed from all course teams
    - course organization membership will be revoked


public/src/managers/CourseManager.ts
```diff
@@ -111,76 +99,55 @@ export class CourseManager {
...
-     * @param state The new state of the relation
+     * Update status of a course enrollment
      */
     public async changeUserState(link: Enrollment, state: Enrollment.UserStatus): Promise<boolean> {
         const ans = await this.courseProvider.changeUserState(link, state);
         return ans;
     }
 
+    /**
+     *
```
meling: Remove blank line.

0xf8f8ff: Done.

public/src/managers/ServerProvider.ts
```diff
@@ -399,6 +400,10 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
         let buildInfo: IBuildInfo;
         let scoreObj: ITestCases[];
 
+        // IMPORTANT: field names in the backend CI structure, frontend model structure, and here, when
+        // parsing, must match
+        // if experiencing an uncaught error in the browser which results in blank page
+        // when addressing lab information for a student/group, it is likely to originate from here
```
meling: It would be easier to understand, if you make specific references to Go files/packages/file names/struct type etc.
Also try to balance the line length of comments.
We should also use punctuation marks and capitalized start of sentences (at least for comments with multiple sentences.)

public/src/models.ts
```diff
@@ -13,25 +13,22 @@ export interface IUser {
 
 export interface IAssignmentLink {
     /**
-     * The course to the group
+     * The current course
```
meling: I think it looks better to use `// Comment...` instead of `/** Comment ... */` when in structs like this, since it takes up less space. Also, not sure we need a comment on all fields here, e.g. `The current course` could either be removed or made explicit by naming it `currentCourse: Course;` in the struct/interface. But it is ok to keep the comment, if it makes sense to have a comment for the other fields (and it does, I think for the other fields...)

public/src/pages/views/MemberView.tsx
```diff
@@ -158,12 +134,39 @@ export class MemberView extends React.Component<IUserViewerProps, IUserViewerSta
...
+
+            if (readyToDelete) {
+                const result =
+             await this.props.courseMan.changeUserState(userRel.link, Enrollment.UserStatus.NONE);
+                if (result) {
+                    switch (userRel.link.getStatus()) {
+                        case Enrollment.UserStatus.PENDING:
+                            const i = this.state.pendingUsers.indexOf(userRel);
+                            if (i >= 0) {
+                            this.state.pendingUsers.splice(i, 1);
```
meling: There are some indentation problems with this method. Perhaps run the linter and typescript formater in vscode?

public/src/pages/views/MemberView.tsx
```diff
@@ -158,12 +134,39 @@ export class MemberView extends React.Component<IUserViewerProps, IUserViewerSta
...
-            const result =
-             await this.props.courseMan.changeUserState(userRel.link, Enrollment.UserStatus.REJECTED);
-            if (result) {
-                userRel.link.setStatus(Enrollment.UserStatus.REJECTED);
+            let readyToDelete =
+             await this.props.courseMan.isEmptyRepo(this.props.course.getId(), userRel.user.getId(), 0);
+            if (!readyToDelete) {
+                readyToDelete = confirm(
+                    `Warning! User repository is not empty.
+                    Do you still want to reject the user?`
```
meling: vscode (the linter probably) tells me that it wants a `,` after the backtick.

meling: An alternative is to move the `);` up to this line.

public/src/managers/ServerProvider.ts
0xf8f8ff: Fixed.

scm/github.go
```diff
@@ -610,6 +610,23 @@ func (s *GithubSCM) UpdateOrgMembership(ctx context.Context, opt *OrgMembershipO
 	return nil
 }
 
+// RevokeOrgMembership implements the SCM interface
+func (s *GithubSCM) RevokeOrgMembership(ctx context.Context, opt *OrgMembershipOptions) error {
```
meling: Suggest to rename it to `RemoveMember`

web/courses.go
```diff
@@ -45,11 +45,38 @@ func (s *AutograderService) updateEnrollment(ctx context.Context, sc scm.SCM, re
 	}
 
 	switch request.Status {
-	case pb.Enrollment_REJECTED:
-		return s.db.RejectEnrollment(request.UserID, request.CourseID)
+	case pb.Enrollment_NONE:
+
```
meling: Avoid empty line here.

public/src/pages/views/MemberView.tsx
0xf8f8ff: My linter fails to see any problem, could you specify?

public/src/pages/views/MemberView.tsx
0xf8f8ff: Done.

web/groups.go
```diff
@@ -56,6 +56,14 @@ func (s *AutograderService) deleteGroup(ctx context.Context, sc scm.SCM, request
 
 	// when deleting an approved group, remove github repository and team as well
 	for _, repo := range repos {
+
```
meling: Avoid extra lines

scm/github.go
0xf8f8ff: Done.

web/scm_helpers.go
```diff
@@ -138,6 +138,26 @@ func updateGroupTeam(ctx context.Context, sc scm.SCM, org *pb.Organization, grou
 	return sc.UpdateTeamMembers(ctx, opt)
 }
 
+func rejectUserFromCourse(ctx context.Context, sc scm.SCM, login string, repo *pb.Repository) error {
+
```
meling: Avoid extra blank line here.

Should this func be called `removeUserFromCourse`. It will be easier to make connection between the naming, if we use remove or delete consistently.

web/courses.go
0xf8f8ff: Fixed.

web/groups.go
0xf8f8ff: Fixed.

scm/github.go
0xf8f8ff: Might be confused with removing team members though.

web/scm_helpers.go
0xf8f8ff: Done.

public/src/models.ts
0xf8f8ff: Done.

public/src/pages/views/MemberView.tsx
meling: The linter only tells me
- Missing trailing comma [142,59]
- block is empty [227,23]

But the problem I'm actually referring to on line 148, 154, 160 and 169 (and several others) is that the line indentations are not as expected. This should normally be fixed by saving the file, I think; at least my vscode is set up to format typescript on save.

0xf8f8ff: Yeah, trailing comma is fixed, and the empty block is a legacy part of passing a method between elements that I rather would not touch right now.
But indentation problems are still invisible for my linter. Found and fixed some by hand.

PR #163, opened 2019-11-27
Approving labs

0xf8f8ff:
Previously, approving a student submission could not be reverted. 
Now a submission approval can be undone.


database/gormdb.go
```diff
@@ -385,8 +385,8 @@ func (db *GormDB) UpdateSubmission(sid uint64, approved bool) error {
 	if err := db.conn.First(&submission, sid).Error; err != nil {
 		return err
 	}
-	submission.Approved = approved
-	return db.conn.Model(&pb.Submission{}).Update(submission).Error
+	// submission.Approved = approved
+	return db.conn.Model(&submission).Update("approved", approved).Error
```
meling: Don’t understand why pass an “approved” string to the Update?

public/src/components/lab/LabResult.tsx
```diff
@@ -35,8 +35,8 @@ export class LabResult extends React.Component<ILabResult, ILabResultState> {
             approveButton = <div className="btn lab-btn approve-btn"> <button type="button"
                 id="approve"
                 className={this.setButtonColor("approve")}
```
meling: I would probably move the <button... to a new line above.

public/src/components/lab/LabResult.tsx
```diff
@@ -35,8 +35,8 @@ export class LabResult extends React.Component<ILabResult, ILabResultState> {
...
-                onClick={this.props.isApproved ?
-                    () => { console.log("Already approved"); } : () => this.approve()}>
+                title={this.setTooltip()}
+                onClick={() => this.approve()}>
                      {this.setButtonString("approve")} </button> </div>;
```
meling: I would also end the </button... on a new line.

database/gormdb.go
0xf8f8ff: Otherwise the `Approved` field with `false` value would be treated as an empty field and not change. As a result, setting a field to `false`, and passing the whole struct to `Update` would result in the database entry which still has `Approved: true`.

public/src/components/lab/LabResult.tsx
0xf8f8ff: Done.

public/src/components/lab/LabResult.tsx
0xf8f8ff: Fixed.

PR #165, opened 2019-11-28
Course repository links

0xf8f8ff:
- all links leading to the course, student and group repositories will open in a new browser tab
- course repository links on the teacher page will no longer attempt to open a popup window

### DIFF
--- a/public/src/components/teacher/Results.tsx
+++ b/public/src/components/teacher/Results.tsx
@@ -116,7 +116,7 @@ export class Results extends React.Component<IResultsProp, IResultsState> {
     }
 
     private generateUserRepoLink(name: string, username: string): JSX.Element {
-        return <a href={this.props.courseURL + username + "-labs"}>{ name }</a>;
+        return <a href={this.props.courseURL + username + "-labs"} target="_blank">{ name }</a>;
     }
 
     private handleOnclick(item: IStudentSubmission): void {
--- a/public/src/components/teacher/Results.tsx
+++ b/public/src/components/teacher/Results.tsx
@@ -115,8 +115,8 @@ export class Results extends React.Component<IResultsProp, IResultsState> {
         return selector;
     }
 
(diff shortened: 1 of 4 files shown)


public/src/components/teacher/groupHelper.tsx
```diff
@@ -81,5 +81,5 @@ export function slugify(str: string): string {
 }
 
 export function generateGroupRepoLink(groupname: string, courseURL: string): JSX.Element {
-    return <a href={courseURL + slugify(groupname)}>{ groupname }</a>;
+    return <a href={courseURL + slugify(groupname)} target="_blank">{ groupname }</a>;
```
meling: groupName

public/src/components/teacher/Results.tsx
```diff
@@ -116,7 +116,7 @@ export class Results extends React.Component<IResultsProp, IResultsState> {
     }
 
     private generateUserRepoLink(name: string, username: string): JSX.Element {
-        return <a href={this.props.courseURL + username + "-labs"}>{ name }</a>;
+        return <a href={this.props.courseURL + username + "-labs"} target="_blank">{ name }</a>;
```
meling: userName

meling: maybe this should use function or some other way to determine the user repo link. For example, maybe someone prefer to not add `-labs`?

public/src/components/teacher/groupHelper.tsx
0xf8f8ff: Done.

public/src/components/teacher/Results.tsx
0xf8f8ff: And done.

0xf8f8ff: Will be fixed (constant fetched from the server) as a part of the PR with Student Page data fetching optimization.

PR #168, opened 2019-11-28
Group updates

0xf8f8ff:
Previously, methods that use github API would sometimes fail due to context cancelling. That would result in group or student repository created correctly on the github, but no corresponding entries created in the database. Because of that, enrolling such user or approving such group would become impossible, as corresponding methods would fail, and github repositories had to be manually removed.

Now an already existing github repository will be correctly reflected in the database on the next attempt to enroll the student or approve the group.

### DIFF
--- a/scm/github.go
+++ b/scm/github.go
@@ -131,12 +131,22 @@ func (s *GithubSCM) CreateRepository(ctx context.Context, opt *CreateRepositoryO
 			Message: fmt.Sprintf("%+v", opt),
 		}
 	}
-	repo, _, err := s.client.Repositories.Create(ctx, opt.Organization.Path, &github.Repository{
-		Name:    &opt.Path,
-		Private: &opt.Private,
-	})
+
+	// first make sure that repo does not already exist for this user or group
+	repo, _, err := s.client.Repositories.Get(ctx, opt.Owner, opt.Path)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


meling:
It is unclear how this bug would manifest. Or rather, will the user/group remain pending, and a retry will "most likely" work, or what does the user/teacher/group need to do? Can you document this somewhere, e.g. add a troubleshooting section in the doc folder?

0xf8f8ff:
> It is unclear how this bug would manifest. Or rather, will the user/group remain pending, and a retry will "most likely" work, or what does the user/teacher/group need to do? Can you document this somewhere, e.g. add a troubleshooting section in the doc folder?

They will remain pending. I suppose we add a specific error message format for scm errors to be shown to teachers when an scm related error occurs. Such error could provide more detailed information and tips on how to address the issue.

meling:
Okay, we can add a reasonable error message for the user then. That’s better than more documentation.

meling:
Is this one ready with relevant error messages to tell the teacher hints about the need to retry later or something?

0xf8f8ff:
I'd rather have it as a part of a separate PR dedicated to scm related error messages. We have many cases that would benefit from showing more detailed error messages to user: creating a course, approving enrollments, approving or editing groups. Does that sound reasonable?

meling:
Yes.

PR #169, opened 2019-11-29
Submission related updates: in progress

0xf8f8ff:
- optimized fetching of the group labs in the same way individual labs were optimized
- only the latest lab submission will be stored in the database (currently, every submission is recorded, but only the last one for each submitter and assignment is ever accessed. As a result, the database has over 8000 submission entries, of which only about 2500 are relevant)

TODO: make compatible with a database that can have several submission entries for the same submitter and lab.


database/gormdb.go
```diff
@@ -337,6 +337,7 @@ func (db *GormDB) GetNextAssignment(cid uint64, uid uint64, gid uint64) (*pb.Ass
 }
 
 // CreateSubmission creates a new submission record
+// or updates the last one for the given lab and student/group
```
meling: Remove TODO below as we talked about—let’s keep access control out of the database package.

database/gormdb.go
```diff
@@ -375,7 +376,31 @@ func (db *GormDB) CreateSubmission(submission *pb.Submission) error {
...
+	}
+
+	// We want the last record as there can be multiple submissions
+	// for the same submitter and lab in the legacy database.
+	if err := db.conn.Last(query, query).Error; err != nil && err != gorm.ErrRecordNotFound {
+		return err
+	}
+
+	return db.conn.Where(query).Assign(submission).FirstOrCreate(query).Error
+	/*
```
meling: We should avoid committing commented code.

database/gormdb.go
```diff
@@ -375,7 +376,31 @@ func (db *GormDB) CreateSubmission(submission *pb.Submission) error {
 	if assignment+group != 2 {
 		return gorm.ErrRecordNotFound
 	}
-	return db.conn.Create(submission).Error
+
+	// If a submission for the given assignment and student/group already exists, update it.
+	// Otherwise create a new submission record
+	query := &pb.Submission{
+		AssignmentID: submission.GetAssignmentID(),
```
meling: Do we need to create a new Submission object here? Can’t we just reuse the submission object?

database/gormdb.go
```diff
@@ -375,7 +376,31 @@ func (db *GormDB) CreateSubmission(submission *pb.Submission) error {
 	if assignment+group != 2 {
 		return gorm.ErrRecordNotFound
 	}
-	return db.conn.Create(submission).Error
+
+	// If a submission for the given assignment and student/group already exists, update it.
+	// Otherwise create a new submission record
```
meling: This comment refers to the database access below, I think. Currently, it is a bit confusing.

database/gormdb.go
```diff
@@ -375,7 +376,31 @@ func (db *GormDB) CreateSubmission(submission *pb.Submission) error {
...
+	query := &pb.Submission{
+		AssignmentID: submission.GetAssignmentID(),
+		UserID:       submission.GetUserID(),
+		GroupID:      submission.GetGroupID(),
+	}
+
+	// We want the last record as there can be multiple submissions
+	// for the same submitter and lab in the legacy database.
```
meling: What is submitter referring to? I assume (group or user),  but we should be clear.

I would remove the word “legacy”.

web/access_control.go
```diff
@@ -78,6 +78,54 @@ func (s *AutograderService) isEnrolled(userID, courseID uint64) bool {
 	})
 }
 
+// isValidSubmission returns true if submitter has active course enrollment or
+// if submitting group belongs to the given course
+func (s *AutograderService) isValidSubmissionRequest(submission *pb.SubmissionRequest) bool {
```
meling: It is a bit confusing to name this isValid... since we use that for field validation in ag package. Need to think of a more representative name; will think more about it...

web/access_control.go
```diff
@@ -78,6 +78,54 @@ func (s *AutograderService) isEnrolled(userID, courseID uint64) bool {
...
+	if !submission.IsValid() {
```
meling: isn’t the IsValid() method called automatically by the interceptor

web/access_control.go
```diff
@@ -78,6 +78,54 @@ func (s *AutograderService) isEnrolled(userID, courseID uint64) bool {
...
+		fmt.Println("isValidSubmissionRequest failed: submission request not valid: ", submission)
```
meling: We should use the logger for debugging statements.

web/access_control.go
```diff
@@ -78,6 +78,54 @@ func (s *AutograderService) isEnrolled(userID, courseID uint64) bool {
...
+			fmt.Println("isValidSubmissionRequest failed: could not get group ")
+			return false
+		}
+		return true
+	}
+	// ensure that user is enrolled in the course
+	if !s.hasCourseAccess(submission.GetUserID(), submission.GetCourseID(), func(e *pb.Enrollment) bool {
+		return e.Status >= pb.Enrollment_STUDENT
+	}) {
+		fmt.Println("isValidSubmissionRequest failed: failed hasCourseAccess ")
```
meling: Unless you want to log this, the if condition can be returned directly from this method without the if and brackets.

web/courses.go
```diff
@@ -227,11 +227,13 @@ func (s *AutograderService) getAllLabs(request *pb.LabRequest) ([]*pb.LabResultL
 
 	// populate cache map with student labs, filtering the latest submissions for every assignment
 	for _, lab := range allLabs {
-		_, ok := labCache[lab.GetUserID()]
+		// submission record has only one ID: user ID or group ID
+		labID := lab.GetUserID() + lab.GetGroupID()
```
meling: This seems to me to be a potential source of a bug. Don’t user ids and group ids have overlapping domains? At least reading the code it is not obvious to me that this would be safe, and not return result of student’s 3 cache, when accessing the cache as group 3.

Wouldn’t it be better to do a bit wise or with a high bit set, if the group id > 0. E.g. the highest bit in the 64-bit id space?

web/courses.go
```diff
@@ -242,25 +244,53 @@ func (s *AutograderService) getAllLabs(request *pb.LabRequest) ([]*pb.LabResultL
...
 	allCourseLabs := make([]*pb.LabResultLink, 0)
-	for _, usr := range course.Enrollments {
-		// collect all submission values for the user
-		userSubmissions := make([]*pb.Submission, 0)
-		for _, v := range labCache[usr.GetUserID()] {
-			tmp := v
-			userSubmissions = append(userSubmissions, &tmp)
+
+	if request.GetGroupLabs() {
+		for _, grp := range course.Groups {
```
meling: This function start to appear a bit long. Consider breaking up into functions.

meling:
I’m not done reviewing this; will try to get to it tomorrow or Friday. Feel free to remind me on Friday if nothing happened by then.

ag/ag.proto
```diff
@@ -250,6 +250,7 @@ message LabRequest {
     uint64 courseID = 1;
     uint64 submissionID = 2;
     uint64 assignmentID = 3;
+    bool groupLabs = 4;
```
meling: Can’t this be derived from the assignment?

meling: Or is this because we want to avoid a database lookup to determine whether or not it is a group lab?

0xf8f8ff: We use this request to rebuild a certain lab (here both submission and lab IDs are provided), and to fetch all the course labs (that is where we need to distinguish between individual and personal labs). 
Should we rather split it into two different request message types?

database/gormdb.go
0xf8f8ff: The submission object includes fields with test results that the query will also attempt to match. This way, when updating a submission, the old submission would be ignored, as these fields would be different. 
And we just want to check whether a submission for the given lab and user/group exists, no matter what's in the other fields.

web/access_control.go
0xf8f8ff: Yes, when the struct comes from the client, it absolutely does. But the method should work for locally created `SubmissionRequest` structs as well.

database/gormdb.go
0xf8f8ff: Done.

database/gormdb.go
0xf8f8ff: Done.

database/gormdb.go
0xf8f8ff: Fixed.

database/gormdb.go
0xf8f8ff: Fixed.

web/access_control.go
0xf8f8ff: Agree, but currently out of ideas.

web/access_control.go
0xf8f8ff: Fixed.

web/access_control.go
0xf8f8ff: Fixed.

web/courses.go
0xf8f8ff: Yeah, it might be even better to split the logic for group and individual labs.

0xf8f8ff: Done.

web/courses.go
0xf8f8ff: Done.

database/gormdb.go
meling: Fight. This is reasonable. But maybe add a comment to explain this, since I’m bound to ask again :)

ag/ag.proto
meling: I was think about that as an option. Still thinking. Will take another look and think if that would make sense.

database/gormdb.go
0xf8f8ff: Just did.

database/gormdb_assignm_test.go
```diff
@@ -117,62 +117,46 @@ func TestGetNextAssignment(t *testing.T) {
 		t.Errorf("expected unapproved assignment to be %v, got %v", assignment1.ID, nxtUnapproved.ID)
 	}
 
-	// approve submission1
-	if err := db.UpdateSubmission(submission1.ID, true); err != nil {
+	// approve submission for assignment1
+	if err := db.UpdateSubmission(1, true); err != nil {
```
meling: Why change this from `submission1.ID` to just `1`?

meling: Or rather, why not use `assignment1.ID`?

0xf8f8ff: `submission1.ID` is 0 here. `assignment1.ID` will work here, but not on the real database. Is this allowed?

database/gormdb.go
meling: This should read:
```
or updates the most recent submission, as defined by the provided submissionQuery.
The submissionQuery must always specify the assignment, and may specify the ID of
either an individual student or a group, but not both.
```
(or something along those lines.)

Thinking about this more deeply. Maybe this should be split into two, one `CreateSubmission` and another for `UpdateSubmission`.

I'm also wondering if it makes more sense to move the "preparation" of the query outside this (and other similar) functions, rather than allowing to pass the submission details... This would make the function work more like database query thing.

meling: PS: We already have a method called `UpdateSubmission`, which only touches the `approved` field. I suppose this method could be redesigned to support both updating the approved field and other submission fields.

meling: Related to my above point about preparing the query outside this method; this is also done in `GetSubmission` and `GetSubmissions`. But not in `CreateAssignment`. However, this is not a query method, so maybe it doesn't make sense to do it for that method... PS: The latter method does not seem to assign the "score limit" (not sure what assign does).

0xf8f8ff: > PS: The latter method does not seem to assign the "score limit" (not sure what assign does).

It does in master, was part of the previous PR. Will update this branch along with above-mentioned changes.

meling: Shall we keep this as is for this PR, and do another to split the Create and Update functionality and add query input support?

0xf8f8ff: Either is fine for me.

database/gormdb_assignm_test.go
meling: Sorry. I'm got confused by the comment referring to `assignment1`. It is a `submissionID`, so we should not use `assignment1.ID` here; you are right.
However, if `submission1.ID` does not work, we should probably make sure that `submission1.ID` is not 0.

meling: But it seems like it is the database that is generating the `Submission.ID`... using the `CreateSubmission` method.

0xf8f8ff: It does have ID in the database, but locally the struct is `submission1 := pb.Submission{AssignmentID: assignment1.ID, UserID: user.ID}`, so, no ID.

0xf8f8ff: Actually, I've fixed it: gorm method will populate the query structs just as with assignments, so that IDs can be used directly.

meling: I thought so... If you can resolve the conflict; I'll approve this PR and we can start another PR to fix the database methods (see Trello card).

0xf8f8ff: Done.

PR #176, opened 2019-12-09
Demote/reject teacher assistants

0xf8f8ff:
- course teachers can now be demoted to students or rejected from course by the course creator
- course creator cannot be demoted/rejected

### DIFF
--- a/web/access_control.go
+++ b/web/access_control.go
@@ -118,13 +118,6 @@ func (s *AutograderService) isValidSubmission(submissionID uint64) bool {
 	return s.isValidSubmissionRequest(request)
 }
 
-// isEnrolled returns true if the given user is enrolled as student for the given course.
-func (s *AutograderService) isStudent(userID, courseID uint64) bool {
-	return s.hasCourseAccess(userID, courseID, func(e *pb.Enrollment) bool {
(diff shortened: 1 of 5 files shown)


web/access_control.go
```diff
@@ -132,6 +132,25 @@ func (s *AutograderService) isTeacher(userID, courseID uint64) bool {
...
+
+	if (currEnrol.GetStatus() == pb.Enrollment_TEACHER) && (enrol.GetStatus() != pb.Enrollment_TEACHER) {
+		return true
+	}
+	return false
+}
+
+// isCourseCreator returns true if the given user is course creator for the given course
+func (s *AutograderService) isCourseCreator(courseID, userID uint64) bool {
+	course, _ := s.db.GetCourse(courseID, false)
```
meling: GetCourse can return `nil`; so here you should also check for `err != nil` and if so, just return false.

meling: Or alternatively, check that `course != nil` before using it for `GetCourseCreatorID()`.

0xf8f8ff: But if `course` is nil, `course.GetCourseCreatorID()` should return 0, shouldn't it?

meling: That is correct. So this should work fine.

PR #185, opened 2020-01-22
submission update fix

0xf8f8ff:
Fix for new submissions not being updated correctly in the database.

### DIFF
--- a/database/gormdb.go
+++ b/database/gormdb.go
@@ -398,7 +398,9 @@ func (db *GormDB) CreateSubmission(submission *pb.Submission) error {
 	// If a submission for the given assignment and student/group already exists, update it.
 	// Otherwise create a new submission record
 	var labSubmission pb.Submission
-	return db.conn.Where(query).Assign(submission).FirstOrCreate(&labSubmission).Error
+	err := db.conn.Where(query).Assign(submission).FirstOrCreate(&labSubmission).Error
+	submission.ID = labSubmission.GetID()
+	return err
 
 }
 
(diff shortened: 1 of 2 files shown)


meling:
The change looks fine on its own, but I'll merge into master once we have revised tests for this.

database/gormdb.go
```diff
@@ -397,7 +397,9 @@ func (db *GormDB) CreateSubmission(submission *pb.Submission) error {
 
 	// If a submission for the given assignment and student/group already exists, update it.
 	// Otherwise create a new submission record
-	return db.conn.Where(query).Assign(submission).FirstOrCreate(submission, query).Error
+	var labSubmission pb.Submission
+	return db.conn.Where(query).Assign(submission).FirstOrCreate(&labSubmission).Error
+
```
meling: Remove empty line

meling:
On the other hand, I see that the test that no longer passes seems to be the test for `GetNextAssignment` which may be deprecated going forward, since we now base our assignment selection on the actual push event (what changed)... However, the changed method should tested with data that passed before, but no longer passes... (Didn't analyze the details of the tests that exists related to `CreateSubmission`...)

database/gormdb.go
0xf8f8ff: Fixed.

0xf8f8ff:
The test was failing due to `FirstOrCreate` not affecting the original structure any more. All submissions in the test had ID=0 because of that. 

Got it working for now, will replace with simpler test when this one becomes deprecated.

PR #192, opened 2020-01-28
Removed group-related error logging

0xf8f8ff:
Every time a user loads a student page, it would check whether the user is in group. If not, an error message would be logged on the server side, resulting in a lot of uninformative spam.
This error message will no longer be logged in situations where this behavior (no group record in the database) is expected.

Bonus change: `make remote` Makefile target will no longer recompile the bundle file as it is never useful.


web/autograder_service.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/autograde/aguis/database"
 	scms "github.com/autograde/aguis/scm"
 	"github.com/autograde/aguis/web/auth"
+	"github.com/jinzhu/gorm"
```
meling: Here we are leaking internal error types from Gorm. I've looked at the upspin `errors` package. I propose we adopt something like that, creating our own error types that can be checked. Let's do it stepwise, by adding this one error type first.

Makefile
```diff
@@ -134,5 +134,4 @@ local:
 
 remote:
 	@echo "Changing grpc client location to remote domain"
-	@cd ./public/src/managers/; sed -i 's/"http:\/\/localhost:8080"/"https:\/\/" + window.location.hostname/g' GRPCManager.ts
-	@cd ./public; webpack
\ No newline at end of file
+	@cd ./public/src/managers/; sed -i 's/"http:\/\/localhost:8080"/"https:\/\/" + window.location.hostname/g' GRPCManager.ts
```
meling: Missing newline.

web/autograder_service.go
0xf8f8ff: Fixed it by adding a new simple error to already existing set of custom errors for `web` package, because our goal here is just to avoid unnecessary logging.

Makefile
0xf8f8ff: Done.

PR #196, opened 2020-01-31
Approving a new group fix

0xf8f8ff:
Sometimes a new group could not be approved because of canceled context on GitHub side. This way, repository or team for the group might have been created on GitHub, but nothing would be reflected in the database. As a result, approving such a group would become impossible because of GitHub and database mismatch.

This fix introduces following changes:

- two groups with names that would result in the same github repository name can no longer be created
- issue when some students could not create a new group with "not in group" error message is now fixed
- if a group repository or team already have been created on GitHub, this information will be updated in the database instead of failing the request
- a group that could not be approved due to GitHub related issues now can be approved later without any actions required from the user

### DIFF
--- a/public/src/components/forms/GroupForm.tsx
+++ b/public/src/components/forms/GroupForm.tsx
@@ -251,15 +251,18 @@ export class GroupForm extends React.Component<IGroupProps, IGroupState> {
             errors.push("Group mush have members.");
         }
 
-        if (this.state.curUser
-            && (this.state.curUser.link.getStatus() === Enrollment.UserStatus.STUDENT
-                || this.state.curUser.link.getStatus() === Enrollment.UserStatus.TEACHER)
-            && !this.isCurrentStudentSelected(this.state.curUser)) {
-                if (this.state.curUser.link.getStatus() !== Enrollment.UserStatus.TEACHER) {
-                    errors.push("You must be a member of the group");
-                }
-        }
+        if (!this.userValidate(this.state.curUser)) {
+            errors.push("You must be a member of the group");
+        }        
         return errors;
+    }
+
+    private userValidate(curUser: IUserRelation | undefined): boolean {
+        if (curUser && curUser.link.getStatus() === Enrollment.UserStatus.TEACHER) {
+            return true
+        } else {
+            return curUser ? curUser.link.getStatus() === Enrollment.UserStatus.STUDENT && this.isCurrentStudentSelected(curUser) : false
+        }
     }
 
     private getFlashErrors(errors: string[]): JSX.Element {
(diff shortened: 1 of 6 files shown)


web/scm_helpers.go
```diff
@@ -194,6 +194,10 @@ func isEmpty(ctx context.Context, sc scm.SCM, repos []*pb.Repository) error {
 // creating a course, approving, changing status of, or deleting
 // a course enrollment or group
 func contextCanceled(ctx context.Context) bool {
+	// debugging context related errors
```
meling: Suggest to add logger to the signature and make this logging more permanent, so that we have a better chance of identifying root cause... Maybe also pass in the method name from where we call this func. Or can this be simplified with a new `errors` package.

0xf8f8ff: It most certainly can be done easily with introduction of the new errors package, right now it just seems that the context is never seen as canceled by the server, so this method is probably unnecessary. 
Since this type of errors is extremely rare, see no other option than trying to catch it in production logs.

scm/github.go
```diff
@@ -389,35 +389,44 @@ func (s *GithubSCM) CreateTeam(ctx context.Context, opt *TeamOptions) (*Team, er
...
+
+	// first check whether the team with this name already exists on this organization
+	team, _, err := s.client.Teams.GetTeamBySlug(ctx, opt.Organization.Path, slug.Make(opt.TeamName))
 	if err != nil {
-		if opt.TeamName != TeachersTeam && opt.TeamName != StudentsTeam {
-			return nil, ErrFailedSCM{
-				Method:   "CreateTeam",
-				Message:  fmt.Sprintf("failed to create GitHub team %s, make sure it does not already exist", opt.TeamName),
-				GitError: fmt.Errorf("failed to create GitHub team %s: %w", opt.TeamName, err),
+		s.logger.Infof("Team %s not found as expected: %s", opt.TeamName, err)
```
meling: I think we should avoid `Info` and use only `Debug` everywhere. Rational: we will not ever show only one of Debug or Info levels.

meling: I mean, it is ok to use `Error` level and `Debug` level.

public/src/components/forms/GroupForm.tsx
```diff
@@ -253,14 +250,14 @@ export class GroupForm extends React.Component<IGroupProps, IGroupState> {
         if (this.state.selectedStudents.length === 0) {
             errors.push("Group mush have members.");
         }
+
         if (this.state.curUser
             && (this.state.curUser.link.getStatus() === Enrollment.UserStatus.STUDENT
```
meling: This if should be simplified through a function that returns true if these conditions are satisfied.

public/src/components/forms/GroupForm.tsx
```diff
@@ -281,8 +278,13 @@ export class GroupForm extends React.Component<IGroupProps, IGroupState> {
     }
 
     private isCurrentStudentSelected(student: IUserRelation): boolean {
-        const index = this.state.selectedStudents.indexOf(student);
-        return index >= 0;
+        let ans = false;
```
meling: It would be better to use `found` or `foundSelected` instead of `ans`.

scm/github.go
0xf8f8ff: Fixed.

public/src/components/forms/GroupForm.tsx
0xf8f8ff: Done.

public/src/components/forms/GroupForm.tsx
0xf8f8ff: Fixed.

PR #200, opened 2020-02-03
Teacher status change restrictions

0xf8f8ff:
Any course teacher can now freely promote and demote other course teachers. The course creator can never be demoted.

### DIFF
--- a/web/autograder_service.go
+++ b/web/autograder_service.go
@@ -470,7 +470,8 @@ func (s *AutograderService) GetSubmissions(ctx context.Context, in *pb.Submissio
 	return submissions, nil
 }
 
-// GetCourseLabSubmissions returns all the latest submissions for every individual course assignment for each course student
+// GetCourseLabSubmissions returns all the latest submissions
+// for every individual or group course assignment for all course students/groups.
 // Access policy: Admin enrolled in CourseID, Teacher of CourseID.
 func (s *AutograderService) GetCourseLabSubmissions(ctx context.Context, in *pb.LabRequest) (*pb.LabResultLinks, error) {
 	usr, err := s.getCurrentUser(ctx)
--- a/web/autograder_service.go
+++ b/web/autograder_service.go
@@ -87,7 +87,7 @@ func (s *AutograderService) UpdateUser(ctx context.Context, in *pb.User) (*pb.Us
 		s.logger.Errorf("UpdateUser failed to update user %d: user is not admin or course creator", in.GetID())
 		return nil, status.Errorf(codes.PermissionDenied, "only admin can update another user")
 	}
-	usr, err = s.updateUser(usr.IsAdmin, usr.Login, in)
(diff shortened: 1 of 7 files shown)


web/autograder_service.go
```diff
@@ -217,14 +217,10 @@ func (s *AutograderService) UpdateEnrollment(ctx context.Context, in *pb.Enrollm
 		return nil, status.Errorf(codes.PermissionDenied, "only teachers can update enrollment status")
 	}
 	if s.isCourseCreator(in.CourseID, in.UserID) {
-		s.logger.Error("UpdateEnrollment failed: user is course creator, whose enrollment cannot be changed")
+		s.logger.Errorf("UpdateEnrollment failed: user %s has attempted to demote course creator", usr.GetName())
```
meling: user %s attempted to

0xf8f8ff: Fixed.

web/users.go
```diff
@@ -52,12 +52,17 @@ func (s *AutograderService) getUsers() (*pb.Users, error) {
 // updateUser promotes the given user to administrator and/or
 // make other changes to the user database entry for the given user.
 // The isAdmin flag must be true to promote the given user to admin.
-func (s *AutograderService) updateUser(isAdmin bool, request *pb.User) (*pb.User, error) {
+func (s *AutograderService) updateUser(isAdmin bool, curUser string, request *pb.User) (*pb.User, error) {
```
meling: Change signature take `currentUser` (`*pb.User` object) instead. Then replace the `isAdmin` bool and `curUser` string with `currentUser.IsAdmin` and `currentUser.Login`.

web/courses.go
```diff
@@ -38,11 +38,15 @@ func (s *AutograderService) createEnrollment(request *pb.Enrollment) error {
...
 // updateEnrollment accepts or rejects a user to enroll in a course.
-func (s *AutograderService) updateEnrollment(ctx context.Context, sc scm.SCM, request *pb.Enrollment) error {
+func (s *AutograderService) updateEnrollment(ctx context.Context, sc scm.SCM, curUser string, request *pb.Enrollment) error {
 	enrollment, err := s.db.GetEnrollmentByCourseAndUser(request.CourseID, request.UserID)
 	if err != nil {
 		return err
 	}
+	// log changes to teacher status
+	if enrollment.Status == pb.Enrollment_TEACHER || request.Status == pb.Enrollment_TEACHER {
+		s.logger.Debugf("User %s attempts changing enrollment status of user %d from %s to %s", curUser, enrollment.UserID, enrollment.Status, request.Status)
```
meling: I'm still thinking about a way to avoid passing `curUser` just to print a debug statement...

0xf8f8ff: Yeah, I'm not very fond of it either. It is just that we have information about the current user at `autograder_service` level, but no information about the database user/enrollment entry, and vice versa.

PR #201, opened 2020-02-03
Group methods refactoring

0xf8f8ff:
- refactored scm-related methods out of `updateGroup` method
- less frequent GitHub talk when updating group info
- updated `fake` provider method
- replaced struct fields with simple string fields in scm struct types
- updated logic when attempting to change group name


cmd/scm/main.go
```diff
@@ -437,9 +437,9 @@ func createTeam(client *scm.SCM) cli.ActionFunc {
 			return cli.NewExitError("team user names must be provided (comma separated)", 3)
 		}
 		opt := &scm.TeamOptions{
-			Organization: &pb.Organization{Path: c.String("namespace")},
-			TeamName:     c.String("team"),
-			Users:        users,
+			OrgPath:  c.String("namespace"),
```
meling: Would be nice to have documentation for more of the commands available through the `scm` command. There are some command line examples at the top of the file that can be expanded upon.

web/scm_helpers.go
```diff
@@ -30,27 +30,31 @@ var (
...
-// the members of the team. For single student repositories, the user names are
-// typically just the one student's user name.
+// specified namespace (typically the course name), and teamName (name of the group).
+// Repository path and name of the GitHub team will coincide with the group name (teamName).
+// UserNames slice contains GitHub names of all group members.
 // This function performs several sequential queries and updates on the SCM.
 // Ideally, we should provide corresponding rollbacks, but that is not supported yet.
-func createRepoAndTeam(ctx context.Context, sc scm.SCM, org *pb.Organization, path, teamName string, userNames []string) (*scm.Repository, *scm.Team, error) {
+func createRepoAndTeam(ctx context.Context, sc scm.SCM, orgID uint64, teamName string, userNames []string) (*pb.Repository, *scm.Team, error) {
+
```
meling: Remove blank line.

web/scm_helpers.go
```diff
@@ -30,27 +30,31 @@ var (
 )
 
 // createRepoAndTeam invokes the SCM to create a repository and team for the
-// specified namespace (typically the course name), the path of the repository
-// (typically the name of the student with a '-labs' suffix or the group name).
-// The team name is the student name or group name, whereas the user names are
-// the members of the team. For single student repositories, the user names are
-// typically just the one student's user name.
+// specified namespace (typically the course name), and teamName (name of the group).
```
meling: I don't recall why I wrote namespace here, but it would be helpful if this were reflected in the variable names. Maybe it should instead refer to `orgID`, e.g.: *... and team for the specified course (represented with organization ID). The SCM team name is also used as the group name and repository path. The provided user names represent the SCM group members.*

web/groups.go
```diff
@@ -142,46 +143,40 @@ func (s *AutograderService) updateGroup(ctx context.Context, sc scm.SCM, request
...
-
-	org, err := sc.GetOrganization(ctx, &scm.GetOrgOptions{ID: course.OrganizationID})
-	if err != nil {
-		return fmt.Errorf("updateGroup: organization not found: %w", err)
-	}
-
 	if len(repos) == 0 {
-		// found no repos for the group; create group repo and team
-		if request.GetName() != "" {
+
```
meling: Remove blank line

web/groups.go
```diff
@@ -142,46 +143,40 @@ func (s *AutograderService) updateGroup(ctx context.Context, sc scm.SCM, request
...
 			return err
 		}
 		group.TeamID = team.ID
-	} else {
-		// github team already exists, update its members
-		// use the group's existing team ID obtained from the database above.
-		if err := updateGroupTeam(ctx, sc, org, group); err != nil {
+		// if updating group with existing team, group name will not be changed
+		// to avoid a mismatch between database and github names
+		s.logger.Debugf("updateGroup: name of the new GitHub team: %s, requested group name: %s", team.Name, request.Name)
```
meling: Should avoid referencing GitHub here and in comments. Use SCM instead.

web/groups.go
```diff
@@ -142,46 +143,40 @@ func (s *AutograderService) updateGroup(ctx context.Context, sc scm.SCM, request
...
+		if team.Name != request.Name {
+			group.Name = team.Name
+		}
+	}
+
+	// if there are changes in group members, update GitHub team
+	if group.ContainsAll(request) {
```
meling: This comment and the check itself appears to be contradicting. Should this be `!group.ContainsAll(request)`?

web/groups.go
```diff
@@ -130,6 +130,7 @@ func (s *AutograderService) updateGroup(ctx context.Context, sc scm.SCM, request
 	if err != nil {
 		return err
 	}
+	request.Users = users
```
meling: I'm uncomfortable to touch the request object in this way. Is this necessary?

web/groups.go
meling: We should augment the test case with a scenario that capture this.

web/groups.go
meling: I tried to comment this line and the tests still pass. So we should augment the test with a scenario that capture the effect of this line.

0xf8f8ff: Yeah, the problem with scm-related tests is that they rely on `Fake` scm client, which is lagging behind a lot with respect to GitHub, so the test will pass here, but it still would not work in the real life. I've updated a few minor things, but maybe it is worth it to keep it more up to date.

meling: Ah, good point about the `Fake` scm client. Can we update this so that it becomes possible to test these particular things that we are interested in here...? Another option is to actually run tests against GitHub, using a similar approach to that in `web/hooks/github_test.go` and other places where we use `GITHUB_ACCESS_TOKEN`. Although the latter is a bit unsatisfying, since you expose your tests to non-determinism of GitHub access and behavior. It would be better to have repeatable test cases using the `Fake` client, that can also be run by the Travis CI...

0xf8f8ff: Yeah, added a task to the backlog, feels like it deserves a separate PR.

cmd/scm/main.go
0xf8f8ff: Is it better to add them to other examples inside the file or to the documentation (we have some there)? Or both?

web/scm_helpers.go
0xf8f8ff: Fixed.

web/groups.go
0xf8f8ff: Oops, how come did it ever managed to correctly update group memberships before?..

cmd/scm/main.go
meling: Maybe it could be part of the documentation of the `scm` command, when run with `scm -h`. Not sure how to add such extra documentation/examples though. Right now it is just showing commands and options. Anyway, I don't want to prioritize that at this time, but if you could add a Trello card to the backlog, so we have the idea written down, and we can fix it later, when and if we get time.

But for this particular case, maybe just add the comment / examples inline in the file for now, and we can decide later where scm command docs should reside.

PR #203, opened 2020-02-06
Gorm updates cleanup

0xf8f8ff:
- added comments for all interface methods
- grouped gormdb methods
- removed deprecated GetNextUnapprovedAssignment method and related test

TODO:
- add tests for assignments/submissions logic
- refactor/consolidate database methods


ag/ag.proto
```diff
@@ -65,6 +65,9 @@ message Course {
     repeated Enrollment enrollments = 10;
     repeated Assignment assignments = 12;
     repeated Group groups = 13;
+    // or rename all Path fields to something else? See also scm structs.
+    // Name sounds best, but would be misleading
```
meling: Update this comment so it is useful instead of a question. Maybe something like:
```go
// The organization's SCM name, e.g. uis-dat520-2020.
```

0xf8f8ff: Fixed.

database/database.go
```diff
@@ -6,61 +6,57 @@ import (
...
-	// CreateSubmission creates a submission in the database.
-	CreateSubmission(*pb.Submission) error
-	// UpdateSubmission updates the specified submission with approved or not approved.
-	UpdateSubmission(submissionID uint64, approved bool) error
-	// GetSubmission returns a single submission matching the given query.
-	GetSubmission(query *pb.Submission) (*pb.Submission, error)
-	// GetSubmissions returns a list of submission entries for the given course, matching the given query.
-	GetSubmissions(cid uint64, query *pb.Submission) ([]*pb.Submission, error)
-	GetCourseSubmissions(uint64, bool) ([]pb.Submission, error)
+	UpdateGroupEnrollment(userID, courseID uint64) error
```
meling: The func UpdateGroupEnrollment is not used in the code (except for one test). Is this still useful to keep? Also the comment has a typo "aproved" and is also confusing that we don't take a groupID or something as input here...
The comment in `gorm.db` says: _UpdateGroupEnrollment sets GroupID of a student enrollment to 0._

0xf8f8ff: Hm, it is supposed to be used, at least the test starts failing when this part is removed. I'll re-check.

PR #207, opened 2020-02-14
Update lab build information on rebuild

0xf8f8ff:
- all lab hand-in information should update on rebuild
- execution time is now in seconds:

image


public/src/components/lab/LastBuildInfo.tsx
```diff
@@ -50,8 +50,11 @@ export class LastBuildInfo extends React.Component<ILastBuildInfo, ILastBuildInf
         if (delivered >= deadline) {
             classString = "past-deadline";
         }
-
         return <div className={classString}>{formatDate(delivered)}</div>;
     }
 
+    private formatTime(executionTime: number): number {
```
meling: We should just do this conversion on the server side; we already do a conversion (Line 61, `ci/results.go`):
```go
			ExecTime:  execTime.Milliseconds(),
```
Could be changed to:
```go
			ExecTime:  execTime.Seconds(),
```

meling:
This looks promising. I wonder could we also update on a push notification from github? Is everything driven by the frontend, or is there a way to notify the client. Maybe when a user navigates to a page with relevant submission/build info, we could invoke the server over a stream method, that can be used to send changes on push events from github. Or alternatively, the client could poll the server at regular intervals. If this does not have an obvious solution, let's put it on Trello for now.

public/src/components/lab/LastBuildInfo.tsx
0xf8f8ff: Should we also change `ExecTime` type to float64 then?

PR #214, opened 2020-02-20
Set assignment folder name as assignment name

0xf8f8ff:
To avoid possible mismatches between assignment name provided in the `assignment.yaml` file, and the actual assignment folder name (must be identical, but no way to enforce it), name the assignment by its folder. 
This will make the `Name` field in the yml files unnecessary.

### DIFF
--- a/assignments/assignments_parser.go
+++ b/assignments/assignments_parser.go
@@ -24,7 +24,6 @@ const (
 // public to allow parsing.
 type assignmentData struct {
 	AssignmentID uint   `yaml:"assignmentid"`
-	Name         string `yaml:"name"`
 	Language     string `yaml:"language"`
 	Deadline     string `yaml:"deadline"`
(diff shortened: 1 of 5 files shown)


assignments/assignments_parser.go
```diff
@@ -55,6 +54,7 @@ func parseAssignments(dir string, courseID uint64) ([]*pb.Assignment, error) {
 				if err != nil {
 					return fmt.Errorf("could not to read %q file: %w", filename, err)
 				}
+				labDirectoryName := filepath.Base(filepath.Dir(path))
```
meling: No need to declare this as a separate variable, just do it inline in the creation of the `pb.Assignment` object on line 74. But also add a comment right above Line 70:
```
// The assignment Name field is the folder name of the assignment.
```
And also remove the TODO at the top of the func, since this solves the problem.

0xf8f8ff: Done.

PR #215, opened 2020-02-20
Error messages for group and member view

0xf8f8ff:
When changing enrollment status, approving, editing or deleting groups fails, there were no error messages shown to the user.

### DIFF
--- a/public/src/managers/CourseManager.ts
+++ b/public/src/managers/CourseManager.ts
@@ -224,25 +224,6 @@ export class CourseManager {
         return null;
     }
 
-    public async getGroupCourseForTeacher(group: Group, course: Course, assignments: Assignment[]):
-        Promise<IAssignmentLink | null> {
-        // Fetching group enrollment status
-        if (group.getCourseid() === course.getId()) {
-            const enrol = new Enrollment();
-            enrol.setGroupid(group.getId());
-            enrol.setCourseid(course.getId());
-            enrol.setGroup(group);
-            const groupCourse: IAssignmentLink = {
-                link: enrol,
-                assignments: [],
-                course,
-            };
-            await this.fillLinksGroup(group, groupCourse, assignments);
-            return groupCourse;
-        }
-        return null;
-    }
-
     public async getGroupByUserAndCourse(courseID: number, userID: number): Promise<Group | null> {
         return this.courseProvider.getGroupByUserAndCourse(courseID, userID);
     }
--- a/public/src/managers/CourseManager.ts
+++ b/public/src/managers/CourseManager.ts
@@ -16,7 +16,7 @@ export interface ICourseProvider {
         Promise<Enrollment[]>;
 
     addUserToCourse(course: Course, user: User): Promise<boolean>;
-    changeUserState(link: Enrollment, state: Enrollment.UserStatus): Promise<boolean>;
+    changeUserState(link: Enrollment, state: Enrollment.UserStatus): Promise<Status>;
     approveAll(courseID: number): Promise<boolean>;
 
     createNewCourse(course: Course): Promise<Course | Status>;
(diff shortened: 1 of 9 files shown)


public/src/managers/TempDataProvider.ts
```diff
@@ -187,9 +187,11 @@ export class TempDataProvider implements IUserProvider, ICourseProvider {
         throw new Error("Method not implemented");
     }
 
-    public async changeUserState(link: Enrollment, state: Enrollment.UserStatus): Promise<boolean> {
+    public async changeUserState(link: Enrollment, state: Enrollment.UserStatus): Promise<Status> {
         link.setStatus(state);
-        return true;
+        const stat = new Status();
+        stat.setCode(0);
+        return stat;
```
meling: Can this be written as: `return new Status().setCode(0);`  or `return (new Status()).setCode(0);`  ?

public/src/managers/CourseManager.ts
```diff
@@ -224,38 +224,19 @@ export class CourseManager {
         return null;
     }
 
-    public async getGroupCourseForTeacher(group: Group, course: Course, assignments: Assignment[]):
```
meling: Is this obsoleted?

web/groups.go
```diff
@@ -107,6 +107,17 @@ func (s *AutograderService) updateGroup(ctx context.Context, sc scm.SCM, request
 		return err
 	}
 
+	// allow changing the name of the group only if the group
```
meling: Could this instead be converted to a method on Group or something? And maybe instead of reassigning the `group.Name` field, that method could return the new name.

web/groups.go
```diff
@@ -193,11 +204,11 @@ func (s *AutograderService) getGroupUsers(request *pb.Group) ([]*pb.User, error)
 }
 
 // isValidGroupName ensures that SCM team and repository names for the given group
-// will not coincide with one of the existing groups
+// will not coincide with one of the existing approved groups
 func (s *AutograderService) isValidGroupName(courseID uint64, groupName string) bool {
 	courseGroups, _ := s.db.GetGroupsByCourse(courseID)
 	for _, group := range courseGroups {
-		if slug.Make(groupName) == slug.Make(group.GetName()) {
+		if slug.Make(groupName) == slug.Make(group.GetName()) && group.GetStatus() == pb.Group_APPROVED {
```
meling: Suggest to change `db.GetGroupsByCourse` to take an optional query parameter that can reduce the number of returned groups to those with a given group status (e.g. `pb.Group_PENDING` or `pb.Group_APPROVED`)? The signature would look something like: 
```go
	GetGroupsByCourse(cid uint64, statuses ...pb.Group_GroupStatus) ([]*pb.Group, error)
```
This change can be implemented without changing existing uses of `GetGroupsByCourse()`, but maybe other uses can take advantage of the extra query parameter. This would be similar to `GetEnrollmentsByCourse` and `GetCoursesByUser`, which already support a similar style.

With such a change, you would avoid this check that the group is approved, while looping over them.

0xf8f8ff: Sure, but I better add this to the gorm related PR #203 to avoid unnecessary conflicts.

public/src/managers/TempDataProvider.ts
0xf8f8ff: Unfortunately, not. Both `new Status().setCode(0)` and `(new Status()).setCode(0);` expressions are functions returning a void, we cannot return them when a promise of type `Status` is expected.

public/src/managers/CourseManager.ts
0xf8f8ff: It was used to retrieve all group lab submissions before, now we have another method that fetches individual or group labs depending on the context.

web/groups.go
0xf8f8ff: Well, it expects a set of the existing groups to check against. Do you suggest a method on Group that just takes a slice of groups and compares the names?

web/groups.go
meling: Now that #203 has been merged, we should change the above before I merge this PR.

0xf8f8ff: Fixed now.

PR #220, opened 2020-02-26
Fetching data for student page

0xf8f8ff:
Attempt to rewrite student page methods to minimize the amount of server calls.


public/src/components/forms/GroupForm.tsx
```diff
@@ -258,10 +258,10 @@ export class GroupForm extends React.Component<IGroupProps, IGroupState> {
     }
 
     private userValidate(curUser: IUserRelation | undefined): boolean {
-        if (curUser && curUser.link.getStatus() === Enrollment.UserStatus.TEACHER) {
+        if (curUser && curUser.enrollment.getStatus() === Enrollment.UserStatus.TEACHER) {
             return true
         } else {
-            return curUser ? curUser.link.getStatus() === Enrollment.UserStatus.STUDENT && this.isCurrentStudentSelected(curUser) : false
+            return curUser ? curUser.enrollment.getStatus() === Enrollment.UserStatus.STUDENT && this.isCurrentStudentSelected(curUser) : false
```
meling: This func can be simplified by doing
```typescript
if (!curUser) {
  return false
}
```

0xf8f8ff: What about status checks?

public/src/components/forms/GroupForm.tsx
```diff
@@ -253,16 +253,19 @@ export class GroupForm extends React.Component<IGroupProps, IGroupState> {
...
     private userValidate(curUser: IUserRelation | undefined): boolean {
-        if (curUser && curUser.link.getStatus() === Enrollment.UserStatus.TEACHER) {
-            return true
-        } else {
-            return curUser ? curUser.link.getStatus() === Enrollment.UserStatus.STUDENT && this.isCurrentStudentSelected(curUser) : false
+        if (!curUser) {
+            return false;
         }
+        if (!((curUser.enrollment.getStatus() === Enrollment.UserStatus.TEACHER) ||
+         (curUser.enrollment.getStatus() === Enrollment.UserStatus.STUDENT && this.isCurrentStudentSelected(curUser)))) {
```
meling: Hmmm. I was expecting this logic to become simpler too. If I understand the logic right here, what you are testing is:
```typescript
const status = curUser.enrollment.getStatus()
return status === Enrollment.UserStatus.TEACHER || (status === Enrollment.UserStatus.STUDENT && this.isCurrentStudentSelected(curUser))
```
(there is no need for an if-statement here)

Another alternative could be make this into a method on `this`, similar to the `isCurrentStudentSelected` method, or better yet make these as methods on the `IUserRelation` object, so that you avoid the `this` object.

web/autograder_service.go
```diff
@@ -271,6 +271,19 @@ func (s *AutograderService) GetCoursesWithEnrollment(ctx context.Context, in *pb
 	return courses, nil
 }
 
+// GetEnrollmentsByUser returns all enrollments for the given user with preloaded courses and groups.
+// Access policy: user with userID or admin
+func (s *AutograderService) GetEnrollmentsByUser(ctx context.Context, in *pb.UserRequest) (*pb.Enrollments, error) {
+	// check that current user ID = user ID in request unless admin
```
meling: Is this a missing check?

public/src/pages/views/UserView.tsx
```diff
@@ -74,9 +75,9 @@ export class UserView extends React.Component<IUserViewerProps, IUserViewerState
         return heading;
     }
 
-    private renderRow(user: IUserRelation): (string | JSX.Element)[] {
-        const selector: (string | JSX.Element)[] = [];
-        if (user.link.getStatus() === Enrollment.UserStatus.TEACHER) {
+    private renderRow(user: IUserRelation): Array<string | JSX.Element> {
+        const selector: Array<string | JSX.Element> = [];
+        if (user.enrollment.getStatus() === Enrollment.UserStatus.TEACHER) {
```
meling: This could also use a method on `user.isTeacher()`

web/autograder_service.go
0xf8f8ff: Yes. Added now.

public/src/pages/views/EnrollmentView.tsx
```diff
@@ -13,21 +13,21 @@ export class EnrollmentView extends React.Component<IEnrollmentViewProps, {}> {
...
-    public createEnrollmentRow(course: IAssignmentLink):
-        (string | JSX.Element)[] {
-        const base: (string | JSX.Element)[] = [course.course.getCode(), course.course.getName()];
-        if (course.link) {
-            if (course.link.getStatus() === Enrollment.UserStatus.STUDENT
-                 || course.link.getStatus() === Enrollment.UserStatus.TEACHER) {
+    public createEnrollmentRow(course: IStudentLabsForCourse):
+        Array<string | JSX.Element> {
+        const base: Array<string | JSX.Element> = [course.course.getCode(), course.course.getName()];
+        if (course.enrollment) {
```
meling: These if-statements can be simplified into methods on `IStudentLabsForCourse`, e.g. `course.isStudent()` and `course.isTeacher()`... although I'm not too happy with the `course` naming here...

public/src/pages/StudentPage.tsx
```diff
@@ -302,7 +314,7 @@ export class StudentPage extends ViewPage {
     private selectGroupAssignment(labId: number) {
         if (this.selectedUserGroupCourse) {
             // TODO: Be carefull not to return anything that sould not be able to be returned
```
meling: Is this TODO useful? (I know it is probably old). Maybe it should be fixed or written as comment.

0xf8f8ff: I have no idea what it even refers to.

public/src/pages/StudentPage.tsx
```diff
@@ -293,7 +305,7 @@ export class StudentPage extends ViewPage {
     private selectAssignment(labId: number) {
```
meling: Maybe `selectAssignment` and `selectGroupAssignment` can be combined into one function.

public/src/pages/StudentPage.tsx
```diff
@@ -244,11 +247,11 @@ export class StudentPage extends ViewPage {
...
     }
 
-    private onlyActiveCourses(studentCourse: IAssignmentLink[]): IAssignmentLink[] {
-        const userCourses: IAssignmentLink[] = [];
+    private onlyActiveCourses(studentCourse: IStudentLabsForCourse[]): IStudentLabsForCourse[] {
+        const userCourses: IStudentLabsForCourse[] = [];
         studentCourse.forEach((a) => {
-            if (a.link && (a.link.getStatus() === Enrollment.UserStatus.STUDENT
-                || a.link.getStatus() === Enrollment.UserStatus.TEACHER)) {
+            if (a.enrollment && (a.enrollment.getStatus() === Enrollment.UserStatus.STUDENT
```
meling: Here the same method for `isStudent` and `isTeacher` can be reused.

public/src/managers/CourseManager.ts
```diff
@@ -378,10 +387,11 @@ export class CourseManager {
         if (assignments.length > 0) {
             const submissions =
                 await this.courseProvider.getAllGroupLabInfos(groupCourse.course.getId(), group.getId());
-
             for (const a of assignments) {
-                const submission = submissions.find((sub) => sub.assignmentid === a.getId());
-                groupCourse.assignments.push({ assignment: a, latest: submission, authorName: group.getName() });
+                if (a.getIsgrouplab()) {
```
meling: Should be `isGroupLab()`, unless this is generated code.

public/src/pages/StudentPage.tsx
0xf8f8ff: Easy.

public/src/pages/StudentPage.tsx
meling: Seems reasonable to remove these TODOs since these functions don't return anything...

public/src/pages/StudentPage.tsx
0xf8f8ff: I'm not sure adding methods to TS interfaces will make life easier, as those methods will have to be re-implemented at several places throughout the whole codebase (everywhere we currently can simply push an anonymous interface to an array).  
But it is also possible that I fail to see some better options. What if we just make it a separate task and I'll check all the options before adding more changes?

public/src/managers/CourseManager.ts
0xf8f8ff: It is a protobuf-defined class, you can only use getters here.

public/src/pages/StudentPage.tsx
0xf8f8ff: Already done.

public/src/components/forms/GroupForm.tsx
0xf8f8ff: Oh, right, got it now. Fixed.

public/src/pages/StudentPage.tsx
meling: Okay, I don't know enough TS to know this, but since this is a "closure" thing, couldn't there be a guard (instead of the if-stmt) to enter the anon function, which would then just be the `push(a)`. I will look for some examples of what I mean.

PR #224, opened 2020-03-05
Updating back-end dependencies.

0xf8f8ff:
This PR is temporarily in draft mode because it should not be applied before other PRs introducing changes to the `scm` package to avoid complications on merge.


web/scm_helpers.go
```diff
@@ -234,6 +234,35 @@ func removeUserFromCourse(ctx context.Context, sc scm.SCM, login string, repo *p
 	return sc.DeleteRepository(ctx, &scm.RepositoryOptions{ID: repo.GetRepositoryID()})
 }
 
+// remove user from teachers team, set organization status from owner to regular member
+func revokeTeacherStatus(ctx context.Context, sc scm.SCM, orgID uint64, userName string) error {
+	// TODO(vera): temp - remove this and pass the organization name directly when the relevant PR is applied
```
meling: Should this be fixed now?

web/auth/auth.go
```diff
@@ -9,8 +9,8 @@ import (
 	pb "github.com/autograde/aguis/ag"
 	"github.com/autograde/aguis/database"
 	"github.com/jinzhu/gorm"
-	"github.com/labstack/echo"
 	"github.com/labstack/echo-contrib/session"
+	echo "github.com/labstack/echo/v4"
```
meling: I don't think the `echo` alias is necessary; it is used in some files, but not in others.

meling: I mean, I think it will use `echo` as the package instead of `v4`...

web/scm_helpers.go
0xf8f8ff: Yes, updated it.

web/auth/auth.go
0xf8f8ff: Fixed.

scm/github.go
```diff
@@ -390,23 +390,23 @@ func (s *GithubSCM) DeleteTeam(ctx context.Context, opt *TeamOptions) error {
...
-	if _, err := s.client.Teams.DeleteTeam(ctx, int64(team.ID)); err != nil {
+	if err != nil {
 		return ErrFailedSCM{
 			Method:   "DeleteTeam",
 			Message:  fmt.Sprintf("failed to delete GitHub team '%s'", opt.TeamName),
 			GitError: fmt.Errorf("failed to get GitHub team '%s': %w", opt.TeamName, err),
 		}
 	}
-	return nil
+	return err
```
meling: This `err` will be nil anyway.

web/scm_helpers.go
```diff
@@ -234,6 +234,28 @@ func removeUserFromCourse(ctx context.Context, sc scm.SCM, login string, repo *p
 	return sc.DeleteRepository(ctx, &scm.RepositoryOptions{ID: repo.GetRepositoryID()})
 }
 
+// remove user from teachers team, set organization status from owner to regular member
+func revokeTeacherStatus(ctx context.Context, sc scm.SCM, org, userName string) error {
+
+	teamOpts := &scm.TeamMembershipOptions{
```
meling: Line break

web/scm_helpers.go
```diff
@@ -234,6 +234,28 @@ func removeUserFromCourse(ctx context.Context, sc scm.SCM, login string, repo *p
...
+		Organization: org,
+		TeamName:     scm.TeachersTeam,
+		Username:     userName,
+	}
+
+	err := sc.RemoveTeamMember(ctx, teamOpts)
```
meling: Why no err check?

PR #226, opened 2020-03-09
Studentpage fix.

0xf8f8ff:
The student page stopped showing course labs because of a missing check. Here is a fix for this issue.

### DIFF
--- a/public/src/managers/CourseManager.ts
+++ b/public/src/managers/CourseManager.ts
@@ -160,7 +160,7 @@ export class CourseManager {
             }
         }
         for (const link of links) {
-            await this.fillLinks(student, link);
+            await this.fillLinks(link, student);
         }
         return links;
     }
@@ -225,7 +225,7 @@ export class CourseManager {
                 labs: [],
                 course,
             };
-            await this.fillLinksGroup(group, groupCourse);
+            await this.fillLinks(groupCourse, undefined, group);
             return groupCourse;
         }
         return null;
@@ -289,7 +289,9 @@ export class CourseManager {
 
     /**
      * Generates a list with last submission build info
-     * for every course assignment for all course students
+     * for every course assignment for all course students.
+     * Used when displaying all course lab submissions for all
+     * course students on TeacherPage.
      */
     public async fillLabLinks(
         course: Course,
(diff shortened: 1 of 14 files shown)


public/src/managers/CourseManager.ts
```diff
@@ -338,7 +338,7 @@ export class CourseManager {
         if (!studentCourse.enrollment) {
             return;
         }
-        if (!assignments) {
+        if (!assignments || assignments?.length < 1) {
```
meling: I don't think you need to protect the second clause with a nil check, since the first clause already does that and will short-circuit the second clause. But it's fine to be safe. Anyway, maybe you should apply a similar fix to `fillLinksGroup`.

0xf8f8ff: Right. Fixed it.

public/src/managers/CourseManager.ts
```diff
@@ -332,47 +334,39 @@ export class CourseManager {
...
-            const submissions =
-                await this.courseProvider.getAllGroupLabInfos(groupCourse.course.getId(), group.getId());
-            for (const a of assignments) {
-                if (a.getIsgrouplab()) {
-                    const submission = submissions.find((sub) => sub.assignmentid === a.getId());
-                    groupCourse.labs.push({ assignment: a, submission: submission, authorName: group.getName() });
-                }
+
+        for (const a of assignments) {
+            if (!a.getIsgrouplab()) {
```
meling: What happens for group labs then?

public/src/managers/CourseManager.ts
```diff
@@ -332,47 +334,39 @@ export class CourseManager {
...
+        let labAuthorName = "";
+        if (student) {
+            submissions =
+                await this.courseProvider.getAllLabInfos(courseLabLink.course.getId(), student.getId());
+            labAuthorName = student.getName();
         }
-        if (!assignments) {
-            assignments = await this.getAssignments(groupCourse.course.getId());
+
+        if (group) {
```
meling: Should this be?:
```typescript
if (student) {
 ...
} else if (group) {
   submissions = await ....
   ...
} else {
   return;
}
```

public/src/managers/CourseManager.ts
0xf8f8ff: Added a Boolean to fix that.

public/src/managers/CourseManager.ts
0xf8f8ff: Well, we check already that it's either or, but can't be too careful. Changed.

PR #238, opened 2020-03-23
GitHub api update.

0xf8f8ff:
- incorporating changes introduced in the new version of `go-github` package
- some minor codebase changes: new GitHub option struct and updated validation methods for such structs
- updated makefile target to run all project tests at once


scm/github.go
```diff
@@ -670,7 +659,7 @@ func (s *GithubSCM) RemoveMember(ctx context.Context, opt *OrgMembershipOptions)
 func (s *GithubSCM) GetUserScopes(ctx context.Context) *Authorization {
 	// Authorizations.List method will always return nil, response struct and error,
```
meling: Comment not update; seems to describe only method.

Makefile
```diff
@@ -94,7 +94,11 @@ protoset:
 
 test:
 	@cd ./web; go test
+	@cd ./web/hooks; go test
```
meling: I think these can all be replaced with a single line:
```
go test ./...
```
Or
```
go test -v ./...
```

scm/github.go
```diff
@@ -466,27 +465,23 @@ func (s *GithubSCM) AddTeamMember(ctx context.Context, opt *TeamMembershipOption
...
 
-	if _, _, err := s.client.Teams.AddTeamMembership(ctx, int64(opt.TeamID), opt.Username,
-		&github.TeamAddTeamMembershipOptions{Role: opt.Role}); err != nil {
-		return ErrFailedSCM{
+	if err != nil {
+		err = ErrFailedSCM{
 			GitError: err,
 			Method:   "AddTeamMember",
-			Message:  fmt.Sprintf("failed to add user %s to team ID %d with role %s", opt.Username, opt.TeamID, opt.Role),
+			Message:  fmt.Sprintf("failed to add user username: %s to team (ID %d, team name: %s) with role %s", opt.Username, opt.TeamID, opt.TeamName, opt.Role),
```
meling: user username sounds weird. I like the old error msg, maybe combined with the new one. Maybe add single quote or parens around the name, e.g. `... add user (meling) to team ...`. That is, `... user (%s) to team (ID %d, ...`

scm/github.go
0xf8f8ff: Fixed.

Makefile
0xf8f8ff: Yeah, -v is just too vocal. Changed to `go test ./...`.

scm/github.go
0xf8f8ff: Fixed.

PR #239, opened 2020-03-24
Fixing MemberView

0xf8f8ff:
This PR addresses three issues:
- when promoting to teacher/demoting a teacher resulted in error, the status of the user would still be changed locally;
- when approving a new group it would be temporarily marked as approved even if the method failed;
- when approving or rejecting a user enrollment, lists of the approved and pending students would not be updated without a manual page reload.

### DIFF
--- a/public/src/components/teacher/CourseGroup.tsx
+++ b/public/src/components/teacher/CourseGroup.tsx
@@ -192,10 +192,11 @@ export class CourseGroup extends React.Component<ICourseGroupProps, ICourseGroup
     private async handleActionOnClick(group: Group, link: ILink): Promise<void> {
         switch (link.uri) {
             case "approve":
-                group.setStatus(Group.GroupStatus.APPROVED);
                 const ans = await this.props.courseMan.updateGroup(group);
                 if (ans.getCode() !== 0) {
-                this.generateErrorMessage(ans);
+                    this.generateErrorMessage(ans);
+                } else {
+                    group.setStatus(Group.GroupStatus.APPROVED);
                 }
                 break;
             case "edit":
--- a/public/src/components/teacher/CourseGroup.tsx
+++ b/public/src/components/teacher/CourseGroup.tsx
@@ -307,7 +307,6 @@ export class CourseGroup extends React.Component<ICourseGroupProps, ICourseGroup
         this.setState({
             approvedGroups: this.props.approvedGroups,
             pendingGroups: this.props.pendingGroups,
-            errorMsg: null,
         });
         return this.forceUpdate();
     }
--- a/public/src/components/teacher/CourseGroup.tsx
+++ b/public/src/components/teacher/CourseGroup.tsx
@@ -140,7 +140,7 @@ export class CourseGroup extends React.Component<ICourseGroupProps, ICourseGroup
                 id="edit"
                 className="btn btn-success member-btn"
                 onClick={() => this.toggleEditState()}
-    >{this.editButtonString()}</button>;
+        >{this.editButtonString()}</button>;
     }
 
     private getMembers(users: User[]): JSX.Element {
@@ -291,7 +291,7 @@ export class CourseGroup extends React.Component<ICourseGroupProps, ICourseGroup
         const err = <div className="alert alert-danger">{status.getError()}</div>;
         this.setState({
                 errorMsg: err,
-            });
+        });
     }
 
     private checkForErrors(status: Status, action?: () => void) {
@@ -311,5 +311,4 @@ export class CourseGroup extends React.Component<ICourseGroupProps, ICourseGroup
         });
         return this.forceUpdate();
     }
-
 }
--- a/public/src/components/teacher/CourseGroup.tsx
+++ b/public/src/components/teacher/CourseGroup.tsx
@@ -160,9 +160,7 @@ export class CourseGroup extends React.Component<ICourseGroupProps, ICourseGroup
 
     private async handleUpdateStatus(group: Group, status: Group.GroupStatus): Promise<void> {
         const ans = await this.props.courseMan.updateGroupStatus(group.getId(), status);
-        if (ans.getCode() !== 0) {
-            this.generateErrorMessage(ans);
-        }
+        this.checkForErrors(ans);
     }
 
     private async deleteGroup(group: Group) {
@@ -183,9 +181,7 @@ export class CourseGroup extends React.Component<ICourseGroupProps, ICourseGroup
 
         if (readyToDelete) {
             const ans = await this.props.courseMan.deleteGroup(courseID, group.getId());
-            if (ans.getCode() !== 0) {
-                this.generateErrorMessage(ans);
-            }
+            this.checkForErrors(ans);
         }
     }
 
@@ -193,11 +189,9 @@ export class CourseGroup extends React.Component<ICourseGroupProps, ICourseGroup
         switch (link.uri) {
             case "approve":
                 const ans = await this.props.courseMan.updateGroup(group);
-                if (ans.getCode() !== 0) {
-                    this.generateErrorMessage(ans);
-                } else {
+                this.checkForErrors(ans, () => {
                     group.setStatus(Group.GroupStatus.APPROVED);
-                }
+                })
                 break;
             case "edit":
                 this.props.navMan
@@ -300,10 +294,20 @@ export class CourseGroup extends React.Component<ICourseGroupProps, ICourseGroup
             });
     }
 
+    private checkForErrors(status: Status, action?: () => void) {
+        if (status.getCode() !== 0) {
+            this.generateErrorMessage(status);
+            return;
+        } else if (action) {
+            action();
+        }
+    }
+
     private refreshState() {
         this.setState({
             approvedGroups: this.props.approvedGroups,
             pendingGroups: this.props.pendingGroups,
+            errorMsg: null,
         });
         return this.forceUpdate();
     }
--- a/public/src/components/teacher/CourseGroup.tsx
+++ b/public/src/components/teacher/CourseGroup.tsx
@@ -139,7 +139,7 @@ export class CourseGroup extends React.Component<ICourseGroupProps, ICourseGroup
         return <button type="button"
                 id="edit"
                 className="btn btn-success member-btn"
-                onClick={() => this.flipEditState()}
+                onClick={() => this.toggleEditState()}
     >{this.editButtonString()}</button>;
     }
 
@@ -283,7 +283,7 @@ export class CourseGroup extends React.Component<ICourseGroupProps, ICourseGroup
         return links;
     }
 
-    private flipEditState() {
+    private toggleEditState() {
         this.setState({
             editing: !this.state.editing,
         }, () => this.refreshState());
--- a/public/src/managers/CourseManager.ts
+++ b/public/src/managers/CourseManager.ts
@@ -5,18 +5,18 @@ import {
     IUserRelation,
 } from "../models";
 
-import { Assignment, Course, Enrollment, Group, Organization, Repository, Status, User, Void } from "../../proto/ag_pb";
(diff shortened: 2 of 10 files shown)


public/src/managers/ServerProvider.ts
```diff
@@ -112,8 +112,12 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
     }
 
     public async changeUserState(link: Enrollment, state: Enrollment.UserStatus): Promise<Status> {
+        const originalState = link.getStatus();
```
meling: Why are these variables called `link` and `state`? (I assume it is legacy reasons)
Would it be reasonable to name these as `enrollment` and `enrollStatus` or something along those lines?

public/src/pages/views/MemberView.tsx
```diff
@@ -108,30 +108,34 @@ export class MemberView extends React.Component<IUserViewerProps, IUserViewerSta
...
             case "teacher":
-                this.handlePromote(userRel);
+                await this.handlePromote(userRel);
                 break;
             case "demote":
-                this.handleDemote(userRel);
+                await this.handleDemote(userRel);
                 break;
         }
+        this.flipEditState();
```
meling: What about `toggleEditState()`?

public/src/components/teacher/CourseGroup.tsx
```diff
@@ -192,10 +192,11 @@ export class CourseGroup extends React.Component<ICourseGroupProps, ICourseGroup
     private async handleActionOnClick(group: Group, link: ILink): Promise<void> {
         switch (link.uri) {
             case "approve":
-                group.setStatus(Group.GroupStatus.APPROVED);
                 const ans = await this.props.courseMan.updateGroup(group);
                 if (ans.getCode() !== 0) {
-                this.generateErrorMessage(ans);
+                    this.generateErrorMessage(ans);
```
meling: I see this pattern a lot:
```typescript
if (ans.getCode() !== 0) {
    this.generateErrorMessage(ans);
} else {
   // do something different each time the pattern is used...
}
```
Would it make sense to define a closure/function for this, to which you pass `ans` and the code in the else-clause?

public/src/managers/ServerProvider.ts
0xf8f8ff: Yeah, absolutely. These were indeed the legacy names for those fields.

public/src/pages/views/MemberView.tsx
0xf8f8ff: Sure.

public/src/components/teacher/CourseGroup.tsx
0xf8f8ff: Yeah, that sounds like a good idea.

public/src/managers/ServerProvider.ts
```diff
@@ -81,114 +81,116 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
...
         return result.status;
     }
 
     public async approveAll(courseID: number): Promise<boolean> {
         const result = await this.grpcHelper.updateEnrollments(courseID);
-        if (result.status.getCode() !== 0 || !result.data) {
-            return false;
-        }
-        return true;
+        return result.data ? this.responseCodeSuccess(result) : false;
```
meling: This pattern is different from the others below. Suggest to harmonize these. See more below.

public/src/managers/ServerProvider.ts
```diff
@@ -455,4 +451,8 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
         };
         return isbm;
     }
+
+    private responseCodeSuccess(response: IGrpcResponse<any>): boolean {
+        return response.status.getCode() === 0;
+    }
```
meling: This can also check the `response.data` field as part of the check. That way, we can skip many of the checks above.
```typescript
    return response.status.getCode() === 0 && result.data;
```
(maybe the method name needs to be changed too, but that's fine.)

public/src/managers/ServerProvider.ts
```diff
@@ -375,23 +374,20 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
...
-            return false;
-        }
-        return true;
+        return this.responseCodeSuccess(result);
     }
 
     public async rebuildSubmission(assignmentID: number, submissionID: number): Promise<ISubmission | null> {
         const result = await this.grpcHelper.rebuildSubmission(assignmentID, submissionID);
-        if (result.status.getCode() !== 0 || !result.data) {
+        if (!this.responseCodeSuccess(result) || !result.data) {
```
meling: With an updated method that also checks `result.data`, this pattern becomes:
```typescript
if (this.responseSuccess(result)) {
    return this.toISubmission(result.data);
}
return null;
```
(and similar for almost all the uses above.)

public/src/managers/ServerProvider.ts
```diff
@@ -375,23 +374,20 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
 
     public async updateSubmission(courseID: number, submissionID: number, approve: boolean): Promise<boolean> {
         const result = await this.grpcHelper.updateSubmission(courseID, submissionID, approve);
-        if (result.status.getCode() !== 0) {
-            return false;
-        }
-        return true;
+        return this.responseCodeSuccess(result);
```
meling: I'm not sure if the `result.data` field is non-null for this case, when response code is success, but probably.

public/src/managers/ServerProvider.ts
```diff
@@ -309,28 +311,25 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
...
     }
 
     public async updateUser(user: User): Promise<boolean> {
         const result = await this.grpcHelper.updateUser(user);
-        if (result.status.getCode() !== 0 || !result.data) {
-            return false;
-        }
-        return true;
+        return result.data ? this.responseCodeSuccess(result) : false;
```
meling: This pattern matches my proposed change of the `responseCodeSuccess`. This would become:
```typescript
return this.isSuccess(result);
```
Naming suggestions:
`isSuccess` or `responseSuccess` or something better...

public/src/managers/ServerProvider.ts
```diff
@@ -233,7 +235,7 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
 
     public async getCourseLabs(courseID: number, groupLabs: boolean): Promise<IStudentLabsForCourse[]> {
         const result = await this.grpcHelper.getCourseLabSubmissions(courseID, groupLabs);
-        if (result.status.getCode() !== 0 || !result.data) {
+        if (!this.responseCodeSuccess(result) || !result.data) {
```
meling: This pattern becomes:
```typescript
if (!this.responseSuccess(result)) {
    return [];
}
...
```

meling: PS: These OR checks are probably not exactly correct, because they require that both the success code is non-zero and that the `result.data` to be null at the same time. I feels more correct that if the success code is non-zero (failure), then the condition should "fail", and that if the success was a "success" then if there is no data, it is also a fail. But I may be wrong. And it may not matter as well...

public/src/managers/ServerProvider.ts
```diff
@@ -81,114 +81,116 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
...
             return null;
         }
         return result.data;
     }
 
-    public async updateGroupStatus(groupID: number, states: Group.GroupStatus): Promise<Status> {
-        const result = await this.grpcHelper.updateGroupStatus(groupID, states);
+    public async updateGroupStatus(groupID: number, status: Group.GroupStatus): Promise<Status> {
+        const result = await this.grpcHelper.updateGroupStatus(groupID, status);
```
meling: Shouldn't this also check the status code?

public/src/managers/ServerProvider.ts
```diff
@@ -81,114 +81,116 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
...
     public async isAuthorizedTeacher(): Promise<boolean> {
         const result = await this.grpcHelper.isAuthorizedTeacher();
-        if (result.status.getCode() !== 0 || !result.data) {
-            return false;
+        if (this.responseCodeSuccess(result) && result.data) {
+            return result.data.getIsauthorized();
         }
-        return result.data.getIsauthorized();
+        return false;
+
```
meling: Ref to comment below with an updated method that also checks `result.data`, this pattern becomes:
```typescript
if (this.responseSuccess(result)) {
    return result.data.getIsauthorized();
}
return false;
```

public/src/managers/ServerProvider.ts
meling:
```typescript
return this.isSuccess(result);
```
(note `isSuccess()` is only one of several names that I mix and match for the same method...)

public/src/managers/CourseManager.ts
```diff
@@ -108,8 +108,8 @@ export class CourseManager {
     /**
      * Update status of a course enrollment
      */
-    public async changeUserState(link: Enrollment, state: Enrollment.UserStatus): Promise<Status> {
-        const ans = await this.courseProvider.changeUserState(link, state);
+    public async changeUserStatus(enrollment: Enrollment, status: Enrollment.UserStatus): Promise<Status> {
+        const ans = await this.courseProvider.changeUserStatus(enrollment, status);
```
meling: Can return directly without `const ans`.

public/src/managers/ServerProvider.ts
meling: So I think a rewritten method can still be used here.

public/src/managers/ServerProvider.ts
0xf8f8ff: Unfortunately, it cannot (this was my first approach as well). The calling method will still refuse to believe that `response.data` is present if checked like this, so it won't compile. But I agree, there has to be a more elegant solution.

public/src/managers/ServerProvider.ts
0xf8f8ff: The gRPC method here returns `Void`, should we care whether it is here or not? Status code will be 0 if the request returned without errors and we check exactly that.

public/src/managers/ServerProvider.ts
0xf8f8ff: We don't want to filter out failed requests here, they carry error messages to show to the user.

public/src/managers/ServerProvider.ts
meling: This is fine to just check the status code.

public/src/managers/ServerProvider.ts
meling: okay, maybe add a comment about that.

public/src/managers/CourseManager.ts
0xf8f8ff: Right. Fixed.

public/src/managers/ServerProvider.ts
0xf8f8ff: Honestly, it looks like making sure that both status is `0` and `data` field is present is pretty redundant: there will be no data if the status is not 0, and there could not be any other status code then 0 if the data is here.

The only reason we check now is because the `data` field is optional and TS with its strictness will not allow working with it unless we check explicitly that the field is set.

It needs some testing, but for now it looks reasonable to change the `data` field from optional to required, and only look at the status code.

meling: All fields in proto3 are required.

0xf8f8ff: OK, making the field required does not really look like a good idea because of the generic.

0xf8f8ff: `IGrpcResponse<T>` is not a Protobuf type. The `data` field in question belongs to it.

meling:
I don't have better ideas on how to resolve those outstanding simplification ideas, so I'll just merge.

PR #242, opened 2020-03-26
API updates

0xf8f8ff:
- more consistent naming for API methods and request types
- fixed a few minor bugs and updated comments
- disentangled front-end methods fetching courses and enrollments
- returning Void instead of unused structs where applicable


ag/ag.proto
```diff
@@ -203,12 +197,19 @@ message OrgRequest {
     string orgName = 1;
 }
 
-// depending on situation frontend will have all enrolled users or only those that are not members of any group
-// set filterOutGroupMembers go get list of students not belonging to any group for given course
+// Depending on situation front-end will display all the enrolled users or only those that are not members of any group.
```
meling: What about this:
_EnrollmentRequest contains information about a course and a list of enrolled users. Currently, the list of enrolled users may be all users enrolled in the course, or alternatively those that are not members of any group. The ignoreGroupMembers field indicates that the list of enrolled users should not belong to any group for the given course._

Note `ignoreGroupMembers` instead of `filterOutGroupMembers`.

ag/ag.proto
```diff
@@ -203,12 +197,19 @@ message OrgRequest {
...
 message EnrollmentRequest {
     uint64 courseID = 1;
     bool filterOutGroupMembers = 2;
-    repeated Enrollment.UserStatus states = 3;
+    repeated Enrollment.UserStatus statuses = 3;
+}
+
+// Used to request courses and enrollments for a given user 
+// where the user has a specific enrollment status.
+message EnrollmentStatusRequest {
```
meling: What about:
_EnrollmentStatusRequest is a request for a given user, with a specific enrollment status._

(we should avoid explaining what these message types are used for, at least as the first sentence. Maybe a use case example can be added at the end. Rationale: Some of these types may, in the future, be reused for something more.)

ag/ag.proto
```diff
@@ -259,9 +260,9 @@ message LabRequest {
     bool groupLabs = 4;
 }
 
-// messages of type Void used when we want a boolean answer from the server
+// Messages of type Void used when we want a boolean answer from the server
 // GRPC status code 0 indicates that the requested action was a success
-// any other status will include a message explaining what went wrong
+// any other status will include a message explaining what went wrong.
 message Void {}
```
meling: What about:
_Void contains no fields. Void still contains a gRPC status code, which can be checked for success or failure. Status code 0 indicates that the requested action was successful, whereas any other status code indicates some failure. As such, the status code can be used as a boolean result from the server._

public/src/managers/CourseManager.ts
```diff
@@ -353,11 +345,11 @@ export class CourseManager {
         let wantGroupLinks = false;
         if (student) {
             submissions =
-                await this.courseProvider.getAllLabInfos(courseLabLink.course.getId(), student.getId());
+                await this.courseProvider.getLabsForStudent(courseLabLink.course.getId(), student.getId());
```
meling, line 348: We should unify naming. That is, we should not refer to any `Lab`, and instead use `Assignment`

I'm also not clear about these legacy link objects, e.g. `courseLabLink` and `wantGroupLink`... I think what is being referred to is that they are the associations between two other tables to resolve a many-to-many relationship.

ag/ag.proto
0xf8f8ff: But EnrollmentRequest does not contain any lists? It's a request sent to the server with course ID and a pair of options.

Renamed the field.

ag/ag.proto
0xf8f8ff: Yeah, I see your point. Fixed.

ag/ag.proto
0xf8f8ff: Done.

ag/ag.proto
meling: I don't see the fix.

ag/ag.proto
meling: Sorry, my misunderstanding. But the doc comment should still be more clear and not make case-specific references as the first order of business (as pointed out in another comment from yesterday below); maybe say something along the lines:

_EnrollmentRequest is a request for enrolled users of a given course, whose enrollment status match those provided in the request. To ignore group members that otherwise match the enrollment request, set ignoreGroupMembers to true._

ag/ag.proto
0xf8f8ff: Just removed the comment. It was supposed to clarify that the same request can be used when fetching courses or enrollments. Otherwise there is no need to explain obvious things. Or maybe the opposite? We could add similar comments for every request type.

ag/ag.proto
0xf8f8ff: Yeah, this sounds reasonable. Changed.

ag/ag.proto
meling: Well, I don't think we need a comment on everything, at this point at least, but, since we already started adding a comment on this one, I proposed this one as a replacement:
_EnrollmentStatusRequest is a request for a given user, with a specific enrollment status._

(is this not accurate?)

0xf8f8ff: Yeah, it is pretty accurate. Added the proposed comment.

public/src/managers/CourseManager.ts
meling: @0xf8f8ff Did you see this?

0xf8f8ff: Yeah, I agree we should replace all "labs" with either "assignment" or "submission" for consistency.
The legacy link objects can probably be replaced by the protobuf defined types more easily now with the last TS changes. I will try to replace them in a separate PR.

0xf8f8ff: Should we also change LabRequest, isGroupLab and similar examples?

meling: Yeah probably. But it's okay to do it as a separate PR, or postpone it to be sure that we get time to implement the important stuff. Let me know, and I can merge this PR, and we continue with other stuff.

PR #250, opened 2020-04-02
Replacing models with protobuf-defined types

0xf8f8ff:
- replaced `IUserRelation` model with `Enrollment`

### DIFF
--- a/public/src/components/forms/GroupForm.tsx
+++ b/public/src/components/forms/GroupForm.tsx
@@ -5,12 +5,11 @@ import { Search } from "../../components";
 import { CourseManager } from "../../managers/CourseManager";
 import { NavigationManager } from "../../managers/NavigationManager";
 import { UserManager } from "../../managers/UserManager";
-import { IUserRelation } from "../../models";
 
 interface IGroupProps {
     className: string;
-    students: IUserRelation[];
-    freeStudents: IUserRelation[];
+    students: Enrollment[];
+    freeStudents: Enrollment[];
     curUser: User;
     courseMan: CourseManager;
     userMan: UserManager;
@@ -21,9 +20,9 @@ interface IGroupProps {
 }
 interface IGroupState {
     name: string;
-    students: IUserRelation[];
-    selectedStudents: IUserRelation[];
-    curUser: IUserRelation | undefined;
+    students: Enrollment[];
+    selectedStudents: Enrollment[];
+    curUser: Enrollment | undefined;
     errorFlash: JSX.Element | null;
     actionReady: boolean;
 }
@@ -31,9 +30,9 @@ interface IGroupState {
 export class GroupForm extends React.Component<IGroupProps, IGroupState> {
     constructor(props: any) {
         super(props);
-        const currentUser = this.props.students.find((v) => v.user.getId() === this.props.curUser.getId());
-        const as: IUserRelation[] = this.getAvailableStudents();
-        const ss: IUserRelation[] = this.getSelectedStudents(currentUser);
+        const currentUser = this.props.students.find((v) => v.getUserid() === this.props.curUser.getId());
+        const as: Enrollment[] = this.getAvailableStudents();
+        const ss: Enrollment[] = this.getSelectedStudents(currentUser);
         this.state = {
             name: this.props.groupData ? this.props.groupData.getName() : "",
             students: as,
@@ -52,8 +51,8 @@ export class GroupForm extends React.Component<IGroupProps, IGroupState> {
         const selectableStudents: JSX.Element[] = [];
         for (const student of this.state.students) {
             selectableStudents.push(
-                <li key={student.user.getId()} className="box-item">
-                    <label>{student.user.getName()}</label>
+                <li key={student.getUserid()} className="box-item">
+                    <label>{student.getUser()?.getName()}</label>
                     <button type="button"
                         className="btn btn-outline-success add-btn"
                          onClick={() => this.handleAddToGroupOnClick(student)}>
@@ -65,12 +64,12 @@ export class GroupForm extends React.Component<IGroupProps, IGroupState> {
         const selectedStudents: JSX.Element[] = [];
         for (const student of this.state.selectedStudents) {
             selectedStudents.push(
-                <li key={student.user.getId()} className="box-item">
+                <li key={student.getUserid()} className="box-item">
                     <button className="btn btn-outline-primary rm-btn"
                         onClick={() => this.handleRemoveFromGroupOnClick(student)}>
                         <i className="glyphicon glyphicon-minus-sign" />
                     </button>
-                    <label>{student.user.getName()}</label>
+                    <label>{student.getUser()?.getName()}</label>
                 </li>);
         }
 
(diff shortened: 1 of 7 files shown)


meling:
Seems this PR has a compile error:
```
ERROR in [at-loader] ./src/managers/GRPCManager.ts:130:17
    TS2339: Property 'setFilteroutgroupmembers' does not exist on type 'EnrollmentRequest'.
```

public/src/components/forms/GroupForm.tsx
```diff
@@ -278,22 +278,22 @@ export class GroupForm extends React.Component<IGroupProps, IGroupState> {
...
+            if (user.getUserid() === student.getUserid()) {
                 foundSelected = true;
             }
         })
         return foundSelected;
     }
 
-    private getSelectedStudents(curUser: IUserRelation | undefined): IUserRelation[] {
-        const ss: IUserRelation[] = [];
+    private getSelectedStudents(curUser: Enrollment | undefined): Enrollment[] {
```
meling, line 291: The `curUser` argument is not used and can be removed.

public/src/pages/views/MemberView.tsx
```diff
@@ -78,18 +77,18 @@ export class MemberView extends React.Component<IUserViewerProps, IUserViewerSta
                 this.state.pendingUsers,
                 [],
                 ActionType.InRow,
-                (user: IUserRelation) => {
-                    return this.generateUserButtons(user.enrollment);
+                (user: Enrollment) => {
```
meling, line 80: Rename `user` to `enrollment`.

public/src/pages/views/MemberView.tsx
```diff
@@ -78,18 +77,18 @@ export class MemberView extends React.Component<IUserViewerProps, IUserViewerSta
...
+                    return this.generateUserButtons(user);
                 });
         }
     }
 
     public renderUsers(
         title: string | JSX.Element,
-        users: IUserRelation[],
+        users: Enrollment[],
```
meling, line 88: users -> enrollments

public/src/pages/views/MemberView.tsx
```diff
@@ -78,18 +77,18 @@ export class MemberView extends React.Component<IUserViewerProps, IUserViewerSta
...
         actions?: ILink[],
         linkType?: ActionType,
-        optionalActions?: ((user: IUserRelation) => ILink[])) {
+        optionalActions?: ((user: Enrollment) => ILink[])) {
```
meling, line 91: user --> enrollment

public/src/pages/views/UserView.tsx
```diff
@@ -1,23 +1,22 @@
...
     userMan?: UserManager;
     navMan?: NavigationManager;
     courseURL: string;
     searchable?: boolean;
     actions?: ILink[];
-    optionalActions?: (user: IUserRelation) => ILink[];
+    optionalActions?: (enrol: Enrollment) => ILink[];
     linkType?: ActionType;
-    actionClick?: (user: IUserRelation, link: ILink) => void;
+    actionClick?: (user: Enrollment, link: ILink) => void;
```
meling, line 19: user --> enrollment

public/src/pages/views/UserView.tsx
```diff
@@ -27,7 +26,7 @@ export enum ActionType {
 }
 
 interface IUserViewerState {
-    users: IUserRelation[];
+    users: Enrollment[];
```
meling, line 29: users --> enrollments

public/src/pages/views/UserView.tsx
```diff
@@ -99,43 +99,43 @@ export class UserView extends React.Component<IUserViewerProps, IUserViewerState
...
-        const tempActions = this.getAllLinks(user);
+        const tempActions = this.getAllLinks(enrol);
         if (tempActions.length > 0) {
             switch (this.props.linkType) {
                 case ActionType.Menu:
-                    return this.renderDropdownMenu(user, tempActions);
+                    return this.renderDropdownMenu(enrol, tempActions);
                 case ActionType.InRow:
-                    actionButtons.push(...this.renderActionRow(user, tempActions));
+                    actionButtons.push(...this.renderActionRow(enrol, tempActions));
```
meling, line 110: Why do we have `break` below here?

public/src/pages/views/UserView.tsx
```diff
@@ -75,22 +74,23 @@ export class UserView extends React.Component<IUserViewerProps, IUserViewerState
         return heading;
     }
 
-    private renderRow(user: IUserRelation): (string | JSX.Element)[] {
+    private renderRow(enr: Enrollment): (string | JSX.Element)[] {
         const selector: (string | JSX.Element)[] = [];
-        if (user.enrollment.getStatus() === Enrollment.UserStatus.TEACHER) {
+        const user = enr.getUser();
+        if (user && enr.getStatus() === Enrollment.UserStatus.TEACHER) {
```
meling, line 80: This code can be further simplified by checking for `!user` and returning an empty selector early (which should probably never happen, right??). After which you don't need to check for `user` again, and you can also avoid the `end.getUser()?` checks below. Like this:

```typescript
        if (!user) {
            return selector;
        }
        if (enr.getStatus() === Enrollment.UserStatus.TEACHER) {
            selector.push(
                <span className="text-muted">
                    <a href={this.gitLink(user.getLogin())} target="_blank">{user.getName()}</a>
                </span>);
        } else {
            selector.push(
                <a href={this.repoLink(user.getLogin())} target="_blank">{user.getName()}</a>);
        }
        selector.push(
            <a href={"mailto:" + user.getEmail()}>{user.getEmail()}</a>, user.getStudentid() ?? "",
        );
```

public/src/components/forms/GroupForm.tsx
```diff
@@ -224,11 +223,12 @@ export class GroupForm extends React.Component<IGroupProps, IGroupState> {
 
     private handleSearch(query: string): void {
         query = query.toLowerCase();
-        const filteredData: IUserRelation[] = [];
+        const filteredData: Enrollment[] = [];
         this.props.freeStudents.forEach((student) => {
-            if ((student.user.getName().toLowerCase().indexOf(query) !== -1
-                || student.user.getEmail().indexOf(query) !== -1
-                || student.user.getLogin().indexOf(query)) !== -1
+            const enr = student.toObject().user;
```
meling, line 228: It is confusing that you use `student` for `Enrollment` type and `enr` for `User` type. I guess also `freeStudents` is a bit confusing, since that's really a list of enrollments. Maybe `availableEnrollments` is a good name?

public/src/pages/views/UserView.tsx
0xf8f8ff: I guess it is because of the switch statement. Not sure if we really need it here though.

public/src/components/forms/GroupForm.tsx
0xf8f8ff: Well, now we work with enrollments here, but we still kinda add users/students (not enrollments) to groups. So naming is still based on the logical UI components, and not underlying structures. Easy to change if you think it makes more sense.
Totally agree about `enr` for `User`, fixing it.

public/src/components/forms/GroupForm.tsx
0xf8f8ff: Fixed.

public/src/pages/views/MemberView.tsx
0xf8f8ff: Renamed here, renamed in all other cases mentioned below as well.

public/src/pages/views/UserView.tsx
0xf8f8ff: Changed.

PR #251, opened 2020-04-03
Course visibility

0xf8f8ff:
Users get to choose which courses to be displayed on the sidebar.

Only non-archived courses show on left side

Editing state


public/src/componentHelper.tsx
```diff
@@ -0,0 +1,158 @@
...
+            case Enrollment.DisplayState.UNSET:
+                active.push(enrol);
+                break;
+        }
+    })
+    sorted = sorted.concat(active, archived);
+    return sorted;
+}
+
+export function sortUsersByAdminStatus(users: Enrollment[]): Enrollment[] {
```
meling, line 32: Nice... good place to collect these funcs

public/src/index.tsx
```diff
@@ -198,21 +196,17 @@ class AutoGrader extends React.Component<IAutoGraderProps, IAutoGraderState> {
...
         topArea = await this.renderActiveMenu(page, 1);
-        // }
 
         const loginLink: ILink[] = [
             { name: "Github", uri: "app/login/login/github" },
             { name: "Gitlab", uri: "app/login/login/gitlab" },
         ];
         const dropDownMenuLinks: ILink[] = [
             { name: "Join Course", uri: "app/student/enroll" },
+            { name: "Change course visibility", uri: "app/student/archive"},
```
meling, line 209: What about “Course List”... or something

public/src/pages/StudentPage.tsx
```diff
@@ -86,6 +89,23 @@ export class StudentPage extends ViewPage {
...
     }
 
+    public async archive(navInfo: INavInfo<any>): View {
+        await this.setupData();
+        const curUser = this.userMan.getCurrentUser();
+        if (!curUser) {
+            return showLoader();
+        }
+        return <div>
+            <h1>Course visibility</h1>
```
meling, line 99: Course List

public/src/pages/TeacherPage.tsx
```diff
@@ -348,32 +350,25 @@ export class TeacherPage extends ViewPage {
...
-            );
-        const fullRepoName = repoMap.get(Repository.Type.COURSEINFO);
-        return fullRepoName ? fullRepoName.split("course-info")[0] : "";
+        const repoMap = this.repositories.get(courseID);
+        return repoMap?.get(Repository.Type.COURSEINFO) ?? "";
     }
 
     private async courseFunc(courseParam: string, fn: (course: Course) => View): View {
         const courseId = parseInt(courseParam, 10);
-        const course = await this.courseMan.getCourse(courseId);
```
meling, line 370: `courseId` on line 369 is no longer needed

meling: And similarly for `courseParam`

meling: I also don’t think you need the function parameter name `course` in `(course: Course)`

public/src/pages/views/CourseView.tsx
```diff
@@ -44,18 +45,9 @@ export class CourseView extends React.Component<ICourseViewProps, ICourseViewSta
...
-                || course.getCode().toLowerCase().indexOf(query) !== -1
-                || course.getYear().toString().indexOf(query) !== -1
-            ) {
-                filteredData.push(course);
-            }
-        });
+        const filteredData = searchForCourses(this.props.courses, query);
         this.setState({
-            courses: filteredData,
+            courses: filteredData as Course[],
```
meling, line 50: Can’t you just call `searchForCourses` directly here.

public/src/components/teacher/Results.tsx
```diff
@@ -134,21 +134,8 @@ export class Results extends React.Component<IResultsProp, IResultsState> {
...
-                    || usr.getEmail().toLowerCase().indexOf(query) !== -1
-                    || usr.getLogin().toLowerCase().indexOf(query) !== -1
-                ) {
-                    filteredData.push(std);
-                }
-            }
-        });
-
+    private handleSearch(query: string): void {
+        const filteredData = searchForLabs(this.props.students, query);
```
meling, line 138: Can be called inline in setState call below

public/src/pages/StudentPage.tsx
```diff
@@ -184,58 +204,60 @@ export class StudentPage extends ViewPage {
...
-                });
-
-                allLinks.push({
-                    name: "New Group", uri: this.pagePath + "/courses/" + courseID + "/members",
-                });
-                coursesLinks.push({
-                    item: { name: course.course.getCode(), uri: this.pagePath + "/courses/" + courseID },
-                    children: allLinks,
-                });
+                if (course.enrollment.getState() !== Enrollment.DisplayState.ARCHIVED) {
```
meling, line 207: This code is a bit long; should be split into one or two methods.
Also, it seems a bit strange to need to check if a course is archived when iterating over the active courses.

public/src/pages/StudentPage.tsx
```diff
@@ -184,58 +204,60 @@ export class StudentPage extends ViewPage {
...
+                    const courseID = course.course.getId();
+                    const allLinks: ILink[] = [];
+                    allLinks.push({ name: "Labs" });
+                    const labs = course.labs;
+                    const gLabs: ILink[] = [];
+                    labs.forEach((lab) => {
+                        if (lab.assignment.getIsgrouplab()) {
+                            gLabs.push({
+                                name: lab.assignment.getName(),
+                                uri: this.pagePath + "/courses/" + courseID + "/grouplab/" + lab.assignment.getId(),
```
meling, line 217: These “url” path creation things should really be done in a func... I really dislike the url paths and don’t understand why they are necessary ... but that’s a different story

public/src/pages/TeacherPage.tsx
```diff
@@ -348,32 +350,25 @@ export class TeacherPage extends ViewPage {
...
 
     private async getCourseURL(courseID: number): Promise<string> {
-        const repoMap = await this.courseMan.getRepositories(
-            courseID,
-            [Repository.Type.COURSEINFO],
-            );
-        const fullRepoName = repoMap.get(Repository.Type.COURSEINFO);
-        return fullRepoName ? fullRepoName.split("course-info")[0] : "";
+        const repoMap = this.repositories.get(courseID);
+        return repoMap?.get(Repository.Type.COURSEINFO) ?? "";
```
meling, line 365: Nice improvement

public/src/pages/TeacherPage.tsx
```diff
@@ -348,32 +350,25 @@ export class TeacherPage extends ViewPage {
         }
     }
 
-    private async getCourses(): Promise<Course[]> {
+    private async getCourses(statuses: Enrollment.UserStatus[]): Promise<Course[]> {
         const curUsr = this.userMan.getCurrentUser();
         if (curUsr) {
-            return this.courseMan.getCoursesForUser(curUsr, []);
+            const courses = await this.courseMan.getCoursesForUser(curUsr, statuses);
+            const enrols = await this.courseMan.getEnrollmentsForUser(curUsr.getId());
```
meling, line 357: Looks like there are two round-trips to the server here... can’t we avoid that

public/src/pages/ViewPage.tsx
```diff
@@ -32,9 +32,6 @@ export abstract class ViewPage {
     public async renderContent(page: string): View {
         const pageContent = await this.navHelper.navigateTo(page);
         this.currentPage = page;
-        if (pageContent) {
-            return pageContent;
-        }
-        return <div className="info"><p>Choose a course or create a new one.</p></div>;
+        return pageContent ?? <div className="info"><p>Choose a course or create a new one.</p></div>;
```
meling, line 35: Nice

public/src/pages/TeacherPage.tsx
```diff
@@ -383,4 +378,21 @@ export class TeacherPage extends ViewPage {
...
+    private async fetchCourseRepos(courseID: number): Promise<Map<Repository.Type, string>> {
+        return this.courseMan.getRepositories(courseID,
+                [Repository.Type.COURSEINFO,
+                Repository.Type.ASSIGNMENTS,
+                Repository.Type.TESTS,
+                Repository.Type.SOLUTIONS]);
+    }
+
+    private setupRepos(): Map<number, Map<Repository.Type, string>> {
+        const courses = this.courses;
```
meling, line 390: Nice func, but do you need to copy `this.courses` here?

public/src/pages/views/VisibilityView.tsx
```diff
@@ -0,0 +1,170 @@
...
+    enrollments: Enrollment[];
+    onChangeClick: (enrol: Enrollment) => Promise<boolean>;
+}
+
+interface VisibilityViewState {
+    sortedCourses: Enrollment[];
+    editing: boolean;
+}
+
+export class CourseVisibilityView extends React.Component<VisibilityViewProps, VisibilityViewState> {
```
meling, line 17: What about `CourseListView` and similar renaming elsewhere

public/src/pages/views/CourseView.tsx
0xf8f8ff: Unfortunately, you can't. It would result in type mismatch: courses in props is `Course[]`, and the method returns `Course[] | Enrollment[]`. TypeScript does not approve.

0xf8f8ff: Or actually you can if you assert the type right after the method call. Changed.

meling: ok, np

public/src/pages/StudentPage.tsx
0xf8f8ff: I guess, "active" here refers to the fact that only the courses where enrollment status in not pending are being included here.
Btw, are names like "archived"/"active" really that suitable when we talk about their presence in the user's page menus? Would it be better to rename them to something like "hidden" and, not sure here, "visible", or something else?

0xf8f8ff: Refactored out the repo part of the long method.

public/src/pages/StudentPage.tsx
0xf8f8ff: Yeah, not even sure we need them that much now that we run on grpc, but don't know enough yet to safely remove or replace them either. Making a trello card to address it.

public/src/pages/TeacherPage.tsx
0xf8f8ff: Sure, I can teach the `getCoursesForUser` method to take the visibility state into consideration. But note that this way we move all the sorting from the client to the server. Is it ok?
I suppose it is because it's still better than two consecutive server calls.

public/src/pages/TeacherPage.tsx
0xf8f8ff: Not really.

public/src/pages/views/VisibilityView.tsx
0xf8f8ff: Yeah, sounds reasonable. Fixing it.

public/src/index.tsx
0xf8f8ff: Changed.

public/src/pages/TeacherPage.tsx
0xf8f8ff: Fixed those too.

public/src/components/teacher/Results.tsx
0xf8f8ff: Done.

public/src/pages/StudentPage.tsx
meling: Show and Hide are good names in this context.

0xf8f8ff: What about `VisibilityState` enum of `Enrollment`? They are called `Active`, `Archived`, `Favorite` now. Let them be, or rename to `Visible`, `Hidden`, `Favorite`?

public/src/pages/TeacherPage.tsx
meling: I prefer to do more on the server, where that makes sense of course. But I also don't want to introduce yet another RPC method (adding to the interface) just to support more functionality when it can be accomplished with existing methods, albeit with some minor added complexity in the frontend. Changing the function to support more needs is ok, unless it becomes "artificial"...

public/src/pages/StudentPage.tsx
meling: I like it

public/src/pages/views/CourseView.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from "react";
 import { Course } from "../../../proto/ag_pb";
 import { DynamicTable, Search } from "../../components";
-import { searchForCourses } from "../../componentHelper";
+import { searchForCourses } from '../../componentHelper';
```
meling, line 4: What's the difference between single and doble quotes? Should probably be consistent with one type of quote, but if we want to move to single quote for some specific reason, then this is fine...

0xf8f8ff: Well, I guess it is just manual and automatic imports done in different styles. I'll fix it.

public/src/pages/StudentPage.tsx
0xf8f8ff: Done.

public/src/pages/TeacherPage.tsx
0xf8f8ff: Fixed that.

public/src/pages/views/CourseView.tsx
0xf8f8ff: Done.

public/src/componentHelper.tsx
```diff
@@ -0,0 +1,144 @@
...
+    courses.forEach((e: Enrollment | Course) => {
+        const course = e instanceof Enrollment ? e.getCourse() : e;
+        if (course && foundCourse(course, query)) {
+            e instanceof Enrollment ? enrollmentList.push(e) : coursesList.push(e);
+        }
+    })
+    return enrollmentList.length > 0 ? enrollmentList : coursesList;
+}
+
+export function searchForLabs(labs: IStudentLabsForCourse[], query: string) {
```
meling, line 75: This one is missing return type. Don't understand why this even compiles.

public/src/pages/StudentPage.tsx
```diff
@@ -86,6 +89,23 @@ export class StudentPage extends ViewPage {
         </div >;
     }
 
+    public async archive(navInfo: INavInfo<any>): View {
```
meling, line 92: Maybe rename to `hide`?

public/src/components/teacher/Results.tsx
```diff
@@ -125,7 +125,7 @@ export class Results extends React.Component<IResultsProp, IResultsState> {
     }
 
     private generateUserRepoLink(name: string, userName: string): JSX.Element {
-        return <a href={generateLabRepoLink(this.props.courseURL, userName)} target="_blank">{ name }</a>;
```
meling, line 128: Similar to `groupRepoLink()` this function should also be moved to `componentHandler.tsx` as `userRepoLink()`.
Also do a search for `generateGitLink` to see if more of these can be replaced with `userRepoLink()`

public/src/componentHelper.tsx
meling: Fixed it in my commit

public/src/pages/StudentPage.tsx
0xf8f8ff: Renamed to `courseList` because it is the course list it is responsible for.

public/src/components/teacher/Results.tsx
0xf8f8ff: Done.

PR #262, opened 2020-04-29
Slip days logic

meling:
This mainly implements the slip days logic on the backend, and should also be usable from the frontend through the `Enrollment` message type. It should be possible to implement a frontend function similar to the `TotalSlipDays()` or `RemainingSlipDays()` methods.

The functionality should work (I hope). However, it has not been hooked up with the database yet.


Makefile
```diff
@@ -33,7 +34,7 @@ ui:
 
 proto:
 	@echo Compiling Autograders proto definitions
-	@cd ag; protoc -I=. -I=$(GOPATH)/src -I=$(GOPATH)/src/github.com/gogo/protobuf/protobuf --gogofast_out=plugins=grpc,\
+	@cd ag; protoc -I=. -I=$(pbpath) --gogofast_out=plugins=grpc,\
```
meling, line 37: Does this work for you?

ag/ag.proto
```diff
@@ -1,6 +1,8 @@
 syntax = "proto3";
 
-import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+//TODO(meling) workaround for Go module system
+import "gogoproto/gogo.proto";
```
meling, line 4: And this also causes no problem?

Makefile
0xf8f8ff: Yes, flawlessly.

ag/ag.proto
0xf8f8ff: Well, somehow my VSCode does not really get the idea of imports in `.proto` files and throws the same error for both paths, while the compiler has no problems with either. So, basically, yes, no problems.

ag/enrollment.go
```diff
@@ -0,0 +1,68 @@
+package ag
+
+import (
+	"fmt"
+	"time"
+)
+
+// TODO(meling) must also handle groups; if group assignment is late; withdraw slip days per group member.
+// TODO(meling) also need to propogate slip day information to frontend to show to user
```
meling, line 9: If these are handled now, we can remove them.

PR #268, opened 2020-06-17
Container timeout

0xf8f8ff:
Containers with solution code submitted by students will now have a timeout. When the timeout is reached, and the container is still running, it is usually because of an infinite loop in a student code.  In such a case, the container will be stopped and removed. A message about the reached timeout will be saved in the submission's build log.
By default, timeout for all containers is 10 minutes. However, course teachers can set timeouts for each course assignment individually by adding a `containertimeout` field inside the `assignment.yaml` file.


ag/ag.proto
```diff
@@ -152,7 +152,8 @@ message Assignment {
     uint32 order = 7;
     bool isGroupLab = 8;
     uint32 scoreLimit = 9;  // defines minimal score limit for auto approval
-    repeated Submission submissions = 10; 
+    uint32 containerTimeout = 10;
```
meling, line 155: Will this mess up the DB?

assignments/assignments_parser.go
```diff
@@ -63,14 +64,15 @@ func parseAssignments(dir string, courseID uint64) ([]*pb.Assignment, error) {
...
-					Deadline:    FixDeadline(newAssignment.Deadline),
-					Language:    strings.ToLower(newAssignment.Language),
-					Name:        filepath.Base(filepath.Dir(path)),
-					Order:       uint32(newAssignment.AssignmentID),
-					AutoApprove: newAssignment.AutoApprove,
-					ScoreLimit:  uint32(newAssignment.ScoreLimit),
-					IsGroupLab:  newAssignment.IsGroupLab,
+					CourseID:         courseID,
+					Deadline:         FixDeadline(newAssignment.Deadline),
+					Language:         strings.ToLower(newAssignment.Language),
```
meling, line 69: While I remember; we should consider to rename this field and corresponding field in the yaml file. It should really be named `template`. But question is if we should support both some how. I guess we should make an github issue or trello card.

ci/ci.go
```diff
@@ -11,5 +14,5 @@ type Job struct {
 // Runner contains methods for running user provided code in isolation.
 type Runner interface {
 	// Run should synchronously execute the described job and return the output.
-	Run(context.Context, *Job, string) (string, error)
+	Run(context.Context, *Job, string, time.Duration) (string, error)
```
meling, line 17: Can't we use the `context.Context` to handle the timeout? 
https://golang.org/pkg/context/#example_WithTimeout

ci/docker.go
```diff
@@ -50,9 +50,19 @@ func (d *Docker) Run(ctx context.Context, job *Job, user string) (string, error)
...
+		timeout = containerTimeout
+	}
+
 	select {
 	case wErr := <-errc:
+		fmt.Println("wErr: ", wErr.Error())
 		return "", wErr
+		// if the container still running after predefined time interval, force kill it
+	case <-time.After(timeout):
+		fmt.Println("Timeout done, stopping container: ", user)
```
meling, line 63: String should say: "Timeout, stopping container: "
(but not sure we should print like this; does Docker struct have a logger?)

ci/docker.go
```diff
@@ -50,9 +50,19 @@ func (d *Docker) Run(ctx context.Context, job *Job, user string) (string, error)
...
+		cli.ContainerKill(ctx, resp.ID, "SIGTERM")
+		return fmt.Sprintf("Container timed out after %d minutes", timeout), nil
```
meling, line 65: The `time.Duration` type has a `String()` method, and so if you print with `%v` or similar, it should print with the appropriate unit, since the timeout might not be in minutes...

ag/ag.proto
0xf8f8ff: Never did for me, the gorm's automigrate always took care of that, but I see your concern. Will change to preserve the old numbering.

assignments/assignments_parser.go
0xf8f8ff: We already have a card about this. The plan was to do it when moving from yaml files to UI forms, where we could provide a dropdown menu with all languages that have a ready template.

ci/ci.go
0xf8f8ff: Technically, sure. But do you mean passing the context with an already set timeout to the `Run()` method? 
The problem with this is (just tried it) that all the Docker API methods use the same context. Because of that, when the context timeout is reached, container just fails and `Run()` returns with an error about this:
`ERROR   ci/run_tests.go:62      Test execution failed: %!w(context.deadlineExceededError={})`
I.e., there will be a log message to debug, but no new submission with updated build log will be saved in the database, and therefore no information about what happened will be available to teachers or students.

ci/docker.go
0xf8f8ff: No loggers as far as I can see. Not sure we have to log it here either, the message is being saved in the build logs of the updated submission.

ci/docker.go
0xf8f8ff: Noted.

ag/ag.proto
meling: If auto migrate works, then I'm not very concerned 👍

ci/ci.go
meling: Don't have time to investigate this now, but I was thinking that we still control the timeout of the context. I'm not sure all of them need to use the same context, but even so, why would they timeout with a X minute long timeout. (Will try to understand more later)

meling:
@0xf8f8ff Can you please resolve the conflicts. We will just merge the current version into master, and if we find a way to use the context timeout, we can fix that later.

PR #281, opened 2020-07-20
Revised golang processing w/custom Docker image

### LINKED ISSUE #273: Autograder runs tests from student repository (which can also affect score)
In addition to running tests from the `tests` repo connected to some lab assignment, AG runs the tests in `<file>_test.go` from the student repository. This can cause some problems:

- Some tests might crash if the original test and the AG equivalent runs at the same time (this was observed e.g. with `TestWrite` in lab 3 for dat320-2020). If unfortunate, this could reduce the score of a correct solution.
- The AG output is cluttered with local tests which should not be there. Especially if students have made some tests of their own which happen to fail.
- Slight waste of resources.

Additionally, if the test from the student repo uses AG score objects, they can give themselves a higher score. While this would count as cheating, it does not make sense that students can write their own tests which affect their score.

Consider the following test (which was tested in lab 3 for dat320-2020):

```go
package lab3

import (
	"os"
	"testing"

	"github.com/autograde/aguis/kit/score"
)

func TestDummy(t *testing.T) {
	sc := score.NewScoreMax(100, 1)
	defer sc.WriteString(os.Stdout)
	defer sc.WriteJSON(os.Stdout)
	t.Error("dummy test failed")
	sc.DecBy(100)
}
```

First of all this test fails and shows up in the AG log. Additionally, the total score of the lab as well as the final score are affected by the `sc.DecBy(100)` operation. It also shows up in the "Test name" column in AG.

### LINKED ISSUE #275: Make custom docker image and revise go.tmpl file
We may speed up test execution by making a custom docker image and revise `go.tmpl` file to be more efficient.
Benefits to be had:

- We can incorporate a precompiled version of the `gosecret` command
- We can cache common GitHub repos, such as the `tests` repo and other dependencies. See this project for [example](https://github.com/montanaflynn/golang-docker-cache).

meling:
This PR fixes some issues with the old go.tmpl file and
improves the performance of running tests. It fixes #273
and fixes #275.

- Rename aguis to quickfeed; all search replace
- added test for unknown fields in yaml file.
- fixed TestDocker to run on macOS
- refactored RunTests to simplify testing with real repos
- added Dockerfile for generating custom Quickfeed image
- updated quickfeed/kit and go-cmp dependencies
- renamed scripts s/tmpl/sh/ and revised go.sh significantly
- fixes a performance problem due to always pulling image.


ci/docker.go
```diff
@@ -31,38 +31,46 @@ func (d *Docker) Run(ctx context.Context, job *Job, user string, timeout time.Du
...
+	case err := <-errCh:
+		// container failed with error; return
+		return "", err
 	case <-time.After(timeout):
-		cli.ContainerKill(ctx, resp.ID, "SIGTERM")
-		return fmt.Sprintf("Container timed out after %v", timeout), nil
-	case <-waitc:
+		// force kill container after predefined time interval
+		err = cli.ContainerKill(ctx, resp.ID, "SIGTERM")
+		return "", fmt.Errorf("container timed out after %v: %w", timeout, err)
```
0xf8f8ff, line 71: Is it really wise to return the error here? This way there is no container output and the user receives no feedback about what has happened. 
Now, instead of receiving a hint about the timeout in the build info on the `Results` page, students will be left guessing what went wrong, because from their perspective it would look like the tests had never run and something is broken on our side.
Could it be better to change the output message to something more helpful, like hinting on possible presence of an infinite loop in the submitted code, instead of just failing?

meling: Valid point. Will fix. But then we can't return an error as well. It would be nice to also log this fact on the server side. Will look into it.

meling:
@0xf8f8ff I'm running this in production now. Let me know if you see anything that worries you.

PR #296, opened 2020-08-07
Return stdout, stderr, error from func CommandLine

EivindSt:
Motivation: The output from the CommandLine function is minimal in
some cases. E.g. in our use-case we wanted to detect a data race by
running `go test -race ...` with the CommandLine function and looking
for "DATA RACE" in the output. However we were unable to show the full
output (stack trace etc.) related to the data race when it was found.

With this update the output to stdout and stderr along with any error
from exec.Command are returned as slices, one for each command that
is run. As a result, callers of CommandLine can investigate the
results and use them as they please when some error occurs. Since
there were no return values before, this change should not conflict
with existing code.

I was not sure about introducing new types, but you might also
consider a "CommandLineError" type similar to the following:

```go
type CommandLineError struct {
  // output to stdout
  Stdout string
  // output to stderr
  Stderr string
  // error from executing the command with the exec package
  Err error
}

func (e CommandLineError) Error() string {
  return e.Err.Error()
}
```

Then CommandLine could return `[]CommandLineError` instead.

### DIFF
--- a/kit/exercise/command_line.go
+++ b/kit/exercise/command_line.go
@@ -28,9 +28,18 @@ type Commands []struct {
 // that students provided. The function requires the list of commands
 // and their expected answers, and a Score object. The function
 // will produce both string output and JSON output.
-func CommandLine(t *testing.T, sc *score.Score, answers Commands) {
+//
+// In addition to displaying error output through the t argument, the
+// output written to stdout and stderr, along with the error message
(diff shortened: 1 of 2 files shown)


meling:
I think introducing the new type should be fine. Let's just do that right away. I don't think there are too many users of this API anyway, and I don't think adding a new type is breaking.

PR #298, opened 2020-08-07
Discord bot helper

0xf8f8ff:
Started implementing API changes to support student login management by the course's discord bot.


ag/ag.proto
```diff
@@ -445,4 +456,5 @@ service AutograderService {
     rpc GetOrganization(OrgRequest) returns (Organization) {}
     rpc GetRepositories(URLRequest) returns (Repositories) {}
     rpc IsEmptyRepo(RepositoryRequest) returns (Void) {}
+    rpc GetStudentForDiscord(DiscordRequest) returns (DiscordResponse) {}
```
meling, line 459: Suggest this:

```proto
    rpc GetUserByCourse(CourseUserRequest) returns (User) {}
```

Where `CourseUserRequest = DiscordRequest`, and `User` is the same as our current `User`. The remote ids are removed before the result is returned anyway. The Discord bot could maybe use the same avatar URL and admin field.

ag/ag.proto
```diff
@@ -371,6 +371,17 @@ message RebuildRequest {
     uint64 assignmentID = 2;
 }
 
+message DiscordRequest {
```
meling, line 374: See below: `CourseUserRequest`

ag/ag.proto
```diff
@@ -371,6 +371,17 @@ message RebuildRequest {
...
+    string courseCode = 1;
+    uint32 courseYear = 2;
+    string userLogin = 3;
+}
+
+message DiscordResponse {
```
meling, line 380: Reuse `User` instead.

ag/validation.go
```diff
@@ -231,3 +231,8 @@ func (bm GradingBenchmark) IsValid() bool {
 func (c GradingCriterion) IsValid() bool {
 	return c.BenchmarkID > 0 && c.Description != ""
 }
+
+// IsValid ensures that course code, year, and student login are set
+func (r DiscordRequest) IsValid() bool {
+	return r.CourseCode != "" && r.UserLogin != "" && r.CourseYear > 0
```
meling, line 237: `r.CourseYear > 2019`, or at least 2018 or something, since there are no courses from earlier years...

database/gormdb_course.go
```diff
@@ -84,6 +84,15 @@ func (db *GormDB) GetCourseByOrganizationID(did uint64) (*pb.Course, error) {
 	return &course, nil
 }
 
+// GetCourseBycodeAndYear returns course with given code and year
```
meling, line 87: `GetCourseByCodeAndYear` with Capital `C` in `Code`

database/gormdb_course.go
```diff
@@ -84,6 +84,15 @@ func (db *GormDB) GetCourseByOrganizationID(did uint64) (*pb.Course, error) {
...
+func (db *GormDB) GetCourseByCodeAndYear(courseCode string, courseYear uint32) (*pb.Course, error) {
```
meling, line 88: To make this more flexible, we should take a `*pb.Course`, letting the caller fill the fields that should be queried.
Maybe it should be renamed to `GetCourseByCourseQuery` or something better. It looks a bit difficult to convert `GetCourse` into a method that works seamlessly, although that would be better IMO...

meling: This all said. Maybe this should be implemented as a Join operation on the db, where we query both course and user in the same transaction. This way we would avoid two database methods... (but maybe not worth it, since it increases complexity a bit...)

public/src/managers/GRPCManager.ts
```diff
@@ -340,6 +342,14 @@ export class GrpcManager {
         return this.grpcSend<Void>(this.agService.isEmptyRepo, request);
     }
 
+    public getStudentForDiscord(courseCode: string, courseYear: number, userLogin: string) {
```
meling, line 345: Do you need to implement this method in the frontend??

public/src/managers/ServerProvider.ts
```diff
@@ -413,6 +414,14 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
         return this.responseCodeSuccess(result);
     }
 
+    public async getStudentForDiscord(courseCode: string, courseYear: number, userLogin: string): Promise<DiscordResponse | Status> {
```
meling, line 417: Do you need to implement this method in the frontend??

public/src/managers/UserManager.ts
```diff
@@ -13,6 +13,7 @@ export interface IUserProvider {
     getLoggedInUser(): Promise<User | null>;
     updateUser(user: User): Promise<boolean>;
     isAuthorizedTeacher(): Promise<boolean>;
+    getStudentForDiscord(courseCode: string, courseYear: number, userLogin: string): Promise<DiscordResponse | Status>;
```
meling, line 16: Do you need to implement this method in the frontend??

web/autograder_service.go
```diff
@@ -855,6 +855,22 @@ func (s *AutograderService) IsEmptyRepo(ctx context.Context, in *pb.RepositoryRe
...
 		return nil, status.Errorf(codes.FailedPrecondition, "group repository does not exist or not empty")
 	}
-
 	return &pb.Void{}, nil
 }
+
+// GetStudentForDiscord fetches student name and ID by the course name and student's GitHub login
+// Access policy: Course teachers
+func (s *AutograderService) GetStudentForDiscord(ctx context.Context, in *pb.DiscordRequest) (*pb.DiscordResponse, error) {
+	usr, err := s.getCurrentUser(ctx)
```
meling, line 864: Will the Discord bot need to authenticate with GitHub to use the method??

web/users.go
```diff
@@ -49,6 +49,16 @@ func (s *AutograderService) getUsers() (*pb.Users, error) {
 	return &pb.Users{Users: users}, nil
 }
 
+//
+func (s *AutograderService) getStudentForDiscord(request *pb.DiscordRequest, currentUser *pb.User) (*pb.DiscordResponse, error) {
```
meling, line 53: Missing code?

web/autograder_service.go
0xf8f8ff: Well, is it not reasonable to expect authentication with a user who is also a course teacher? Because otherwise we would just expose a method returning a personal information on any user that could be called by literally anyone. 
And GitHub authentication is the only option we support for now.

public/src/managers/GRPCManager.ts
0xf8f8ff: Probably not.

database/gormdb_course.go
0xf8f8ff: Alternatively, we could remove this method and keep the one fetching the user. We can pass the course query to the `GetUserByLogin` method (while renaming it to something like `GetUserByCourseAndLogin`). This way we can return an error both when there is no such course, no such user, or user is not enrolled, directly from here, without extra database calls and checks. If every condition is met, we would return the user already with the preloaded course. This way we also end up with just one method.

web/autograder_service.go
meling: Yes, we should definitely have authentication; we just need to provide an example, e.g. in the form of a test case or something, so that it becomes easy to integrate into the discord bot.

database/gormdb_course.go
meling: Sounds good to me!  Maybe we could just call it `GetUserByCourse(login string, course *pb.Course)` implying that it will also require the `login`, but that it's the course that counts... but it is debatable what to call it...

web/users.go
```diff
@@ -51,12 +51,7 @@ func (s *AutograderService) getUsers() (*pb.Users, error) {
 
 // getUserByCourse returns the user matching the given GitHub login if
 // the user is enrolled in the given course.
-<<<<<<< HEAD
 func (s *AutograderService) getStudentByCourse(request *pb.CourseUserRequest, currentUser *pb.User) (*pb.User, error) {
```
meling, line 54: Isn't this better named `getUserByCourse`?

This check:

```go
	if !(currentUser.IsAdmin || s.isTeacher(currentUser.ID, course.ID)) {
	}
```

Is only checking that the teacher or an admin is performing the operation... So technically, this can also return another teacher as well... or?

meling:
Let me know when this can be merged...

web/users.go
0xf8f8ff: Well, sure. As far as I understand, the teaching staff will be present on the channel, as well as all the students. Am I wrong?

meling: Yes, there needs to be a bootstrap I guess... but maybe also TAs can use `!register github_login` to sign up ... which would call this method...

web/discord_test.go
```diff
@@ -19,17 +20,19 @@ import (
...
+		t.Fatal("Connection failed, make sure the server is running on :9090")
 	}
 	defer conn.Close()
 
 	client := pb.NewAutograderServiceClient(conn)
 
 	currentUserID := os.Getenv("DISCORD_USER")
 	if currentUserID == "" {
-		t.Fatal("User ID is not set")
+		t.Skip("This test requires a 'DISCORD_USER' environmental variable with a valid ID of a registered user")
```
meling, line 35: This check should be before the `DialContext` above, since otherwise the test will block if not able to connect to a server or something...

0xf8f8ff: Right. Fixed it.

PR #299, opened 2020-08-11
Rename language field in assignment.yml and proto file

meling:


meling:
@0xf8f8ff I'm done with this PR, but left it as draft since the DB must also be updated, which needs your expertise. Let me know if you spot any issues. And let me know when we can merge this, preferably soon, so that I can convert the current labs to use `scriptfile`...

database/gormdb_assignment.go
```diff
@@ -34,7 +34,7 @@ func (db *GormDB) CreateAssignment(assignment *pb.Assignment) error {
 		Assign(map[string]interface{}{
 			"name":              assignment.Name,
 			"order":             assignment.Order,
-			"language":          assignment.Language,
+			"language":          assignment.ScriptFile,
```
0xf8f8ff, line 37: Shouldn't it be `"script_file": assignment.ScriptFile`? Now it attempts to save to the `language` field of the `Assignment` table, which should not even exist.

meling: Yes, thanks! That's a mistake. Will fix

PR #335, opened 2020-09-23
fix for the too many open files problem

### LINKED ISSUE #319: Too many open files
Currently, quickfeed is apparently not closing some connection properly, which can cause this error and essentially take down the whole server:

```text
echo: http: Accept error: accept tcp [::]:3005: accept4: too many open files; retrying in 1s
echo: http: Accept error: accept tcp [::]:3005: accept4: too many open files; retrying in 1s
echo: http: Accept error: accept tcp [::]:3005: accept4: too many open files; retrying in 1s
echo: http: Accept error: accept tcp [::]:3005: accept4: too many open files; retrying in 1s
```

There is [this issue here](https://stackoverflow.com/questions/37454236/net-http-server-too-many-open-files-error), that is related but we are not doing regular REST API, so it is unclear if they are related. We still do some bits with REST still, so it could be that we are not closing login/auth connections properly.

While we wait for a fix, and if this problems becomes even more annoying, we can increase the open files limit here `/etc/security/limits.conf`; if so, we should use `prlimit` to change it e.g.: 

```console
prlimit --pid 7368 --nofile=1024:4095
```

One idea would be to get rid of the whole REST API/echo framework... That would hopefully avoid this problem.
But that is a bit of a risky move mid-semester, but if we test it well, then it would be a good time to get good feedback and testing...

meling:
This commit fixes #319.
The problem was that we only stopped the container, but didn't
actually remove it, even though we waited for it to finish.
This is mainly an issue resulting from not understanding that
docker containers remains "available" even though the job
running inside the docker container had exited.

We now remove containers that finish without timeout, but only
after having extracted the logs.
Otherwise, we would not be able to get the logs.
If a container times out, we stop the container before removing it.

### DIFF
--- a/ci/docker.go
+++ b/ci/docker.go
@@ -63,23 +63,37 @@ func (d *Docker) Run(ctx context.Context, job *Job) (string, error) {
 	// wait until the container stops or context times out.
 	_, err = cli.ContainerWait(ctx, resp.ID)
 	if err != nil {
-		if errors.Is(err, context.DeadlineExceeded) {
-			timeout := time.Duration(1 * time.Second)
-			// ContainerStop seems to work better than ContainerKill in my tests.
-			stopErr := cli.ContainerStop(context.Background(), resp.ID, &timeout)
-			if stopErr != nil {
-				// if we failed to stop, report it up the chain
-				err = stopErr
-			}
-			userMsg := "Container timeout. Please check for infinite loops or other slowness."
-			return userMsg, err
+		if !errors.Is(err, context.DeadlineExceeded) {
+			return "", err
 		}
-		return "", err
+
+		// stop runaway container whose deadline was exceeded
+		timeout := time.Duration(1 * time.Second)
+		stopErr := cli.ContainerStop(context.Background(), resp.ID, &timeout)
+		if stopErr != nil {
+			return "", stopErr
+		}
+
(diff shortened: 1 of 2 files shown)


ci/docker.go
```diff
@@ -63,27 +63,41 @@ func (d *Docker) Run(ctx context.Context, job *Job) (string, error) {
...
+	// extract the logs before removing the container below
 	logReader, err := cli.ContainerLogs(ctx, resp.ID, types.ContainerLogsOptions{
 		ShowStdout: true,
 	})
 	if err != nil {
 		return "", err
 	}
 
+	// remove the container when finished to prevent too many open files
+	err = cli.ContainerRemove(ctx, resp.ID, types.ContainerRemoveOptions{})
```
0xf8f8ff, line 96: When we tried the `ContainerRemove()` last year we would get no container output. But the docker version was older, so it might work now, just needs to be tested.

meling: Yeah, see the log message. That's why I remove it after `cli.ContainerLogs()` which is used to get the logs from the container before removing it.

meling: Also, the tests passes locally. Meaning that the output is as expected, even though the image was removed. I confirmed that the image was removed with `docker ps -a`.

0xf8f8ff: Oh, right. Then it should be ok.

PR #346, opened 2020-10-01
Fix to the recent slipdays problem

0xf8f8ff:
Will no longer run tests if a submission to an individual assignment was pushed to a group repo, or a submission for a group assignment pushed to a user repo.


web/hooks/github.go
```diff
@@ -81,15 +81,19 @@ func (wh GitHubWebHook) handlePush(payload *github.PushEvent) {
...
 		assignments := wh.extractAssignments(payload, course)
 		for _, assignment := range assignments {
-			runData := &ci.RunData{
-				Course:     course,
-				Assignment: assignment,
-				Repo:       repo,
-				CloneURL:   payload.GetRepo().GetCloneURL(),
-				CommitID:   payload.GetHeadCommit().GetID(),
-				JobOwner:   payload.GetSender().GetLogin(),
+			if repo.IsOfMatchingType(assignment.IsGroupLab) {
```
meling, line 84: I propose that we more clearly separate this logic. Let's instead make a separate:

```golang
case repo.IsGroupRepo():
    ...
```

With similar code as here. The reason for this is that then it will be easier to make future customizations for group repos etc. Perhaps we can just make a function for the common parts...

web/hooks/github.go
```diff
@@ -81,15 +81,19 @@ func (wh GitHubWebHook) handlePush(payload *github.PushEvent) {
...
+					Course:     course,
+					Assignment: assignment,
+					Repo:       repo,
+					CloneURL:   payload.GetRepo().GetCloneURL(),
+					CommitID:   payload.GetHeadCommit().GetID(),
+					JobOwner:   payload.GetSender().GetLogin(),
+				}
+				ci.RunTests(wh.logger, wh.db, wh.runner, runData)
+			} else {
+				wh.logger.Debugf("Push was for repository type %s, while %s is %s assignment ", repo.RepoType, assignment.Name, assignment.TypeString())
```
meling, line 95: If we make a separate case for group repo, as I propose above, we can write the same info in the processing push event message on line 81 instead and we don't need logging here.

web/hooks/github.go
0xf8f8ff: Fixed, if I understood you correctly. Could you take a look?

PR #348, opened 2020-10-02
assignment info url

meling:
- some assignment info and url clean up
- minor log fix


0xf8f8ff (approved):
All the test seem to run correctly.

PR #353, opened 2020-10-19
Active/inactive students in the list of course members

### LINKED ISSUE #349: Members view should show inactive user status
All users are shown in the Results view, even though they are inactive, as in never having pushed anything to their repository. We want to remove such users from the course, delete their repositories and remove them from the organization. Currently, we have implemented this feature via the Members view, but we must first find such users manually in the Results view and go to the Members view to remove them.

I propose instead that the Members view can display an additional column titled `Activity` that should display 

- `Inactive` if the user hasn't committed any changes to their repository or
- `AssignmentX`, that is the name of the last approved assignment.
- Or maybe the date and assignment of the last push.

In addition, we should have some way to remove all Inactive users with one button, obviously with the necessary warnings explaining that this action will `remove these users from the course, delete their user and group repositories and remove them from the organization.`

0xf8f8ff:
Closes #349.


web/hooks/github.go
```diff
@@ -88,6 +90,11 @@ func (wh GitHubWebHook) handlePush(payload *github.PushEvent) {
 		}
 	case repo.IsGroupRepo():
 		wh.logger.Debugf("Processing push event for group repo %s", payload.GetRepo().GetName())
+		jobOwner, _, err := wh.db.GetUserByCourse(course, payload.GetSender().GetLogin())
+		if err != nil {
+			wh.logger.Debugf("Failed to find user %s in the course %s", payload.GetSender().GetLogin(), course.GetName())
```
meling, line 95: I think this should probably be logged using `Errorf` and it should `return` after, so that it doesn't pass through to the code below.

web/hooks/github.go
```diff
@@ -152,3 +159,19 @@ func (wh GitHubWebHook) runAssignmentTests(assignment *pb.Assignment, repo *pb.R
...
+	today := time.Now()
+
+	query := &pb.Enrollment{
+		UserID:           userID,
+		CourseID:         courseID,
+		LastActivityDate: today.Format("02 Jan"),
+	}
+
+	if err := wh.db.UpdateEnrollment(query); err != nil {
+		wh.logger.Debugf("Failed to update the last activity date for user %d: %s", userID, err)
```
meling, line 175: This should log with `Errorf`

web/hooks/github.go
```diff
@@ -152,3 +159,19 @@ func (wh GitHubWebHook) runAssignmentTests(assignment *pb.Assignment, repo *pb.R
...
+
+// updateLastActivityDate sets a current date as a last activity date of the student
+// on each new push to the student repository.
+func (wh GitHubWebHook) updateLastActivityDate(userID, courseID uint64) {
+	today := time.Now()
+
+	query := &pb.Enrollment{
+		UserID:           userID,
+		CourseID:         courseID,
+		LastActivityDate: today.Format("02 Jan"),
```
meling, line 171: Can replace `today.Format(...)` with `time.Now().Format("02 Jan")`

web/hooks/github.go
0xf8f8ff: Fixed.

web/hooks/github.go
0xf8f8ff: Fixed as well.

web/hooks/github.go
0xf8f8ff: Done.

PR #356, opened 2020-10-22
Added DeleteRepositorySubscription and ListWatched funcs

meling:


meling:
@0xf8f8ff can you take a look and see if you can integrate this, so that we can avoid subscriptions to everything... I have tested that this func works to remove my current subscriptions, but it would be nice to integrate it with the user/group creation process. Alternatively, if we can create user/group repos etc without subscription initially, perhaps we don't even need these funcs in the `scm` package.

0xf8f8ff:
But wouldn't integrating these methods with user, group, or repo creation process only remove subscriptions for the currently authorized user (will it just be the Autograder user in our case)? The point is to disable subscriptions/notifications coming from all the student and group repos for all organization owners, am I correct?

meling:
I think so.

meling:
Maybe this can be done more simply at the organization configuration level?

0xf8f8ff:
> Maybe this can be done more simply at the organization configuration level?

This is what I was trying to find a way doing, but it looks like it is only allowed to change your own subscription/notification settings. Still looking for a possible solution to automate it for every organization owner at once, but alternatively we could just add some kind of a button ("Unsubscribe" or something) to the course UI for teachers that would remove all repo subscriptions or notifications for the current user.

meling:
@0xf8f8ff did you find a proper solution to this problem? I would prefer to close this PR.

meling:
I think the solution to this problem is simply to follow [these instructions](https://stackoverflow.com/questions/25108169/how-do-i-turn-off-automatic-notification-subscription-for-new-repositories-in-a). I'll instead update some markdown file with the relevant information, and when I know that this stops the notifications, I'll close this PR.

meling:
Closing since this seems to have been fixed after following the instructions that I added to a new notifications section in doc/teacher.md.

PR #359, opened 2020-10-24
Inactive students

0xf8f8ff:
Updates to #349.


web/courses.go
```diff
@@ -469,11 +479,53 @@ func (s *AutograderService) enrollTeacher(ctx context.Context, sc scm.SCM, enrol
...
+		var totalApproved uint64
+		for _, submissionLink := range enrolLink.Submissions {
+			if submissionLink.Submission != nil && submissionLink.Submission.Status == pb.Submission_APPROVED {
+				totalApproved++
+			}
+		}
+		enrol.TotalApproved = totalApproved
+
+		if enrol.LastActivityDate == "" {
+			for i := range enrolLink.Submissions {
```
meling, line 500:
```go
for i := range enrolLink.Submissions {
  submissionlink := enrolLink.Submissions[i]
```
can be replace with:
```go
for _, submissionlink := range enrolLink.Submissions {
```

web/courses.go
```diff
@@ -469,11 +479,53 @@ func (s *AutograderService) enrollTeacher(ctx context.Context, sc scm.SCM, enrol
...
+		enrol := enrolLink.Enrollment
+		var totalApproved uint64
+		for _, submissionLink := range enrolLink.Submissions {
+			if submissionLink.Submission != nil && submissionLink.Submission.Status == pb.Submission_APPROVED {
+				totalApproved++
+			}
+		}
+		enrol.TotalApproved = totalApproved
+
+		if enrol.LastActivityDate == "" {
```
meling, line 499: Move this test inside the loop above, and you can avoid the loop below, avoding to iterate over `enrolLink.Submissions` twice. Though, I suggest to separate the content of the loop below into a separate function, e.g. `extractLastActivity` or something.

0xf8f8ff: Fixed.

PR #375, opened 2020-12-09
Parse grading criteria from a JSON file

0xf8f8ff:
Option to parse grading criteria from a JSON file inside the `Tests` folder will enable easy transfer of the criteria lists from the last year assignments. 
image

Calculate custom total score for the assignment if grading crhiteria in JSON file have individual scores set.


0xf8f8ff:
Temporarily converted back to draft to fix an issue.

assignments/assignments.go
```diff
@@ -26,16 +26,19 @@ func UpdateFromTestsRepo(logger *zap.SugaredLogger, db database.Database, repo *
 		logger.Errorf("Failed to fetch assignments from '%s' repository: %w", pb.TestsRepo, err)
 		return
 	}
+
```
meling, line 29: Don't think we need to add line breaks here; the single `}` is already quite an empty line, so I don't feel that adding another blank line here helps code comprehension.

database/gormdb_submission.go
```diff
@@ -80,8 +80,17 @@ func (db *GormDB) GetSubmission(query *pb.Submission) (*pb.Submission, error) {
 	return &submission, nil
 }
 
+// GetSubmissions returns all submissions matching the query.
+func (db *GormDB) GetSubmissions(query *pb.Submission) ([]*pb.Submission, error) {
```
meling, line 84: Do we need both these methods?
For this method, do you need all submissions, or just the latest?
From its point of use in `web/assignments.go:156`, it looks like you have access to the `courseID` in the `pb.Assignment` object, so you could use the same signature as `GetSubmissionsForCourse`.

doc/teacher.md
```diff
@@ -155,3 +158,44 @@ Comments can be left to every criterion checkpoint or to the whole group of grad
...
+            {
+                "description": "Looks nice",
+                "score": 10
+            }
+        ]
+    }
+]
+```
+
+`score` field is optional. If set, the max score for the assignment will be equal to the sum of all scores for each criteria. Otherwise, the max total score will be 100%.
```
meling, line 201: Missing newline.

To avoid confusion, I wonder if this should be `max_score` instead, since this is a declaration of scoring criteria.
If not set, I think this should mean that `each criterion counts equally towards the total score`, or something like that.

public/src/components/manual-grading/Release.tsx
```diff
@@ -330,11 +332,12 @@ export class Release extends React.Component<ReleaseProps, ReleaseState>{
...
                 reviews: ready,
-                score: totalScore(ready),
+                score: scoreFromReviews(ready),
                 status: this.props.submission?.status ?? Submission.Status.NONE,
             });
         } else {
             this.setState({open: !this.state.open});
         }
     }
+
```
meling, line 342: Maybe remove this blank line and add a newline at the end of the file.

public/src/components/manual-grading/Review.tsx
```diff
@@ -377,4 +396,5 @@ export class ReviewPage extends React.Component<ReviewPageProps, ReviewPageState
         }
         return rw;
     }
+
```
meling, line 399: Don't need this newline.

scm/github.go
```diff
@@ -692,3 +692,31 @@ func toRepository(repo *github.Repository) *Repository {
...
+func (s *GithubSCM) GetFileContent(ctx context.Context, opt *FileOptions) (string, error) {
+	if !opt.valid() {
+		return "", ErrMissingFields{
+			Method:  "GetFileContent",
+			Message: fmt.Sprintf("%+v", opt),
+		}
+	}
+
+	fileContent, _, _, err := s.client.Repositories.GetContents(ctx, opt.Owner, opt.Repository, opt.Path, nil)
+	if err != nil || fileContent == nil {
```
meling, line 706: Can `fileContent` be `nil` if `err != nil`??
This would be bad practice to have such an API that allows to returning `nil` for both `err` and `fileContent`

meling: Never mind. Checked the API, and it appears to be like this because the fields ignored can be directory content.

web/assignments.go
```diff
@@ -74,6 +79,55 @@ func (s *AutograderService) deleteCriterion(query *pb.GradingCriterion) error {
 	return s.db.DeleteCriterion(query)
 }
 
+func (s *AutograderService) loadCriteria(ctx context.Context, sc scm.SCM, request *pb.LoadCriteriaRequest) ([]*pb.GradingBenchmark, error) {
+
+	assignment, err := s.db.GetAssignment(&pb.Assignment{ID: request.AssignmentID, CourseID: request.CourseID})
```
meling, line 84: The method `loadCriteria` is 47 lines long. Consider if lines 84-92 can be made into its own method, but only if there are other similar use cases for getting assignment and course objects. Alternatively, lines 84-103 could also be its own method.

web/assignments.go
```diff
@@ -74,6 +79,55 @@ func (s *AutograderService) deleteCriterion(query *pb.GradingCriterion) error {
...
+	}
+
+	opts := &scm.FileOptions{
+		Path:       filepath.Join(assignment.GetName(), criteriaFile),
+		Owner:      course.OrganizationPath,
+		Repository: pb.TestsRepo,
+	}
+
+	criteriaString, err := sc.GetFileContent(ctx, opts)
+	if err != nil || criteriaString == "" {
```
meling, line 101: What if `err == nil` and `criteriaString == ""`. This method will return `nil, nil`. Is that expected?

meling: Some would argue that here one should return an empty `[]*pb.GradingBenchmark`. Or maybe it should be an error if the file content is empty?

0xf8f8ff: Right. I think there should be an error in this case, it will now be returned by the `GetFileContent` method if the file is empty and logged properly.

public/src/components/manual-grading/Release.tsx
0xf8f8ff: Done.

public/src/components/manual-grading/Review.tsx
0xf8f8ff: Fixed.

doc/teacher.md
0xf8f8ff: But it is the score for a single criterion, not the max score for the whole set of criteria? Isn't the name `max_score` equally confusing_

web/assignments.go
0xf8f8ff: You are right, there are a few other use cases where we want an assignment with the corresponding course. Refactored.

database/gormdb_submission.go
0xf8f8ff: `GetSubmissionsForCourse` is twice as slow, but I guess it is not that important here. Removed the new method.

assignments/assignments.go
0xf8f8ff: Done.

doc/teacher.md
0xf8f8ff: Updated the docs.

database/gormdb_submission.go
0xf8f8ff: `GetSubmissionsForCourse` will return one submission per each course assignment, and here we need all submissions for a single assignment, so these methods are not really interchangeable. Now that you changed the `GetSubmissionsByCourse` to be `GetSubmissions` there is a conflict here. What do you think I should name the new method than? It can be used to fetch multiple submissions by any parameter in the query, so, while i name it `GetSubmissionsByAssignment` for now, it is not really descriptive.

meling: My issue was not the name of the method, it was whether or not they were both needed. Looking at the code, the following naming feels more natural to me: 

```text
Current                      --> New
GetSubmissions()             --> GetLastSubmissions()  OR  GetLastSubmissionForEachAssignment()
GetSubmissionsByAssignment() --> GetSubmissions()
```

meling: To clarify, I'm okay with the two methods, since they provide different functionality.

0xf8f8ff: Renamed.

doc/teacher.md
meling: First of all, I don't see any changes in the docs as mentioned above. Did something happen to the commit?

Second, the problem I have with this is that these criterion definitions use the same terminology as in the [`kit/score.Score`](https://github.com/autograde/quickfeed/blob/725ddfa95cf3ac20c53d7d698473dc31a72dd158/kit/score/score.go#L33) type, where `MaxScore` is defined for one score object, and the `Score` can take any value in the range `[0,MaxScore]`.

The question I have is what these `score` lines in the JSON format really mean?

Do they each define the maximum score that can be awarded for one of these criterion?

Or looking at this now, it appears as if these are binary answers, each giving `score` points. Meaning that each criterion either gives `0` or `score` points, and nothing in-between. If that's the case, then maybe use the term `points` to distinguish this from the other use of `score`. 

Mapping this to the `Score` type:

```go
sumPoints := <..>
sc := score.NewMaxScore(testName, sumPoints, weight)
// for each incorrect criterion
sc.DecBy(criterionPoints)
```

(currently the signature shown below won't work, but in a future update it will)

PS: Don't know if these JSON things are even mapped to score objects, but either way it makes sense to avoid confusing the terminology.

PPS: Note that `criterion` is singular, and `criteria` is plural.

web/assignments.go
meling: Looks good to me!

doc/teacher.md
0xf8f8ff: No, they are not mapped to score objects. Agree that `points` is a better option. Renamed.

PR #384, opened 2021-01-12
Revised scoring mechanism based on registration of tests pre-test execution

### LINKED ISSUE #27: Grading scheme support should be revised slightly
Autograder should support different grading schemes. At least three grading schemes will be supported by default, but teachers (or admins) can create custom grading schemes.

- Default grading schemes: 
  1. Pass/Fail (60 % passing level)
  2. C bias ECTS scheme (University of Stavanger)
  3. No bias ECTS scheme (NTNU)
- Each assignment can use a different grading scheme. 
- A grading scheme is specified as a mapping from a set of non-overlapping intervals in the range 0-100 to a specific grade for an interval.

```Go
type GradingScheme struct {
  Name        string   // name of grading scheme
  GradePoints []uint8  // grade boundaries; from highest grade's level to 0 for the failing grade
  GradeNames  []string // names for the different grades; corresponds to the GradePoints
}
```

### LINKED ISSUE #283: Document and add Example of how to use the kit package
Add a few Example tests to illustrate scoring in the kit module... for several languages.

### LINKED ISSUE #374: Fix score.Total to tolerate nil entries or revise the signature
The `score.Total` function should tolerate `nil` entries and give zero points for such entries. That is, it should set the `MaxScore = 1` (or some other non-zero value), and set `Score = 0`. The issue is what should be the default for the `Weight` parameter, for such `nil` entries. 

A second, maybe better option, would be to require a second parameter with a map of test weights, which makes it easier to keep track of the expected number of score objects. The length of the two input slices must be the same.

```go
func Total(scores []*score.Score, weights []uint8) uint8
```

Or

```go
func Total(scores []*score.Score, weights map[string]uint8) uint8
```

where `weights` map from `TestName` to its weight.

Note that `Total` should return `uint8` (instead of `uint32`) to make it compatible with the `score.GradeScheme` type.

### LINKED ISSUE #433: Update frontend to use revised Submission message
The `Submission` message in `ag.proto` previously used `BuildInfo` and `ScoreObjects` as JSON strings. These are being replaced with the new `score.Results` message that is the `Results` field in `Submission`. The server-side code has been updated (see PR #413), but the frontend code has yet to be updated. This update needs assistance from several students.

meling:
This PR implements a revised scoring mechanism that requires tests to be registered before test execution. This will allow QuickFeed to become aware of all tests to be expected, avoiding the problem that some tests may be lost due to a panic or other failures that prevent tests from emitting their results. Tests that panic are not acceptable, and so students must fix such panics to get scores for tests that may follow a panicking test.

Resolves #283, resolves #27, resolves #374, resolves #433.


meling:
When reviewing this PR, please consider the examples in `kit/score/testdata/sequence` and other test cases within the `score` package itself.

The new API is mainly documented in `doc.go`, `score.go` and `registry.go`. The new API is slightly less convenient, in that:
- a test must now be registered in either `TestMain()` or an `init()` function, and 
- that `TestMain()` must call `score.PrintTestInfo()` before executing the tests.

However, the benefit is that this new API should eliminate the problems related to panicking student code that causes some tests to not appear in the frontend, showing incorrect score results, because their JSON string is never emitted to stdout. Furthermore, obtaining a score object from within a test is easier.

The design of the API is based on the notion that
- All test names in one test process execution are unique; that is, you cannot reuse the same test names across packages being tested from one test execution.
- This is because there is a single registry of tests per execution of `go test -v ./...`

An alternative design was considered, where one would allow multiple registries, but this was deemed to add complexity for no good reason. However, I'm willing to discuss this, if anyone can show its merits.

meling:
I've discovered a problem with the current design that has me rethinking the decision to use a single registry. Can others spot the problem too?

meling:
Have implemented a fix that should prevent non-test code from looking up the score object. This should fix my concern.

0xf8f8ff:
The only thing which is not quite working on the existing database is that the `Members` tab is empty for all courses, the reason is `ERROR   web/autograder_service.go:347   GetEnrollmentsByCourse failed: submission missing build information`, which is probably expected?

The rest seems to work just fine.

database/temp_transformation.go
```diff
@@ -0,0 +1,48 @@
...
+			sub.BuildInfo = buildInfo
+			sub.Scores = scores
+			sub.OldBuildInfo = ""
+			sub.ScoreObjects = ""
+		}
+	}
+}
+
+func unmarshalBuildInfo(submission *pb.Submission) *score.BuildInfo {
+	if submission.OldBuildInfo == "" {
```
0xf8f8ff, line 29: But won't this `OldBuildInfo` always be empty in the old database? I mean, a Submission in the old database will still have its jsonified build info in the `buildInfo` column, and every old Submission will have an empty `oldBuildInfo` . Are we supposed to rename the column manually in the old database or am I missing something completely?

meling:
Note that I’m working on a grpc-based score service that I originally started this pr for. Still not pushed because it involves some more integration. So I’m holding off merging this pr for a few more days…

database/temp_transformation.go
meling: Another idea is to just load the database and use these transformations and write back things to a new database file that uses the new fields/columns. (As a one time task…)

0xf8f8ff: Yea, you sure can edit a column name in the database file, probably just worth mentioning somewhere that this is a necessary step to get things compatible.

meling:
> > Note that I’m working on a grpc-based score service that I originally started this pr for. Still not pushed because it involves some more integration. So I’m holding off merging this pr for a few more days…
> 
> I'm starting working on converting json string based reviews/benchmarks to database entries and would rather work with the already updated database schema, do you think it is ok to base it off of this PR?

Yeah, that’s probably the best solution, since this already lays much of the ground work.

kit/score/score.proto
```diff
@@ -0,0 +1,34 @@
...
+package score;
+option go_package = "github.com/autograde/quickfeed/kit/score";
+
+import "patch/go.proto";
+
+// Score give the score for a single test named TestName.
+message Score {
+    uint64 ID = 1;
+    uint64 SubmissionID = 2;
+    string Secret = 3;       // the unique identifier for a scoring session
```
0xf8f8ff, line 11: Would it possibly be better to add a tag to get this field ignored by the database? Because right now we have a `Secret` column filled with the same value: "hidden" for each and every score, seems unnecessary to put it in the database at all.

PR #395, opened 2021-01-30
Add ansible playbooks

r0qs:
Add initial playbooks to setup quickfeed in a docker environment for development or in a server machine for production.
Still missing the nginx and envoy configs that will be added in future commits.


meling:
Playing with this stuff, I ran:
```sh
% ansible-playbook dev.yml

PLAY [development] *********************************************************************************************************************************************

TASK [Gathering Facts] *****************************************************************************************************************************************
ok: [localhost]

TASK [build_image : Create temp build directory] ***************************************************************************************************************
ok: [localhost]

TASK [build_image : Apply dockerfile template] *****************************************************************************************************************
ok: [localhost]

TASK [build_image : Create quickfeed image] ********************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to import the required Python library (Docker SDK for Python: docker (Python >= 2.7) or docker-py (Python 2.6)) on heins-cyclone.lan's Python /Applications/Xcode.app/Contents/Developer/usr/bin/python3. Please read the module documentation and install it in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter, for example via `pip install docker` or `pip install docker-py` (Python 2.6). The error was: No module named 'requests'"}
```
Does it require old Python 2.7? I ran `pip install docker`, after the error, but got the same error, when running again.

meling:
I believe we don't need the `GO111MODULE` env var for this stuff.

ansible/inventory/group_vars/all/default.yml
```diff
@@ -0,0 +1,17 @@
...
+autograder_group: autograder
+autograder_home: /home/autograder
+
+quickfeed_root: /home/autograder/quickfeed
+quickfeed_ui_root: /home/autograder/quickfeed/public
+quickfeed_bin: /home/autograder/go/bin
+quickfeed_repo: https://github.com/r0qs/quickfeed.git
+quickfeed_branch: dev-tooling
+
+# Define the real values in the production config that should never be committed.
```
meling, line 15: Should these be defined in this file, or only as env vars?

ansible/inventory/group_vars/development/default.yml
```diff
@@ -0,0 +1,4 @@
+---
+docker_build: true
+quickfeed_endpoint: dev.itest.run
+ag_port: 8090
```
meling, line 4: Should be changed to `http_addr` or something, because it is really the `-http.addr` flag for the https endpoint of the echo framework. (The plan is to refactor away the echo framework at some point... but we will probably still need a `http.addr` flag.)

ansible/roles/base/tasks/setup_go.yml
```diff
@@ -0,0 +1,13 @@
...
+  - name: Define GOROOT variable
+    ansible.builtin.set_fact:
+      GOROOT: "{{ go_root }}"
+    when: GOROOT is not defined
+
+  - name: Save GOROOT configuration
+    ansible.builtin.copy:
+      content: |
+        export PATH=$PATH:{{ go_root }}/bin
+        export GO111MODULE=on
```
meling, line 11: The `GO111MODULE` should no longer be needed.

ansible/roles/build_image/templates/Dockerfile.j2
```diff
@@ -0,0 +1,7 @@
+FROM {{ builder.image.base }}
```
meling, line 1: The Docker plugin for VSCode complains about this line: 
```text
Second argument should be AS
invalid name for build stage: "}}", name can't start with a number or contain symbols
```
Not sure if this is just the plugin or if this is a real problem.

ansible/roles/quickfeed/tasks/main.yml
```diff
@@ -0,0 +1,106 @@
...
+  - name: Define GO111MODULE facts
+    become_user: "{{ quickfeed_user }}"
+    ansible.builtin.set_fact:
+      GO111MODULE: "on"
+    when: GO111MODULE is not defined
+
+  # go dependencies
+  - name: Get Go dependencies packages
+    become_user: "{{ quickfeed_user }}"
+    ansible.builtin.command: go get -u {{ item }}
```
meling, line 45: The recommended way to install go dependencies is to use `go install {{ item }}`.

ansible/roles/quickfeed/tasks/main.yml
```diff
@@ -0,0 +1,106 @@
...
+      GOPATH: "{{ quickfeed_home }}/go"
+    when: GOPATH is not defined
+  
+  - name: Define PATH facts
+    become_user: "{{ quickfeed_user }}"
+    ansible.builtin.set_fact:
+      PATH: "{{ ansible_env.PATH }}:{{ GOPATH }}/bin"
+    when: GOPATH is defined
+
+  - name: Define GO111MODULE facts
```
meling, line 36: Don't think this is needed anymore.

ansible/roles/quickfeed/tasks/main.yml
```diff
@@ -0,0 +1,106 @@
...
+
+  - name: Build quickfeed source
+    become_user: "{{ quickfeed_user }}"
+    ansible.builtin.shell: "cd {{ quickfeed_root }} && make install"
+    environment:
+      PATH: "{{ PATH }}"
+      GOPATH: "{{ GOPATH }}"
+      GO111MODULE: "{{ GO111MODULE }}"
+
+  - name: Build gosecret command
```
meling, line 100: The `gosecret` command is no longer need for quickfeed.

ansible/inventory/group_vars/all/default.yml
```diff
@@ -0,0 +1,17 @@
...
+hostname: ag
+docker_build: false
+
+quickfeed_user: autograder
+quickfeed_group: autograder
+quickfeed_home: /home/autograder
+
+quickfeed_root: {{ quickfeed_home }}/quickfeed
+quickfeed_ui_root: {{ quickfeed_root }}/public
+quickfeed_bin: {{ quickfeed_home }}/go/bin
```
meling, line 11: The `quickfeed_bin` is the same as `GOPATH/bin` and `GOPATH/bin` is added to `PATH` by `name: Define PATH facts`. Wouldn't it be enough to use `PATH` in `ExecStart`?

ansible/inventory/group_vars/all/default.yml
r0qs: These are the default values for all group of hosts (dev or prod), and they can be overwritten in run time. But The values are wrong. It should point the to autograde organization and master branch.

ansible/roles/build_image/templates/Dockerfile.j2
r0qs: I guess is just the plugin. I did not have such problem.

ansible/inventory/group_vars/all/default.yml
r0qs: True, I'll change that.

ansible/roles/quickfeed/tasks/main.yml
r0qs: But `go install` will download the package? I thought that the install command just install local packages (i.e. packages that already exists in GOPATH), and that `go get` downloads and installs the packages. And in this case, I need to download the packages.

https://golang.org/pkg/cmd/go/internal/get/#pkg-variables

meling: Yes. `go install` will also download the package or command/tool. For example, in this case, I believe it is being used to download and install various tools for grpc and protobuf. This [thread has a nice solution](https://github.com/go-modules-by-example/index/issues/77) for using the `tools.go` approach to specify the dependencies; meaning that make targets like this:

```sh
cat tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go install %
```
can be replaced with something like this:
```sh
go install $(go list -f "{{range .Imports}}{{.}} {{end}}" tools.go)
```
Running this:
```sh
go list -f "{{range .Imports}}{{.}} {{end}}" tools.go
```
should give you the list of tools to install.

ansible/inventory/group_vars/all/default.yml
meling: I think I made a mistake when commenting; I clicked on line 16, referring to these:
```ansible
quickfeed_github_key: <ghkey>
quickfeed_github_secret: <ghsecret>
```
I guess these can be in said file, as long as we don't commit it.

ansible/roles/quickfeed/tasks/main.yml
r0qs: Oh! Cool, I didn't know that. I'll do in this way then :)

ansible/inventory/group_vars/all/default.yml
r0qs: Yes, we can also define in the `inventory/hosts` or pass it as parameters when deploying. I just declare them because they were used here when running the daemon: https://github.com/autograde/quickfeed/blob/b71aec1f1123f277596c35a9cb789aa8ad58b1ce/ansible/roles/quickfeed/templates/systemd/quickfeed.service.d/agenv.conf.j2

r0qs:
Closing in favor of https://github.com/autograde/quickfeed/pull/382

PR #397, opened 2021-02-01
UpdateSlipdays with a graceperiod

Bibben:
Proposed implementation of UpdateSlipdays() including a graceperiod. 
The grace period is currently set to 2 hours. can and should be an int between [0,1,...,23].

This change can affect all assignments, that are not yet approved, on the currently running autograder service.

Currently doesn't support 1,5 hours grace-periods, using minutes instead grants this functionality.
`const gracePeriod time.Duration = time.Duration(90 * time.Minutes)`

### DIFF
--- a/ag/enrollment.go
+++ b/ag/enrollment.go
@@ -5,6 +5,12 @@ import (
 	"time"
 )
 
+const (
+	//graceP is the grace-Period for deliveries after the deadline. It can only be <int> hours.
+	//Currently there is a 2 hour grace period. It should be between [0,1,..,23]
+	graceP = time.Duration(2 * time.Hour)
+)
+
 // UpdateSlipDays updates the number of slipdays for the given assignment/submission.
 func (m *Enrollment) UpdateSlipDays(start time.Time, assignment *Assignment, submission *Submission) error {
 	if m.GetCourseID() != assignment.GetCourseID() {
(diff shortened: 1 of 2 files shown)


ag/enrollment.go
```diff
@@ -5,6 +5,12 @@ import (
 	"time"
 )
 
+//gracePeriod is the grace-Period for deliveries after the deadline. It can only be <int> hours.
```
meling, line 8: Please use space after `//` comments; using the `gofumpt` formatter should fix this when you save.
Please write "grace period" without the hyphen.
Please combine the limitations into one sentence, e.g. "The grace period must be in the range 0-23 hours."
You can remove the "current 2 hour" comment. It is clear from the code.

ag/enrollment.go
```diff
@@ -5,6 +5,12 @@ import (
...
+//Currently there is a 2 hour grace period. It should be between [0,1,..,23]
+const gracePeriod time.Duration = time.Duration(2 * time.Hour)
+
+//Keep in mind this will be the same across all enrollments (courses) on the AG service.
```
meling, line 12: Please move this comment above the definition of the grace period, and rephrase, e.g. "Note that the grace period applies to all enrollments (courses)."

ag/enrollment.go
```diff
@@ -20,7 +26,13 @@ func (m *Enrollment) UpdateSlipDays(start time.Time, assignment *Assignment, sub
 	// if score is less than limit and it's not yet approved, update slip days if deadline has passed
 	if submission.Score < assignment.ScoreLimit && submission.Status != Submission_APPROVED && sinceDeadline > 0 {
 		// deadline exceeded; calculate used slipdays for this assignment
-		m.updateSlipDays(assignment.GetID(), uint32(sinceDeadline/days))
+		slpDays, slpHours := uint32(sinceDeadline/days), sinceDeadline%days
```
meling, line 29: Could you please add a few tests for this feature? There are some tests in `slipdays_test.go` that can be expanded with tests for grace period.

ag/enrollment.go
```diff
@@ -20,7 +26,13 @@ func (m *Enrollment) UpdateSlipDays(start time.Time, assignment *Assignment, sub
...
+		//slpHours is time after deadline. This is also correct on subsequent slipdays after deadline has passed
+		if slpHours > gracePeriod {
+			slpDays++
+		}
+		//updating the slipdays
```
meling, line 34: Please remove useless comments ;-)

ag/enrollment.go
meling: PS: The logic Looks good to me! But it is always good to have tests for this stuff.

PR #401, opened 2021-02-06
Fix for submissions without automated tests

0xf8f8ff:
- fixes the issue where `SkipTests` status of an assignment would not update properly in the database
- hides the table with empty test results when displaying the build results for such assignments

### DIFF
--- a/database/gormdb_assignment.go
+++ b/database/gormdb_assignment.go
@@ -41,6 +41,7 @@ func (db *GormDB) CreateAssignment(assignment *pb.Assignment) error {
 			"is_group_lab":      assignment.IsGroupLab,
 			"reviewers":         assignment.Reviewers,
 			"container_timeout": assignment.ContainerTimeout,
+			"skip_tests":        assignment.SkipTests,
 		}).FirstOrCreate(assignment).Error
 }
 
--- a/web/hooks/github.go
+++ b/web/hooks/github.go
@@ -1,6 +1,7 @@
 package hooks
 
 import (
+	"encoding/json"
 	"net/http"
 	"strings"
 	"time"
@@ -167,6 +168,7 @@ func (wh GitHubWebHook) runAssignmentTests(assignment *pb.Assignment, repo *pb.R
 		JobOwner:   payload.GetSender().GetLogin(),
 	}
 	if assignment.SkipTests {
+		wh.logger.Debugf("Assignment %s for course %s has no automated test, skipping tests", assignment.Name, course.Name)
 		wh.recordSubmissionWithoutTests(runData)
 		return
 	}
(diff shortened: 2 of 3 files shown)


database/gormdb_assignment.go
```diff
@@ -41,6 +41,7 @@ func (db *GormDB) CreateAssignment(assignment *pb.Assignment) error {
 			"is_group_lab":      assignment.IsGroupLab,
 			"reviewers":         assignment.Reviewers,
 			"container_timeout": assignment.ContainerTimeout,
+			"skip_tests":        assignment.SkipTests,
```
meling, line 44: I've been confused about the naming of this field since I first saw it. Doesn't this just mean `manual`? I mean as in manually reviewed assignment. I'm okay to leave this as-is for now in this PR, but I think we should change the name. As it is now, it sounds as if there are tests for this assignment, but they are to be skipped.

web/hooks/github.go
```diff
@@ -167,6 +168,7 @@ func (wh GitHubWebHook) runAssignmentTests(assignment *pb.Assignment, repo *pb.R
 		JobOwner:   payload.GetSender().GetLogin(),
 	}
 	if assignment.SkipTests {
+		wh.logger.Debugf("Assignment %s for course %s has no automated test, skipping tests", assignment.Name, course.Name)
```
meling, line 171: Suggested phrasing: `Assignment %s for course %s is manually reviewed.`

web/hooks/github.go
```diff
@@ -176,8 +178,18 @@ func (wh GitHubWebHook) runAssignmentTests(assignment *pb.Assignment, repo *pb.R
...
 // for a manually graded assignment.
 func (wh GitHubWebHook) recordSubmissionWithoutTests(data *ci.RunData) {
+	noTestBuildInfo, err := json.Marshal(&ci.BuildInfo{
+		BuildID:   0,
+		BuildDate: time.Now().Format("2006-01-02T15:04:05"),
+		BuildLog:  "No automated tests for this assignment",
+		ExecTime:  1,
+	})
+	if err != nil {
+		wh.logger.Errorf("Error marshalling build info for %s of course %s for student %s", data.Course.Name, data.Assignment.Name, data.JobOwner)
```
meling, line 188: Does it make sense to record this in the database if it couldn't marshal the build info, or should we return after logging the error?

web/hooks/github.go
```diff
@@ -176,8 +178,18 @@ func (wh GitHubWebHook) runAssignmentTests(assignment *pb.Assignment, repo *pb.R
...
+	}
 	newSubmission := &pb.Submission{
 		AssignmentID: data.Assignment.ID,
+		BuildInfo:    string(noTestBuildInfo),
```
meling, line 192: (Comment not related to this line, but to Lines 198-200): Suggest returning after the error has been logged and remove the else clause. Further, the log statement on Line 200 should say:
```
Saved manual review submission for user %s for assignment %d.
```

0xf8f8ff: Fixed.

web/hooks/github.go
0xf8f8ff: Right, made it return on error.

web/hooks/github.go
0xf8f8ff: Done.

database/gormdb_assignment.go
0xf8f8ff: Honestly, both `skip_tests` and `manual` sound equally confusing to me out of context. I can change it to `manual`, but I'd rather do it later to avoid delays caused by necessary changes to yaml files for all related assignments, as there is a lab deadline closing in now for one of those.

PR #408, opened 2021-02-10
Replace deprecated jsonpb package

### LINKED ISSUE #407: github.com/golang/protobuf/jsonpb is deprecated
Replace `jsonpb` with another marshaling mechanism. Add a few tests for this, before replacing. It is causing the golangci-lint to fail.
```
SA1019: package github.com/golang/protobuf/jsonpb is deprecated: Use the "google.golang.org/protobuf/encoding/protojson" package instead. (staticcheck)
```

meling:
This commit adds tests for the Review type's marshal and unmarshals functions. This also removes the protobuf/jsonpb package since it is being deprecated and is causing GitHub workflow actions to complain. This fixes #407.


meling:
@0xf8f8ff Note that the standard library `json` package gives incompatible output compared to the previously used `protobuf/jsonpb`. But in principle, I don't think this version is any worse than the other for the purpose it is being used. That said, you probably don't want to merge this PR until some safe point, when you don't need to extract these strings from the database.

Alternatively, you could make a small script that translates the database content; this should be fairly trivial to do; 
- just extract stuff from the database (for all users that has such data)
- convert to the new json format
- write back to the database

At any rate, it would be very nice to have this merged sooner rather than later because it is annoying that this deprecation notice make our tests fail.

meling:
Have added a simple conversion step using a regular expression to remove quotes around JSON number fields to make it compatible with the JSON format accepted by the stdlib json package.

PR #413, opened 2021-02-22
Revised score api again

meling:
This PR introduces a few additional improvements for recording score objects so that we can more easily validate the fields and compute the sum. This also revises the registry API to be more flexible.


meling:
Closing in favor of #384, since it overlaps.

PR #421, opened 2021-04-12
Option to call GetSubmissions with less data to be transmitted

oskargj:
- Can be used to reduce rendering time for the result page.
- Introduces the skipBuildInfo field in SubmissionRequest.
- BuildInfo, ScoreObjects, and Review data is not transmitted when skipBuildInfo is true.


database/gormdb_assignment.go
```diff
@@ -123,6 +123,19 @@ func (db *GormDB) GetCourseAssignmentsWithSubmissions(courseID uint64, submissio
 	return filteredAssignments, nil
 }
 
+// GetCourseAssignmentsWithSubmissionsNoBuildInfo
+// returns data required for results page (score and status)
+func (db *GormDB) GetCourseAssignmentsWithSubmissionsNoBuildInfo(courseID uint64, submissionType pb.SubmissionsForCourseRequest_Type) ([]*pb.Assignment, error) {
+	var assignments []*pb.Assignment
+
+	if err := db.conn.Preload("Submissions").Where(&pb.Assignment{CourseID: courseID}).Order("order").Find(&assignments).Error; err != nil {
+		fmt.Println(err.Error())
```
meling, line 132: Should not print errors, since it is returned and printed by the caller or some other caller up the call stack.

database/gormdb_assignment.go
```diff
@@ -123,6 +123,19 @@ func (db *GormDB) GetCourseAssignmentsWithSubmissions(courseID uint64, submissio
...
+		return nil, err
+	}
+
```
meling, line 135: Avoid blank line.

database/gormdb_assignment.go
```diff
@@ -123,6 +123,19 @@ func (db *GormDB) GetCourseAssignmentsWithSubmissions(courseID uint64, submissio
 	return filteredAssignments, nil
 }
 
+// GetCourseAssignmentsWithSubmissionsNoBuildInfo
+// returns data required for results page (score and status)
+func (db *GormDB) GetCourseAssignmentsWithSubmissionsNoBuildInfo(courseID uint64, submissionType pb.SubmissionsForCourseRequest_Type) ([]*pb.Assignment, error) {
+	var assignments []*pb.Assignment
+
```
meling, line 130: Prefer to avoid blank lines, unless function is long, and there are natural reasons to add some spacing.

web/courses.go
```diff
@@ -148,7 +148,14 @@ func (s *AutograderService) getSubmissions(request *pb.SubmissionRequest) (*pb.S
...
 func (s *AutograderService) getAllCourseSubmissions(request *pb.SubmissionsForCourseRequest) (*pb.CourseSubmissions, error) {
-	assignments, err := s.db.GetCourseAssignmentsWithSubmissions(request.GetCourseID(), request.Type)
+	var assignments []*pb.Assignment
+	var err error
+	if request.GetSkipBuildInfo() {
+		assignments, err = s.db.GetCourseAssignmentsWithSubmissionsNoBuildInfo(request.GetCourseID(), request.Type)
+	} else {
+		assignments, err = s.db.GetCourseAssignmentsWithSubmissions(request.GetCourseID(), request.Type)
+	}
+
```
meling, line 158: The above code could be made a bit more nice like this:

```go
var getCourseSubFn func(uint64, pb.SubmissionsForCourseRequest_Type) ([]*pb.Assignment, error)
if request.GetSkipBuildInfo() {
	getCourseSubFn = s.db.GetCourseAssignmentsWithSubmissionsNoBuildInfo
} else {
	getCourseSubFn = s.db.GetCourseAssignmentsWithSubmissions
}
assignments, err := getCourseSubFn(request.GetCourseID(), request.Type)
if err != nil {
	return nil, err
}
```

meling (changes requested):
Looks mostly good. I didn't check how it actually works. If you think this can be taken out of draft mode and merged, please let me know.

PS: I also like to add some tests to the go code itself to make sure that things don't break other stuff. But not sure what exactly to test for this one. Maybe that the returned data doesn't contain things that it shouldn't...

meling (approved):

PR #422, opened 2021-04-13
Localhost instructions

plakolaki:
- Setup instructions for localhost use. 
- Runs with NGINX and Envoy.
- Script used to set up NGINX and SSL certification for localhost, created by @Bibben.


meling:
Could you move the files into the `docs` folder, please? You can keep a sub-directory called e.g. `local-setup`

PR #448, opened 2021-06-18
Specifying port number in GRPCManager

oskargjolga:


meling:
The plan is to make this documentation item obsolete. Right now, the `grpcPort` must still be hardcoded in the `GRPCManager.ts,` but hopefully, we can do this in some other way, e.g., by importing the `grpcPort` from a `.ts` file or something that can also be used to configure the Envoy proxy and so on. Thus, leaving this open until we have a proper solution, so we don't forget.

meling:
Closing this as obsoleted by new code and envoy configurations.

PR #449, opened 2021-06-22
Basic authentication & updated envoy config

JosteinLindhom:
This PR introduces an updated envoy configuration that allows gRPC and regular HTTP communication to occur on a single port. This should remove the need for documentation introduced in #448 .

The PR also introduces basic authentication by session token.

### DIFF
--- a/ag/validation.go
+++ b/ag/validation.go
@@ -59,7 +59,7 @@ func Interceptor(logger *zap.Logger, userMap map[string]uint64) grpc.UnaryServer
 		if !ok {
 			return nil, errors.New("Could not grab metadata from context")
 		}
-		meta, err := UserValidation(meta, userMap)
+		meta, err := userValidation(meta, userMap)
 		if err != nil {
 			return nil, err
 		}
@@ -82,20 +82,20 @@ func Interceptor(logger *zap.Logger, userMap map[string]uint64) grpc.UnaryServer
 	}
 }
 
-// Returns modified metadata containing a valid user. Returns an error if the user is not authenticated.
-func UserValidation(meta metadata.MD, userMap map[string]uint64) (metadata.MD, error) {
-	token := meta.Get(Cookie)
+// userValidation returns modified metadata containing a valid user. An error is returned if the user is not authenticated.
+func userValidation(meta metadata.MD, userMap map[string]uint64) (metadata.MD, error) {
+	cookie := meta.Get(Cookie)
 
-	if len(token) > 0 {
-		token := token[0]
-		user := userMap[token]
-		if user == 0 {
-			return nil, errors.New("Could not associate token with a user")
-		}
-		meta.Set(UserKey, strconv.FormatUint(user, 10))
-	} else {
+	if len(cookie) == 0 {
 		return nil, errors.New("Request does not contain a session token")
 	}
+	token := cookie[0]
+	user := userMap[token]
+	if user == 0 {
+		return nil, errors.New("Could not associate token with a user")
+	}
+	meta.Set(UserKey, strconv.FormatUint(user, 10))
+
 	return meta, nil
 }
 
--- a/ag/validation.go
+++ b/ag/validation.go
@@ -84,19 +84,13 @@ func Interceptor(logger *zap.Logger, userMap map[string]uint64) grpc.UnaryServer
 
 // userValidation returns modified metadata containing a valid user. An error is returned if the user is not authenticated.
 func userValidation(meta metadata.MD, userMap map[string]uint64) (metadata.MD, error) {
-	cookie := meta.Get(Cookie)
-
-	if len(cookie) == 0 {
-		return nil, errors.New("Request does not contain a session token")
+	for _, cookie := range meta.Get(Cookie) {
+		if user := userMap[cookie]; user > 0 {
+			meta.Set(UserKey, strconv.FormatUint(user, 10))
+			return meta, nil
+		}
 	}
-	token := cookie[0]
-	user := userMap[token]
-	if user == 0 {
-		return nil, errors.New("Could not associate token with a user")
-	}
-	meta.Set(UserKey, strconv.FormatUint(user, 10))
-
-	return meta, nil
+	return nil, errors.New("Request does not contain a valid session cookie.")
 }
 
 // IsValid on void message always returns true.
--- a/ag/validation.go
+++ b/ag/validation.go
@@ -15,8 +15,6 @@ import (
 
 // MaxWait is the maximum time a request is allowed to stay open before aborting.
 const MaxWait = 2 * time.Minute
-const Cookie = "cookie"
-const UserKey = "user"
 
 type validator interface {
 	IsValid() bool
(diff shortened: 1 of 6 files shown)


web/auth/auth.go
```diff
@@ -323,6 +326,22 @@ func AccessControl(logger *zap.Logger, db database.Database, scms *Scms) echo.Mi
 				return echo.NewHTTPError(http.StatusBadRequest, err)
 			}
 
+			token, err := c.Cookie("session")
```
meling, line 329: Use const `SessionKey`

web/auth/auth.go
```diff
@@ -323,6 +326,22 @@ func AccessControl(logger *zap.Logger, db database.Database, scms *Scms) echo.Mi
...
+			if err != nil {
+				return err
+			}
+
+			// Check if user exists in map with an old session token
```
meling, line 334: This code could be broken out as 
- a function that takes `token` as input or 
- a method that operates on a struct that encapsulates the `TokenToUserMap` map

The latter version may require some additional refactoring. The next comments may give more clues which is better.

Ideally, we should avoid exporting a map directly like this. It would be better to use functions in the `auth` package to control the map accesses.

web/auth/auth.go
```diff
@@ -45,6 +45,9 @@ func newUserSession(id uint64) *UserSession {
 	}
 }
 
+// TokenToUserMap
+var TokenToUserMap = make(map[string]uint64)
```
meling, line 49: Ideally, we want to avoid exporting the map directly. Instead, map accesses should be done via methods or functions controlled from this file.

main.go
```diff
@@ -114,7 +114,7 @@ func main() {
 	if err != nil {
 		log.Fatalf("failed to start tcp listener: %v\n", err)
 	}
-	opt := grpc.UnaryInterceptor(pb.Interceptor(logger))
+	opt := grpc.UnaryInterceptor(pb.Interceptor(logger, auth.TokenToUserMap))
```
meling, line 117: This argument to interceptor is not necessary since `auth.TokenToUserMap` is accessible also from inside the interceptor implementation. That said, it might be that we do want to pass something into the interceptor from main if we had created some users session map in main. Right now, it is not clear what is the better solution.

meling: On another note. I've been meaning to rewrite the current interceptor, splitting the logic into separate interceptors, and use chaining of the different interceptors. See [ChainUnaryInterceptor](https://pkg.go.dev/google.golang.org/grpc#ChainUnaryInterceptor).

ag/validation.go
```diff
@@ -49,6 +52,23 @@ func Interceptor(logger *zap.Logger) grpc.UnaryServerInterceptor {
 		}
 		ctx, cancel := context.WithTimeout(ctx, MaxWait)
 		defer cancel()
+
+		meta, ok := metadata.FromIncomingContext(ctx)
```
meling, line 56: Ideally, this logic should be in a separate interceptor. See comment in main.go.

ag/validation.go
```diff
@@ -49,6 +52,23 @@ func Interceptor(logger *zap.Logger) grpc.UnaryServerInterceptor {
...
+		if !ok {
+			return nil, errors.New("Could not grab metadata from context")
+		}
+
+		if len(meta.Get("cookie")) > 0 {
```
meling, line 61: Do this only once:
```golang
token := meta.Get(Cookie)
```
Define `Cookie` as a const. 

Then you can do `if len(token) > 0` and save a lookup in meta.

ag/validation.go
```diff
@@ -49,6 +52,23 @@ func Interceptor(logger *zap.Logger) grpc.UnaryServerInterceptor {
...
+			token := meta.Get("cookie")[0]
+			if userMap[token] == 0 {
+				return nil, errors.New("Could not associate token with a user")
+			}
+			meta.Set("user", strconv.FormatUint(userMap[token], 10))
```
meling, line 66:
```golang
strconv.FormatUint(user, 10))
```

meling: Maybe also the `"user"` string should be made a const, if used several places.

ag/validation.go
```diff
@@ -49,6 +52,23 @@ func Interceptor(logger *zap.Logger) grpc.UnaryServerInterceptor {
...
+			ctx = metadata.NewOutgoingContext(ctx, meta)
+		} else {
+			return nil, errors.New("x")
```
meling, line 69: Error message?

ag/validation.go
```diff
@@ -49,6 +52,23 @@ func Interceptor(logger *zap.Logger) grpc.UnaryServerInterceptor {
...
 		defer cancel()
+
+		meta, ok := metadata.FromIncomingContext(ctx)
+		if !ok {
+			return nil, errors.New("Could not grab metadata from context")
+		}
+
+		if len(meta.Get("cookie")) > 0 {
+			token := meta.Get("cookie")[0]
+			if userMap[token] == 0 {
```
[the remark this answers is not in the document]
JosteinLindhom: `meta.Get(...)` returns a `[]string` as a request could include more than one cookie.

ag/validation.go
```diff
@@ -85,6 +82,23 @@ func Interceptor(logger *zap.Logger, userMap map[string]uint64) grpc.UnaryServer
...
+	token := meta.Get(Cookie)
+
+	if len(token) > 0 {
+		token := token[0]
+		user := userMap[token]
+		if user == 0 {
+			return nil, errors.New("Could not associate token with a user")
+		}
+		meta.Set(UserKey, strconv.FormatUint(user, 10))
+	} else {
```
meling, line 96: Seeing the logic now, you can write this code more clearly like this, avoiding else and indentations (this is recommended Go-style):

```golang
if len(token) == 0 {
 	return nil, errors.New("Request does not contain a session token")
}
token := token[0]
user := userMap[token]
if user == 0 {
 	return nil, errors.New("Could not associate token with a user")
}
meta.Set(UserKey, strconv.FormatUint(user, 10))
return meta, nil
```

ag/validation.go
```diff
@@ -85,6 +82,23 @@ func Interceptor(logger *zap.Logger, userMap map[string]uint64) grpc.UnaryServer
 	}
 }
 
+// Returns modified metadata containing a valid user. Returns an error if the user is not authenticated.
+func UserValidation(meta metadata.MD, userMap map[string]uint64) (metadata.MD, error) {
```
meling, line 86: Maybe this function doesn't need to be exported??

If the export is useful, then the comment should be phrased like this:
```golang
// UserValidation returns modified ... An error is returned if the user ...
```

This is the recommended Go-style for comments, to always begin with the function name.

ag/validation.go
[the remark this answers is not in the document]
meling: So, does that mean you would want to check all cookies if one of them is in the `userMap` or just the first one? If the former, you could avoid some checks, e.g. like this:

```golang
for _, cookie := range meta.Get(Cookie) {
 	if user := userMap[cookie]; user > 0 {
 	 	meta.Set(UserKey, strconv.FormatUint(user, 10))
 	 	return meta, nil
 	}
}
return nil, errors.New("Request does not contain a session cookie")
```

meling (approved):

PR #452, opened 2021-06-23
Updated npm and various typescript issues

meling:
This also fixes a recent dependabot security issue in set-getter.

### DIFF
--- a/public/src/pages/views/ReleaseView.tsx
+++ b/public/src/pages/views/ReleaseView.tsx
@@ -165,11 +165,11 @@ export class ReleaseView extends React.Component<ReleaseViewProps, ReleaseViewSt
                             courseURL={this.props.courseURL}
                             isSelected={this.state.selectedGroup === grp}
                             setGrade={async (status: Submission.Status, approved: boolean) => {
-                                const current = this.state.submissionsForGroupAssignment.get(grp);
-                                if (current && current.submission) {
-                                    current.submission.status = status;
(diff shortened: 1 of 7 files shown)


public/src/pages/views/ReleaseView.tsx
```diff
@@ -178,10 +178,9 @@ export class ReleaseView extends React.Component<ReleaseViewProps, ReleaseViewSt
                                 if (current) {
                                     current.released = release;
                                     current.score = scoreFromReviews(current.reviews);
-                                    const ans = this.props.onUpdate(current);
-                                    if (ans) return true;
+                                    const ans = await this.props.onUpdate(current);
+                                    if (ans) return;
```
0xf8f8ff, line 182: The component calling this function expects a boolean, will it return a boolean?

meling: It expects a boolean input (`release`), but a `void` return value. If you hover over the `release` JSX attribute on line 176 (the `release` before the `=` sign), then it will show the `void`. Also, I would expect that this wouldn't compile otherwise. Although, I don't understand why it compiled with the `return true;` statements that were used before...

meling: Does it not compile on your end?

0xf8f8ff: Oh right, my bad, got lost in the methods. No, everything compiles just fine.

PR #456, opened 2021-06-25
Fixes docker issues

### LINKED ISSUE #287: Docker image not found (error msg: test execution failed: repository name must be canonical)
PR #281 depends on our custom `quickfeed:go` docker image. This has not been uploaded to docker hub, and will not be found, resulting in the following error message in `ag.log`:

```log
2020-07-26T00:11:07.040+0200    [ERROR]      ci/run_tests.go:47      Failed to run tests: %!w(*fmt.wrapError=&{test execution failed: repository name must be canonical 0xc000110ea0})
```

The solution is to run:

```sh
make docker
```

This will take a few moments, but there is no need to restart the server after this. The image will be found once it is available locally.

You can check that the image is available locally with one of these commands:

```sh
docker images
docker image ls quickfeed:go
```

Still thinking about what to do to solve this. 

1. We could upload it to Docker hub or
2. Just keep the docker images locally.

In the latter case, we may consider to add the `docker` make target as a dependency to the `install` target to ensure that every time we compile the server again, the docker images are rebuilt. That said, currently the `docker` target only builds the `quickfeed:go` docker image. If we do this, then we should make sure to build also other images that we need.

Alternatively, we could add the `make docker` part of the install instructions/setup script.

### LINKED ISSUE #400: Automate the pulling of a custom docker image when a newer image is available
Right now it has to be manually pulled by someone on autograder.

meling:
This upgrades docker to v20.10.7, fixes an issue that prevented pulling images and adds support for building image from Dockerfile on demand.

This upgrades docker, requiring that a few of our uses had to change accordingly due to API changes in docker's SDK.
This also adds support for building a new image if the requested image isn't locally available and cannot be pulled from docker.io.

Also fixes pulling images from docker that was previously broken, probably due to changes in docker's SDK.

Also upgrades uber/zap to v1.17 and adds support for logger in the docker client and provides a new log.Zap function to create quickfeed-specific logger.

Fixes #400.
Fixes #287.


meling:
Note to self: `docker.go:Run()` method is now quite long (106 lines). Break into smaller pieces and add more tests.

Issue #434, opened 2021-06-07
Semicolon in review text fails to decode


meling:
```json
{"ID":"60",
"points":"2",
"benchmarkID":"12",
"description":"\u003caside\u003e has a fixed width of 300px",
"grade":1,
"comment":"I see you have min-width: 300px, this does not specify a fixed width. Use width: 300px; instead"},
```

Note the `;` almost at the end of the `comment` text. This conflicts with the use of `;` as a separator character for different benchmarks for the review. The cause is this one (in `ag/review.go:32`):
```golang
	rs := strings.Split(review, ";")
```

The fix is to store the `GradingBenchmark` field in the database (remove the `sql` tag):
```proto
    repeated GradingBenchmark benchmarks = 8 [(gogoproto.moretags) = "sql:\"-\""];
```

This should be possible with some extra GORM work in the `database` package.
And remove the `review` field from `ag/ag.proto` and corresponding marshal/unmarshal code in `review.go`.

This is related to the removal of JSON for encoding stuff (see issue #429)

PR #460, opened 2021-07-01
Upgrading GORM to v2

### LINKED ISSUE #331: Upgrade to GORM v2
Review the [release notes](https://gorm.io/docs/v2_release_note.html) to understand how and if it impacts our use. Now uses import path `gorm.io/gorm`.

meling:
This migrates QuickFeed to use GORM v2.

Fixes #331.


database/gormdb_assignment.go
```diff
@@ -31,18 +29,8 @@ func (db *GormDB) CreateAssignment(assignment *pb.Assignment) error {
...
-			"deadline":          assignment.Deadline,
-			"auto_approve":      assignment.AutoApprove,
-			"score_limit":       assignment.ScoreLimit,
-			"is_group_lab":      assignment.IsGroupLab,
-			"reviewers":         assignment.Reviewers,
-			"container_timeout": assignment.ContainerTimeout,
-			"skip_tests":        assignment.SkipTests,
-		}).FirstOrCreate(assignment).Error
+		Assign(assignment).
+		FirstOrCreate(assignment).Error
```
meling, line 33: This change should be tested. We may need to use the map approach, since updating may not work if changing from boolean true to false or something... (zero-value issue).

But since this is a `CreateAssignment` (first version), maybe this isn't a problem.

0xf8f8ff:
When running on the old database `GetCourseLabSubmissions` and `GetEnrollmentsByCourse` methods fail with `unexpected end of JSON input`: 
```
DEBUG   web/autograder_service.go:544   GetCourseLabSubmissions: courseID:4
ERROR   web/autograder_service.go:548   GetCourseLabSubmissions failed: unexpected end of JSON input
DEBUG   web/access_control.go:70        (user=3, course=4) has enrollment status TEACHER
ERROR   web/autograder_service.go:347   GetEnrollmentsByCourse failed: unexpected end of JSON input
```
, looks like the old database still is not that compatible even with `DisableForeignKeyConstraintWhenMigrating`?
Which means that either a test database must be ready or the webhook issue resolved before I can properly test this PR.

meling:
> When running on the old database `GetCourseLabSubmissions` and `GetEnrollmentsByCourse` methods fail with `unexpected end of JSON input`:
> 
> ```
> DEBUG   web/autograder_service.go:544   GetCourseLabSubmissions: courseID:4
> ERROR   web/autograder_service.go:548   GetCourseLabSubmissions failed: unexpected end of JSON input
> DEBUG   web/access_control.go:70        (user=3, course=4) has enrollment status TEACHER
> ERROR   web/autograder_service.go:347   GetEnrollmentsByCourse failed: unexpected end of JSON input
> ```
> 
> , looks like the old database still is not that compatible even with `DisableForeignKeyConstraintWhenMigrating`?
> Which means that either a test database must be ready or the webhook issue resolved before I can properly test this PR.

It would be interesting to know what the JSON input actually is here.

Note: We have forgotten to update the log string; it should print `GetSubmissionsByCourse` instead of `GetCourseLabSubmissions` and the test should be renamed to `TestGetSubmissionsByCourse` instead of `TestGetCourseLabSubmissions`.

0xf8f8ff:
> It would be interesting to know what the JSON input actually is here.
> 
> Note: We have forgotten to update the log string; it should print `GetSubmissionsByCourse` instead of >`GetCourseLabSubmissions` and the test should be renamed to `TestGetSubmissionsByCourse` instead of >`TestGetCourseLabSubmissions`.

So the reason for json errors was actually the #434, otherwise everything works as expected. Updating method names as a new commit before merging.

PR #461, opened 2021-07-05
Removal of API endpoint & modification to gRPC auth

JosteinLindhom:
- Fetch self through gRPC request
- Updated solution to work with removal of API endpoint.
- Removed GetSelf & JSONuser struct.


web/auth/auth.go
```diff
@@ -17,7 +17,7 @@ import (
 
 func init() {
 	gob.Register(&UserSession{})
-	TokenStore = &UserToken{store: make(map[string]uint64)}
+	TokenStore = &UserToken{store: make(map[string]*http.Request)}
```
meling, line 20: This can be initialized at the point of declaration of `TokenStore`.

web/auth/auth.go
```diff
@@ -49,20 +49,27 @@ func newUserSession(id uint64) *UserSession {
...
+	if sess, err := gothic.Store.Get(ut.store[token], SessionKey); err == nil {
+		if i, ok := sess.Values[UserKey]; ok {
+			us := i.(*UserSession)
+			return us.ID
+		}
+	}
+	return 0
 }
 
 var TokenStore *UserToken
```
meling, line 75: Suggest to move this line between the `UserToken` struct declaration and the `add` method.

web/users.go
```diff
@@ -1,45 +1,9 @@
...
-	ID        uint64 `json:"id"`
-	IsAdmin   *bool  `json:"isadmin"`
-	Name      string `json:"name"`
-	StudentID string `json:"studentid"`
-	Email     string `json:"email"`
-	AvatarURL string `json:"avatarurl"`
-}
-
-// GetSelf redirects to GetUser with the current user's id.
-func GetSelf(db database.Database) echo.HandlerFunc {
```
meling, line 23: There is a `TestGetSelf` in `web/users_test.go` that now fails to compile. Can you rewrite `TestGetSelf` to work with the gRPC approach? You would probably need to call the `UserVerifier` interceptor or something, putting a dummy cookie/token into the context metadata, and checking that you get logged in. And similarly, that if you don't provide the required cookie, you won't be logged in.

public/src/managers/ServerProvider.ts
```diff
@@ -301,20 +301,13 @@ export class ServerProvider implements IUserProvider, ICourseProvider {
...
-        const iusr = result.data;
-        const usr = new User();
-        usr.setId(iusr.id);
-        usr.setStudentid(iusr.studentid);
-        usr.setName(iusr.name);
-        usr.setEmail(iusr.email);
-        usr.setAvatarurl(iusr.avatarurl);
-        usr.setIsadmin(iusr.isadmin);
-        return usr;
+        const iusr = result.data
```
meling, line 309: Here you can just `return result.data`

meling:
Is this ready to be reviewed in more detail? It is easy to fix the conflicting files mentioned above. You can just do:
```shell
git checkout master
git pull
gh pr checkout 461
git merge master
<fix the conflicting files>
git add <updated files>
git commit
git push
```

meling:
@JosteinLindhom ping :-)

web/auth/auth.go
```diff
@@ -47,28 +51,25 @@ func newUserSession(id uint64) *UserSession {
...
+func (us *UserSession) enableProvider(provider string) {
+	us.Providers[provider] = struct{}{}
 }
 
-func (ut *UserToken) add(token string, id uint64) {
-	for token, userID := range TokenStore.store {
-		if userID == id {
-			delete(TokenStore.store, token)
+// map from session cookies to user IDs.
+var tokenStore = make(map[string]uint64)
```
meling, line 59: Should we use the term `cookie` instead of `token` for this stuff? (including the `Add(token ...` method below)

meling (changes requested):

web/auth/auth.go
JosteinLindhom: Agreed, that would make more sense.

web/auth/auth.go
```diff
@@ -310,8 +327,8 @@ func AccessControl(logger *zap.Logger, db database.Database, scms *Scms) echo.Mi
 
 			i, ok := sess.Values[UserKey]
 			if !ok {
-				logger.Error(echo.ErrUnauthorized.Error())
-				return echo.ErrUnauthorized
+				// logger.Error(echo.ErrUnauthorized.Error())
```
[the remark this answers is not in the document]
JosteinLindhom: It should be removed.

Side note: I activated AccessControl for the whole site. Users not logged in, and logged in users should be allowed access to the website. Access control for data retreival is handled by TokenStore and the gRPC interceptor.

PR #464, opened 2021-07-07
Added IsManual() and HasTests() for Assignment type

meling:
This adds IsManual() and HasTests() for the Assignment type, which should allow us to replace the SkipTests field.
Although we may need to implement similar functionality on the different frontends.
Follow up commits should fix the current frontend code.


meling:
No longer relevant.

PR #466, opened 2021-07-16
Replacing manual grading stuff stored as json string by proper database records

### LINKED ISSUE #434: Semicolon in review text fails to decode

0xf8f8ff:
Grading benchmarks and criteria will no longer be stored as a single json string in the database. Instead, a database record will be created for each of those.

Ready, only to add/update a few tests.

Resolves #434.


meling:
I've merged #384 into master, so maybe master should be merged into this one, to more clearly show the actual changes in this PR. (I think that will do it...)

ag/ag.proto
```diff
@@ -243,15 +243,12 @@ message GradingCriterion {
...
-    uint64 score = 7;
-    // TODO(meling) Consider if benchmarks field is needed; it can be obtained from Assignments via submissionID??
-    // TODO(meling) If we need benchmarks here, there must be a foreign key `ReviewID` in `GradingBenchmark`; then we can also remove the gorm tag.
-    repeated GradingBenchmark benchmarks = 8 [(go.field) = {tags: 'gorm:"foreignKey:ID"'}];
-    string edited = 9;
+    uint64 ReviewerID = 3;   // UserID of the reviewer
+    string feedback = 4;
+    bool ready = 5;
+    uint64 score = 6;
+    repeated GradingBenchmark gradingBenchmarks = 7 [(go.field) = {tags: 'gorm:"foreignKey:ReviewID"'}];
```
meling, line 250: Do we need to specify the tag for foreign key here? If so, we should keep a doc comment explaining why.

Also, I was wondering if we need to repeat this field here since it could also be obtained via the assignments table via submissionID. If this doesn’t work, we should add a comment explaining it. Database ORMs are not easy, so it needs documentation…

database/gormdb_assignment.go
```diff
@@ -145,8 +144,8 @@ func (db *GormDB) CreateBenchmark(query *pb.GradingBenchmark) error {
 
 // UpdateBenchmark updates the given benchmark
 func (db *GormDB) UpdateBenchmark(query *pb.GradingBenchmark) error {
-	return db.conn.Model(query).
-		Where(&pb.GradingBenchmark{ID: query.ID, AssignmentID: query.AssignmentID}).
+	return db.conn.
+		Where(&pb.GradingBenchmark{ID: query.ID, AssignmentID: query.AssignmentID, ReviewID: query.ReviewID}).
```
meling, line 148: Can we add a test for this to satisfy Codecov?

database/gormdb_assignment.go
```diff
@@ -163,7 +162,7 @@ func (db *GormDB) CreateCriterion(query *pb.GradingCriterion) error {
 
 // UpdateCriterion updates the given criterion
 func (db *GormDB) UpdateCriterion(query *pb.GradingCriterion) error {
-	return db.conn.Model(query).
+	return db.conn.
```
meling, line 165: Can we add a test for this to satisfy Codecov?

kit/proto/score.proto
```diff
@@ -0,0 +1,20 @@
...
+    string TestName = 2; // name of the test
+    int32 Score = 3;     // the score obtained
+    int32 MaxScore = 4;  // max score possible to get on this specific test
+    int32 Weight = 5;    // the weight of this test; used to compute final grade
+    string Message = 6;  // error message; could be displayed in frontend also
+}
+
+message Void {}
+
+service ScoreService {
```
meling, line 17: This is a very old version of the kit stuff. How did this reappear?

kit/proto/generate.go
```diff
@@ -0,0 +1,2 @@
+//go:generate protoc --go_out=paths=source_relative:. --go-grpc_out=paths=source_relative:. score.proto
```
meling, line 1: The `kit/proto` directory is old code and should be deleted. It is not in master. Don't understand how this has reappeared.

kit/score/find_tests.go
```diff
@@ -0,0 +1,131 @@
+package score
```
meling, line 1: This file is also no longer part of the master branch.

kit/score/score.proto
```diff
@@ -31,4 +31,4 @@ message Results {
     uint64 ID = 1;
     BuildInfo BuildInfo = 2 [(go.field) = {tags: 'gorm:"foreignKey:ID"'}];   // build info for tests
     repeated Score Scores = 3 [(go.field) = {tags: 'gorm:"foreignKey:ID"'}]; // list of scores for different tests
-}
+}
```
meling, line 34: Please don't remove newlines at end of files.

kit/proto/score.proto
0xf8f8ff: Yeah, this was the most problematic rebase I've ever encountered, still cleaning up the consequences. Everything will be up to date soon.

public/src/components/manual-grading/Review.tsx
```diff
@@ -20,7 +20,7 @@ interface ReviewPageProps {
 interface ReviewPageState {
     review: Review | undefined;
     open: boolean;
-    benchmarks: GradingBenchmark[];
+    // benchmarks: GradingBenchmark[];
```
meling, line 23: Let's remove rather than comment these lines, or is there a reason to keep them?

kit/proto/score.proto
meling: Okay. I have used merge rather than rebase for everything lately... and haven't had any problems. We leave some merge commits, but that's fine.

public/src/components/manual-grading/Review.tsx
0xf8f8ff: Yes, I'll remove all the commented parts once everything is debugged. It is still a draft PR.

public/src/components/student/CoursePanel.tsx
```diff
@@ -30,7 +30,7 @@ export class CoursePanel extends React.Component<IPanelProps> {
                             selector={(item: ISubmissionLink) => {
                                 let score = "N/A";
                                 if (item.submission) {
-                                    score = item.assignment.getSkiptests() ? scoreFromReviews(item.submission.reviews).toString() : item.submission.score.toString();
+                                    score = item.assignment.getReviewers() > 0 ? scoreFromReviews(item.submission.reviews).toString() : item.submission.score.toString();
```
meling, line 33: This code could be converted to a helper function, including the preceding `if` statement. The same code appears also in `student/SingleCourseOverview.tsx`. But note that one case uses `toString()`, while the other doesn't, so this should be made consistent.

public/src/components/teacher/Results.tsx
```diff
@@ -140,7 +140,7 @@ export class Results extends React.Component<IResultsProps, IResultsState> {
             const previousStatus = selectedSubmission.status;
             selectedSubmission.status = status;
             // if the submission is for manual review, update submission score from reviews
-            if (currentSubmissionLink.assignment.getSkiptests()) {
+            if (currentSubmissionLink.assignment.getReviewers() > 0) {
```
meling, line 143: Also, since we seem to use `getReviewers() > 0` to determine whether or not the assignment has tests to be executed, we should instead create a helper function, or extend the assignment type with a method `hasTests()` or `isManual()`. This way we can more easily change these things in one place later, if necessary. See also PR #464.

meling: Alternatively, it could be added as an extension method on the SubmissionLink type, if that's the main use for such a method.

public/src/components/teacher/Results.tsx
```diff
@@ -209,7 +209,7 @@ export class Results extends React.Component<IResultsProps, IResultsState> {
         let selector: (string | JSX.Element | ICellElement)[] = [displayName, groupName];
         selector = selector.concat(student.labs.map(
             (e) => {
-                const setScore = e.assignment.getSkiptests() ? scoreFromReviews(e.submission?.reviews ?? []) : e.submission?.score ?? 0;
+                const setScore = e.assignment.getReviewers() > 0 ? scoreFromReviews(e.submission?.reviews ?? []) : e.submission?.score ?? 0;
```
meling, line 212: Maybe it would make more sense with a helper function on the `e` type here that returns the score.

public/src/components/student/CoursePanel.tsx
meling: See also below. Seems like the helper method could be something like `item.getScore()` or `getScore(item)`, where `item` is the `ISubmissionLink` type. This could then be reused several places.

public/src/pages/views/LabResultView.tsx
```diff
@@ -20,7 +20,7 @@ export class LabResultView extends React.Component<ILabInfoProps> {
         if (this.props.submissionLink.submission) {
             const latest = this.props.submissionLink.submission;
             const buildLog = latest.buildInfo.getBuildlog().split("\n").map((x, i) => <span key={i} >{x}<br /></span>);
-            const score = this.props.submissionLink.assignment.getSkiptests() ? scoreFromReviews(latest.reviews) : latest.score;
+            const score = this.props.submissionLink.assignment.getReviewers() > 0 ? scoreFromReviews(latest.reviews) : latest.score;
```
meling, line 23: Can also reuse a helper function, if `latest` is replaced accordingly.

web/assignments.go
```diff
@@ -140,18 +141,77 @@ func (s *AutograderService) createReview(query *pb.Review) (*pb.Review, error) {
 			submission.ID, assignment.Name, assignment.Reviewers)
 	}
 	query.Edited = time.Now().Format("02 Jan 15:04")
+
```
meling, line 144: Can this `Format` call use the `layout` const rather than a string as in this case?

ag/ag.proto
0xf8f8ff: Benchmarks can probably be obtained via an assignment, the question is how to pass them to the frontend with the corresponding review if a Review type will not have a []GradingBenchmark attribute. 

I guess we can just fetch from the db all the copies of benchmarks for one assignment and all submissions/reviews and add some frontend logic that will select the correct set of benchmarks to display for a single review. But it will result in adding even more complexed frontend logic, is it worth it just to remove a single duplicate field?

public/src/components/student/CoursePanel.tsx
0xf8f8ff: I move all the score calculation to the server side, this way a submission score will always be updated when a review score is updated (maybe we can even drop the `score` field on the review altogether), so I just replace all these occasions with using the Submission score directly.

meling: This is a good choice.

ag/ag.proto
meling: We can keep duplicate fields. But maybe then we shouldn't keep two copies in the database. That is, an alternative is to have a separate message for the database and another message for passing data to the frontend. Agree that we should keep the frontend logic as simple as possible.

0xf8f8ff: Not sure I understand what do you mean by two copies in the database? There is still one `GradingBenchmarks` table in the database as of now. But there will be an initial set of benchmarks created for the assignment, and then a "copy" of the same set for each reviewed submissions. They will of course have different grades and comments for different reviews. 

For example now:
image
the first three belong to the assignment, the next three belong to the review with ID 1. I don't really see the way to avoid having those copies (other then encode them in a single json string as before).

meling: ok

ag/assignments.go
```diff
@@ -47,7 +47,11 @@ func (a *Assignment) CloneWithoutSubmissions() *Assignment {
 		IsGroupLab:        a.IsGroupLab,
 		ScoreLimit:        a.ScoreLimit,
 		Reviewers:         a.Reviewers,
-		SkipTests:         a.SkipTests,
 		GradingBenchmarks: a.GradingBenchmarks,
 	}
 }
+
+// IsForManualGrading returns true if the assignment will be graded manually.
+func (a *Assignment) IsForManualGrading() bool {
```
meling, line 55: Suggest to name it `ManualGrading`.

ag/review.go
```diff
@@ -1,53 +1,31 @@
...
-	}
-	r.Review = strings.Join(str, "; ")
-	return nil
-}
-
-var re = regexp.MustCompile(`("\w+"):"(\d+)",`)
+// CalculateScore returns total score for the review. If grading criteria give
+// a predefined amount of points (not necessary summing up to 100), returns the sum of such points.
+// Otherwise, gives each criterion an equal weight and sets max score to 100
+func (r *Review) CalculateScore() {
```
meling, line 8: - We should have tests for this method.
- Also, suggest renaming it to `ComputeScore`.
- Remove blank line after method signature.
- The method comment appears to return the score rather than assign the value to the `Review` object. What about the comment below.

```golang
// ComputeScore computes the total score for the review and assigns it to r.
// If the grading criteria have predefined points, the score is the sum of these points.
// Otherwise, each criterion is given equal weight, such that the max score is 100.
```

database/gormdb_assignment.go
```diff
@@ -69,7 +68,9 @@ func (db *GormDB) GetAssignmentsByCourse(courseID uint64, withGrading bool) ([]*
 	if withGrading {
 		for _, a := range assignments {
 			var benchmarks []*pb.GradingBenchmark
-			if err := db.conn.Where("assignment_id = ?", a.ID).Find(&benchmarks).Error; err != nil {
+			if err := db.conn.Where("assignment_id = ?", a.ID).
```
meling, line 71: Suggest to move `.Where()` to a separate line to make it line up with the next `Where()` clause.

meling, line 71: Suggest to move `.Where()` to a separate line to make it line up with the next `Where()` clause.

database/gormdb_assignment.go
```diff
@@ -69,7 +68,9 @@ func (db *GormDB) GetAssignmentsByCourse(courseID uint64, withGrading bool) ([]*
 	if withGrading {
```
meling, line 68: Do we have a test for `withGrading == true`?

database/gormdb_assignment.go
```diff
@@ -145,8 +148,8 @@ func (db *GormDB) CreateBenchmark(query *pb.GradingBenchmark) error {
 
 // UpdateBenchmark updates the given benchmark
 func (db *GormDB) UpdateBenchmark(query *pb.GradingBenchmark) error {
-	return db.conn.Model(query).
-		Where(&pb.GradingBenchmark{ID: query.ID, AssignmentID: query.AssignmentID}).
+	return db.conn.
+		Where(&pb.GradingBenchmark{ID: query.ID, AssignmentID: query.AssignmentID, ReviewID: query.ReviewID}).
```
meling, line 152: Here we should probably move the different fields to new lines.

meling, line 152: Here we should probably move the different fields to separate lines.

public/src/componentHelper.tsx
```diff
@@ -371,11 +357,15 @@ export function getDaysAfterDeadline(deadline: Date, delivered: Date): number {
...
     return i >= 0 && i <= len - 1;
 }
+
+export function forManualReview(a: Assignment): boolean {
+    return a && a.getReviewers() > 0;
+}
+
+export function setScoreString(s: ISubmission | undefined) {
+    return s?.score.toString() ?? "0";
+}
```
meling, line 371: Missing newline at end of the file.

public/src/componentHelper.tsx
```diff
@@ -371,11 +357,15 @@ export function getDaysAfterDeadline(deadline: Date, delivered: Date): number {
...
-export function isValidUserName(username: string): boolean {
-    const onlyCharsAndSpaces = /^[a-zA-Z\s]*$/;
-    return onlyCharsAndSpaces.test(username);
-}
 
 export function legalIndex(i: number, len: number): boolean {
     return i >= 0 && i <= len - 1;
 }
+
+export function forManualReview(a: Assignment): boolean {
```
meling, line 365: Use the same name as on the server-side, e.g., `GradedManually`.

ag/assignments.go
0xf8f8ff: Renamed to `GradedManually`.

database/gormdb_assignment.go
0xf8f8ff: Moved.

database/gormdb_assignment.go
0xf8f8ff: Looks like we don't. Will add one.

database/gormdb_assignment.go
0xf8f8ff: Done.

public/src/componentHelper.tsx
0xf8f8ff: Fixed.

public/src/componentHelper.tsx
0xf8f8ff: Done.

database/gormdb_assignment.go
0xf8f8ff: Added.

database/gormdb_assignment.go
0xf8f8ff: Done.

ag/review.go
0xf8f8ff: Renamed, test added.

database/gormdb_assignment.go
0xf8f8ff: Done.

database/gormdb_assignment.go
0xf8f8ff: And done.

database/gormdb_assignment_test.go
```diff
@@ -204,7 +204,7 @@ func TestGetAssignmentsWithSubmissions(t *testing.T) {
 		Score:        45,
 		Reviews: []*pb.Review{
 			{
-				ReviewerID: 1, Review: "LGTM!", Feedback: "SGTM!", Score: 42, Ready: true,
+				ReviewerID: 1, Feedback: "SGTM!", Score: 42, Ready: true,
 				// TODO(meling) Remove Benchmarks, since benchmarks should only be kept in one place and associated with the assignment (see also TODO in ag.proto)
```
meling, line 208: This TODO is no longer valid and can be removed.

0xf8f8ff: Done.

meling: Will the test now work with the `Benchmarks` field added again?

0xf8f8ff: All tests pass.

database/gormdb_assignment_test.go
```diff
@@ -204,16 +204,7 @@ func TestGetAssignmentsWithSubmissions(t *testing.T) {
...
-				// TODO(meling) Remove Benchmarks, since benchmarks should only be kept in one place and associated with the assignment (see also TODO in ag.proto)
-				// Benchmarks: []*pb.GradingBenchmark{
-				// 	{
-				// 		Heading: "Ding Dong", Comment: "Communication",
-				// 		Criteria: []*pb.GradingCriterion{
-				// 			{Points: 50, Description: "Loads of ding"},
-				// 		},
-				// 	},
-				// },
+				ReviewerID: 1, Feedback: "SGTM!", Score: 42, Ready: true,
```
meling, line 207: Should the `Benchmarks` field be re-added now? And I think we should remove the TODO below.

public/src/components/manual-grading/Release.tsx
```diff
@@ -25,7 +25,7 @@ interface ReleaseState {
     open: boolean;
     reviews: Review[];
     reviewers: Map<User, Review>;
-    score: number;
+    // score: number;
```
meling, line 28: Can these commented lines be removed?

database/gormdb_assignment_test.go
0xf8f8ff: Oh, right. Yes, adding it back. Test still passes with it.

public/src/pages/views/ReleaseView.tsx
```diff
@@ -168,7 +168,8 @@ export class ReleaseView extends React.Component<ReleaseViewProps, ReleaseViewSt
                                 const current = this.state.submissionsForGroupAssignment.get(grp)?.submission;
                                 if (current) {
                                     current.status = status;
-                                    current.score = scoreFromReviews(current.reviews);
+                                    // TODO(vera): make sure these changes don't break things on Release pages
```
meling, line 171: Can this be removed?

database/gormdb_assignment_test.go
0xf8f8ff: Done.

public/src/components/manual-grading/Release.tsx
0xf8f8ff: Fixed.

public/src/pages/views/ReleaseView.tsx
0xf8f8ff: Fixed too.

web/assignments.go
```diff
@@ -139,19 +142,57 @@ func (s *AutograderService) createReview(query *pb.Review) (*pb.Review, error) {
 		return nil, fmt.Errorf("Failed to create a new review for submission %d to assignment %s: all %d reviews already created",
 			submission.ID, assignment.Name, assignment.Reviewers)
 	}
-	query.Edited = time.Now().Format("02 Jan 15:04")
+	query.Edited = time.Now().Format(reviewLayout)
```
meling, line 145: Suggest renaming `query` to `review`, since it now holds review data as well. Same applies to next method as well.

0xf8f8ff: Done.

BjorniTe1 (approved):
Looks good to me.

Issue #446, opened 2021-06-14
Make deploy-selfcert.md


meling:
We now have a `doc/deploy.md` in PR #444. We should add a version of this file that use `127.0.0.1` and self-signed certificates for testing on localhost only.

meling:
Already fixed

PR #467, opened 2021-07-16
Testing Github webhooks on localhost

0xf8f8ff:
- updated scm tool to allow adding webhooks on the organization level (was only possible for a single repository before)
- updated docs


doc/local-setup/local-setup.md
```diff
@@ -359,3 +359,23 @@ Example:
 ```
 
 Note: For a local dev server using a port is no issue, and should work with no additional steps needed than described above. To access the gRPC server externally through gRPC-web, however, the port you choose in the Envoy config file (and set gRPC to in GRPCManager.tsx) has to be opened.
+
+
+### Github Webhooks when running on localhost with self-certs
```
meling, line 364: Not sure if this text belongs in this file. Ideally, I'm looking to remove the `local-setup` folder, when we have a suitable replacement file, e.g., `deploy-local.md` or something.

0xf8f8ff: Where should i put it then? Is a temporary separate .md file ok while we don't have a replacement file?

meling: Probably, we can add a temp .md file for now, and move it into another file, when #446 is done. I've assigned #446 to @JosteinLindhom for now, as I think he already has some material or ideas in this regard... If not @JosteinLindhom let me know and I will reassign.

0xf8f8ff: Ready.

PR #479, opened 2021-07-29
Internal package for test helpers; includes TestDB()

### LINKED ISSUE #478: Add internal package for test helpers and other non-exported stuff
We need an internal package for various test helpers. In particular, we have three versions of `setup(t)` for creating a test database.

meling:
- Added internal package with TestDB test helper
- Updated database pkg to use internal.TestDB() helper
- Service accept Database interface instead of *GormDB
- Replace util_test.go:setup() with internal.TestDB() helper
- Removed another instance of setup() test helper in web/auth

Fixes #478.

### DIFF
--- a/internal/test_helper.go
+++ b/internal/test_helper.go
@@ -1,0 +1,37 @@
+package internal
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/autograde/quickfeed/database"
+	"github.com/autograde/quickfeed/log"
+)
+
+// TestDB returns a test database and close function.
+// This function should only be as a test helper.
+func TestDB(t *testing.T) (database.Database, func()) {
+	t.Helper()
+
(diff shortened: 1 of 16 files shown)


internal/test_helper.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/autograde/quickfeed/database"
+	"github.com/autograde/quickfeed/log"
+)
+
+// TestDB returns a test database and close function.
+// This function should only be as a test helper.
+func TestDB(t *testing.T) (database.Database, func()) {
+	t.Helper()
+
+	f, err := ioutil.TempFile(t.TempDir(), "test.db")
```
0xf8f8ff, line 17: This line returns `t.TempDir undefined (type *testing.T has no field or method TempDir)` for me.

meling: Ah, right. We need to upgrade to a newer Go version.

PR #483, opened 2021-07-30
GetSubmissionsByCourse cleanup

### LINKED ISSUE #462: GetSubmissionsByCourse should only return the essentials
We currently do the following in `getAllCourseSubmissions`:
```golang
	var getCourseSubFn func(uint64, pb.SubmissionsForCourseRequest_Type) ([]*pb.Assignment, error)
	if request.GetSkipBuildInfo() {
		getCourseSubFn = s.db.GetCourseAssignmentsWithSubmissionsNoBuildInfo
	} else {
		getCourseSubFn = s.db.GetCourseAssignmentsWithSubmissions
	}
	assignments, err := getCourseSubFn(request.GetCourseID(), request.Type)
	if err != nil {
		return nil, err
	}
```
And:
```golang
	if request.GetSkipBuildInfo() {
		sbm.BuildInfo = ""
		sbm.ScoreObjects = ""
	} else {
		err = sbm.MakeSubmissionReviews()
		if err != nil {
			return nil, err
		}
	}
```
Is the `GetSubmissionsByCourse` method also used in the frontend to fetch individual assignments/submission/score/build info??

Ideally, we should only use this method to get the data for the results view, without build info. Then fetch build info and scores when clicking individual links, or fetch them in the background, once the results view has been rendered.

oskargjolga:
fixes #462


go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/docker/docker v20.10.7+incompatible
 	github.com/docker/go-connections v0.4.0 // indirect
 	github.com/fatih/color v1.9.0 // indirect
+	github.com/golang/protobuf v1.5.2 // indirect
```
meling, line 14: This is the old protobuf library; we should not depend on this. Check your code for imports using `github.com/golang/protobuf` and replace it with `google.golang.org/protobuf`.

web/courses.go
```diff
@@ -163,7 +163,7 @@ func (s *AutograderService) getAllCourseSubmissions(request *pb.SubmissionsForCo
 	course.SetSlipDays()
 	for _, a := range assignments {
 		for _, sbm := range a.Submissions {
-			if request.GetSkipBuildInfo() {
+			if !request.GetWithBuildInfo() {
```
meling, line 166: Suggest passing this boolean to the database function above. That way we avoid fetching build info from the database when we don't need it. See my suggestions in issue #462 that uses the `withBuildInfo` bool; this can be passed as an argument to `GetAssignmentsWithSubmissions`... I don't think there are many uses of this method, so should be a simple change.

meling:
@oskargjolga Let me know if you need some feedback on the remaining parts of this. It is not marked ready for review and there are a few things remaining it looks like...

oskargjolga:
@meling Busy week... Will do a couple commits and mark it ready for review tomorrow!

go.mod
oskargjolga: My outdated protoc-gen-go imported this in ag.pb.go... Fixed it now.

meling (approved):

PR #486, opened 2021-08-03
Replace string-based build date with timestamppb

### LINKED ISSUE #432: BuildDate is string in score.proto
Consider replacing `BuildDate` with a proper date type that is easy to use across languages.
This would avoid `time.Parse` in `ag.NewestBuildDate` and `ci.updateSlipDays`.

meling:
This replaces the string-based BuildDate field in BuildInfo with google.protobuf.Timestamp.

Fixes #432.


meling:
@0xf8f8ff Can you take a look at this? It isn't quite ready to merge yet, but I would like some input. You should run `make proto` or `make hack-timestamp` locally. Tests are failing on GitHub because the workflows do not run this make target yet... I will try to fix this later if this approach is working.

0xf8f8ff:
> @0xf8f8ff Can you take a look at this? It isn't quite ready to merge yet, but I would like some input. You should run `make proto` or `make hack-timestamp` locally. Tests are failing on GitHub because the workflows do not run this make target yet... I will try to fix this later if this approach is working.

Tested, runs just fine on a fresh database. The test database will be incompatible though, probably worth mentioning in doc (until it is updated).

meling:
Superseded by #947.

PR #491, opened 2021-08-06
Quickfeed dev docker setup

r0qs:
This PR attempt to provide a quick and easy to use a docker development setup that may also be used for production.


.env
```diff
@@ -0,0 +1,11 @@
+# Make a copy of this file and update the variables.
```
meling, line 1: Should this file be called `.env-template` or something, and `.env` be added to `.gitignore` to prevent that someone accidentally pushes changes to GitHub?

envoy/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+ADD ./envoy/envoy.yaml /etc/envoy/envoy.yaml
+
+ARG DOMAIN
+COPY ./envoy/certs /etc/letsencrypt/live/${DOMAIN}
+
+RUN chmod 644 /etc/letsencrypt/live/${DOMAIN}/fullchain.pem
+RUN chmod 644 /etc/letsencrypt/live/${DOMAIN}/key.pem
+
+CMD ["envoy", "-c", "/etc/envoy/envoy.yaml"]
```
meling, line 11: Remember the newline at end of the file.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+        sudo \
+        make \
+        protobuf \
+        protobuf-dev \
+        npm
+ADD . /quickfeed
+WORKDIR /quickfeed
+RUN make devtools && make ui && make proto && make install
+
+ENTRYPOINT ["quickfeed"]
```
meling, line 17: Remember the newline at end of the file.

.dockerignore
```diff
@@ -0,0 +1,3 @@
+.git
+.github
+.vscode
```
meling, line 3: Remember the newline at end of the file.

envoy/envoy.go
```diff
@@ -46,7 +410,7 @@ func StartEnvoy(l *zap.Logger) {
 	if !hasEnvoyImage(ctx, l, cli) {
 		// if there is no active Envoy image, we build it
 		l.Info("building Envoy image...")
-		// TODO(meling) use docker api to build image: "docker build -t ag_envoy -f ./envoy/envoy.Dockerfile ."
+		// TODO(meling) use docker api to build image: "docker build -t ag_envoy -f ./envoy/Dockerfile ."
```
meling, line 413: The `StartEnvoy` and other unused functions can be removed.

envoy/envoy.tmpl
```diff
@@ -112,4 +183,5 @@ static_resources:
         - lb_endpoints:
           - endpoint:
               address:
-                socket_address: { address: localhost, port_value: $HTTP_PORT }
+                socket_address: { address: localhost, port_value: {{ .HTTPPort }} }
+{{ end -}}
```
meling, line 187: GitHub and I prefer to have newlines also at the end of files.

.env
r0qs: Good point. I'll add it to `.gitignore` and rename the example.
I think they can be defined in the same file. I was thinking to have only one configuration file for the whole project, which could be set using environmental variables, through the `.env` file or explicitly by the user, or through a configuration in `json`, `toml` or `yaml` format.

envoy/envoy.tmpl
```diff
@@ -112,4 +121,5 @@ static_resources:
         - lb_endpoints:
           - endpoint:
               address:
-                socket_address: { address: localhost, port_value: $HTTP_PORT }
+                socket_address: { address: {{ .ServerHost }}, port_value: {{ .HTTPPort }} }
+{{ end }}
```
meling, line 125: Add newline

public/src/index.tsx
```diff
@@ -314,7 +313,9 @@ async function main(): Promise<void> {
 
     if (curRunning === DEBUG_SERVER) {
         const httpHelper = new HttpHelper("/api/v1");
-        const grpcHelper = new GrpcManager();
+        // TODO: use process.env.QUICKFEED_PORT
+        // https://webpack.js.org/plugins/environment-plugin/
+        const grpcHelper = new GrpcManager("https://" + window.location.hostname);
```
meling, line 318: Can this TODO be resolved easily, or will this require much work?? Also, check out what @nicro950 did in #493 regarding this stuff, to ensure things are compatible. Note that, I think `"https://"` can be replaced with `window.location.protocol`...

PS: @JosteinLindhom managed to remove these arguments to `GrpcManager`, but I'm clueless about this stuff...

envoy/envoy.go
```diff
@@ -1,81 +1,378 @@
...
 }
 
-// hasEnvoyImage returns true if the docker client has the latest Envoy image.
-func hasEnvoyImage(ctx context.Context, l *zap.Logger, cli *client.Client) bool {
-	l.Debug("no running Envoy container found")
-	images, err := cli.ImageList(ctx, types.ImageListOptions{})
+// TODO: improve parameter handling when generating certificates (keyType, etc).
+// TODO: save certs at: /etc/ssl/certs and private keys at: /etc/ssl/private by default.
+func main() {
+	flag.BoolVar(&genConfig, "genconfig", false, "generate envoy config")
```
meling, line 353: We should consider if this should be a standalone tool (as is now), or if this should be moved to another command line tool inside `cmd/` folder. Right now, we have too many standalone tools.

Another option is to make this part of the `quickfeed` command itself, but running it something like this:

```sh
% quickfeed gen-envoy-config --tls --cert cert.pem --key private.pem
```

Which would also entail that:

```sh
% quickfeed run -service.url $DOMAIN
```

At any rate, even a standalone tool should avoid the `--genconfig` flag, and instead have it be a command like `genconfig`.

But this comment is mainly added for context and brainstorming... Maybe we should create a separate issue for this.

envoy/envoy.go
```diff
@@ -1,81 +1,378 @@
...
-	ctx := context.Background()
-	cli, err := client.NewClientWithOpts(client.FromEnv)
+type CertificateConfig struct {
+	CertFile string
+	KeyFile  string
+}
+
+type EnvoyConfig struct {
+	Domain     string
+	ServerHost string
```
meling, line 37: It could be useful to have doc comment for `ServerHost`

envoy/envoy.go
```diff
@@ -1,81 +1,378 @@
...
+		TLSEnabled: withTLS,
+	}
+
+	if withTLS {
+		if certConfig.CertFile != "" && certConfig.KeyFile != "" {
+			config.CertConfig = certConfig
+			return config, nil
+		}
+		err := generateSelfSignedCert(certOptions{
+			hosts: fmt.Sprintf("%s,%s", "localhost", domain),
```
meling, line 59: Should this be `localhost` as string or from the `config.ServerHost` variable?

meling (changes requested):

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+        protobuf-dev \
+        npm
+ADD . /quickfeed
+WORKDIR /quickfeed
+RUN make devtools
+RUN make proto
+RUN make ui
+RUN make install
+
+EXPOSE 8080 9091
```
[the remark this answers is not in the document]
r0qs: The `EXPOSE` instruction actually works more like documentation of what ports are listening on the container than actually publish the port. To make them accessible in the host machine, you need to specify the `ports` section in the docker-compose or run `docker run` with the `-p` option.

.env-template
```diff
@@ -0,0 +1,16 @@
...
+GITHUB_KEY="KEY"
+GITHUB_SECRET="SECRET"
+
+# Envoy Config
+# Quickfeed server domain or ip
+DOMAIN=localhost
+# Quickfeed container ip or container_name or hostname
+SERVER_HOST=quickfeed
+# Quickfeed grpc port
+GRPC_PORT=9090
```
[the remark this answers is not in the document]
r0qs: Ops, indeed the ports are wrong. Maybe it would be better to have the ports as environment variables in the Dockerfile. I will fix that.

Makefile
```diff
@@ -78,20 +81,28 @@ brew:
...
+	@echo "You must set required environment variables before configuring Envoy (see doc/scripts/envs.sh)." && false
+else
+	@echo "Generating Envoy configuration for '$$DOMAIN'."
+	@go run ./envoy/envoy.go --genconfig
+endif
+
 envoy-build:
 	@echo "Building Autograder Envoy proxy"
-	@cd envoy; docker build -t ag_envoy -f envoy.Dockerfile .
+	@docker-compose build --force-rm envoy
```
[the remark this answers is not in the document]
r0qs: I don't think that they are really necessary. I kept it because I didn't know if someone was using it.

doc/deploy.md
```diff
@@ -81,57 +179,23 @@ If you like Certbot, please consider supporting our work by:
...
-% sudo systemctl disable nginx
-```
+Please choose one of the options below to run envoy in your system:
 
-### Running with Locally Installed Envoy (macOS homebrew)
+1. Running with Locally Installed Envoy (macOS homebrew)
 
 ```sh
-% sudo envoy -c $ENVOY_CONFIG &
+% sudo envoy -c $QUICKFEED/envoy/envoy.yaml &
```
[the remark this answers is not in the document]
r0qs: I removed it because when building the docker image, the `envoy.yaml` config must be inside the context of the Dockerfile, otherwise commands like ADD or COPY do not work.
In the case of the envoy, the context is the folder `$QUICKFEED/envoy`. So, if you specify a config in a folder outside that one, the docker builder will not be able to find it. I could not find an easy way to switch between the two approaches other than requiring users to create the `envoy.yaml` inside `$QUICKFEED/envoy`.

One possible alternative would be to have a script that copies the config to the right place before building the image.

envoy/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM envoyproxy/envoy-alpine:v1.19-latest
+
+ADD ./envoy/envoy.yaml /etc/envoy/envoy.yaml
```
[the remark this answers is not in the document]
r0qs: Yes, this may be possible, but please take a look at my other comment about the context when building the image. If `ENVOY_CONFIG` is outside the specified context in the docker-compose: `services.envoy.build.context` the builder will not be able to find the configuration.

Maybe there is a way to do it, but I couldn't find a solution for that yet.

doc/deploy.md
```diff
@@ -227,17 +296,17 @@ quickfeed -service.url uis.itest.run &> quickfeed.log &
...
-### Flags
+#### Flags
 
 | **Flag**        | **Description**                        | **Example**     |
 |-----------------|----------------------------------------|-----------------|
 | `service.url`   | Base DNS name for QuickFeed deployment | `uis.itest.run` |
 | `database.file` | Path to QuickFeed database             | `qf.db`         |
 | `grpc.addr`     | Listener address for gRPC service      | `:9090`         |
-| `http.addr`     | Listener address for HTTP service      | `:3005`         |
+| `http.addr`     | Listener address for HTTP service      | `:8081`         |
```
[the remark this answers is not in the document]
r0qs: Yes, this is also my understanding. I guess we can explain that in the beginning in the `Preparing the Environment` section, saying which ports will be exposed by default when deploying.

docker-compose.yml
```diff
@@ -1,27 +1,40 @@
...
+      - ./envoy/envoy.yaml:/etc/envoy/envoy.yaml
 
   quickfeed:
-    image: quickfeed
+    container_name: quickfeed
+    image: quickfeed/quickfeed
     build:
       context: ./
-      dockerfile: ./dev/quickfeed/Dockerfile
+      dockerfile: ./Dockerfile
```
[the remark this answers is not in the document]
r0qs: Yes, we can move it around. As there is no quickfeed folder, I thought it was reasonable to keep it in the root directory.
Creating a sub-folder for `packaging` or `build` would alleviate the directory listing you mentioned? Or we can just move it to `ci`.

public/src/index.tsx
r0qs: Yes, I let it as TODO because I was think to make a separate PR to also adjust this `DEBUG_SERVER` and `DEBUG_BROWSER` flags.

envoy/envoy.go
r0qs: I moved it to `cmd` and change the flags. But I let it as standalone tool for now, but we can change this in another PR later.

meling (approved):

PR #493, opened 2021-08-09
[WIP] Dev containers and local without public domain and https

nicro950:
This pr adds support for developing quickfeed in a devcontainer as well as changes required to start quickfeed without https and a public domain


meling:
I will review this later but note that we should make sure it is compatible with #491. Also, I thought I mention that we should try to consolidate the `Build` and `Rebuild` RPCs.

meling:
I still have not reviewed this in-depth, but please see also #538. I would split the content of this PR into two:

- Code relevant for building (gRPC API) -- if this PR has a particular need, we should aim to solve it with the same single RPC call, refer to #538. The code for `buildSubmission` appears to have sufficient overlap to at least share a common entry point rather than have two or three RPCs.
- Code and configuration that are relevant for dev containers etc.

meling:
Won't merge. Unable to get the setup stuff to work (long ago), and this has become outdated.

PR #515, opened 2021-08-22
Fix for some session-related issues

0xf8f8ff:
- generate random secret key for a new cookie store
- moved random string generating method to `internal` as it is useful in several places
- don't return an http error when an outdated session found

### DIFF
--- a/internal/qutil/util.go
+++ b/internal/qutil/util.go
@@ -1,0 +1,15 @@
+package qutil
+
+import (
+	"crypto/rand"
+	"crypto/sha1"
+	"fmt"
+)
+
+func RandomString() string {
+	randomness := make([]byte, 10)
+	if _, err := rand.Read(randomness); err != nil {
+		panic("couldn't generate randomness")
+	}
+	return fmt.Sprintf("%x", sha1.Sum(randomness))
+}
--- a/web/auth/auth.go
+++ b/web/auth/auth.go
@@ -61,7 +61,7 @@ var cookieStore = make(map[string]uint64)
 // Add adds cookie for userID, replacing userID's current cookie, if any.
 func Add(cookie string, userID uint64) {
 	for currentCookie, id := range cookieStore {
-		if id == userID && currentCookie != cookie {
+		if id != userID || (id == userID && currentCookie != cookie) {
 			delete(cookieStore, currentCookie)
 		}
 	}
--- a/web/auth/auth.go
+++ b/web/auth/auth.go
@@ -61,7 +61,7 @@ var cookieStore = make(map[string]uint64)
 // Add adds cookie for userID, replacing userID's current cookie, if any.
 func Add(cookie string, userID uint64) {
 	for currentCookie, id := range cookieStore {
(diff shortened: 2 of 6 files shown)


internal/qutil/util.go
```diff
@@ -0,0 +1,15 @@
+package qutil
```
meling, line 1: Suggest to name this package `rand` or `random`. Then we can use `random.String()` instead of `qutil.RandomString()`

internal/qutil/util.go
```diff
@@ -0,0 +1,15 @@
...
+	"crypto/sha1"
+	"fmt"
+)
+
+func RandomString() string {
+	randomness := make([]byte, 10)
+	if _, err := rand.Read(randomness); err != nil {
+		panic("couldn't generate randomness")
+	}
+	return fmt.Sprintf("%x", sha1.Sum(randomness))
```
meling, line 14: Maybe we should use a strong hash function such as `sha256`.

web/auth/auth.go
```diff
@@ -61,7 +61,7 @@ var cookieStore = make(map[string]uint64)
 // Add adds cookie for userID, replacing userID's current cookie, if any.
 func Add(cookie string, userID uint64) {
 	for currentCookie, id := range cookieStore {
-		if id == userID && currentCookie != cookie {
+		if id != userID || (id == userID && currentCookie != cookie) {
```
meling, line 64: Please add a comment to explain this. Also, we should make sure all cases are covered by the tests.

internal/qutil/util.go
0xf8f8ff: Renamed to rand.

internal/qutil/util.go
0xf8f8ff: Fixed.

meling:
> This change will delete every cookie except id == userID, essentially logging everyone else out.
> 
> I think the change with random secret for cookieStore in [8cd9e49](https://github.com/autograde/quickfeed/commit/8cd9e490bc7b4e471468038f5cf1c5dfb7cf18b6) should be enough to enforce only valid sessions cookies are allowed.

I agree that this is weird, but when I reverted this it won't let me stay logged in... Although I think this worked last night. But not sure what's going on here... I need to enable more logging.

r0qs:
> This change will delete every cookie except id == userID, essentially logging everyone else out.
> 
> I think the change with random secret for cookieStore in [8cd9e49](https://github.com/autograde/quickfeed/commit/8cd9e490bc7b4e471468038f5cf1c5dfb7cf18b6) should be enough to enforce only valid sessions cookies are allowed.

It would be interesting to have a alternative for non-broswer authentications. For instance the helpbot lost the ability to use the grpc endpoint since it does not has valid session cookies. So, we could also allow the use of application authentication tokens.

meling:
Yes, we should definitely add support for application tokens. This could also help to resolve the problem with the Swift UI client. @oskargjolga and @BjorniTe1 ... Any takers for this functionality??

PR #522, opened 2021-08-23
Fix confusing log messages during user/repo creation

### LINKED ISSUE #511: Fix confusing log statements
These types of log messages cause confusion:

```text
scm/github.go:122    CreateRepository got expected error ...
scm/github.go:353    Team allteachers not found as expected ...
createStudentRepo: repo not found (as expected) ...
```

Rephrase these to be more readable.

meling:
- Improved logging for CreateRepository
- Improved logging for CreateTeam
- Removed useless GetRepo check and logging
- Removed allocation and added context to error msg

Fixes #511

### DIFF
--- a/scm/github.go
+++ b/scm/github.go
@@ -351,13 +351,15 @@ func (s *GithubSCM) CreateTeam(ctx context.Context, opt *NewTeamOptions) (*Team,
 		}
 	}
 
-	// first check whether the team with this name already exists on this organization
+	// check that the team name does not already exist for this organization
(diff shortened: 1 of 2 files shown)


meling:
Ideally, we should add more tests for github interaction, but maybe something to do as part of a larger clean up project for a bachelor group. For now, unless there are other issues, let's merge this one. Note the removal of a GetRepo interaction that I don't think is necessary; at least as long as CreateRepo does the GetRepo first.

PR #525, opened 2021-08-28
Fix for the gRPC auth problem with helpbot

### LINKED ISSUE #519: Add support for application authentication tokens
As described in [this comment](https://github.com/autograde/quickfeed/pull/515#issuecomment-903731223) to PR #515, it would be useful to have an alternative non-browser authentication mechanism, perhaps using application authentication tokens. It should be possible to use such tokens as replacements for password-based logins via OAuth2 with GitHub.

This will be useful for the [helpbot](https://github.com/quickfeed/helpbot) and the [SwiftUI client](https://github.com/autograde/quickfeed-swiftui).

Both these clients should allow configuring an authentication token for the user.

meling:
This adds support for custom auth for gRPC connections.

- Added TestGetUserByCourse
- Added Logger test helper to print log output
- Replace TestDiscordClient with TestGrpcAuth

Fixes #519.


meling:
The corresponding fix in the helpbot is [here](https://github.com/quickfeed/helpbot/pull/10).

I'm aware this is not an ideal solution, but hopefully it can work until we can get a more permanent fix. @r0qs has some ideas that we should discuss.

One issue with this implementation may be that if I log into the user with id 1, the cookie database will be cleared and we lose access from helpbot. Last year, I never logged into this user after it was used to set things up, so hopefully this will not be a problem...

0xf8f8ff (approved):

PR #533, opened 2021-09-02
Fix same submission being queued multiple times for rebuild

### LINKED ISSUE #532: Rebuilding all submissions
- When rebuilding all submissions for an assignment, tests are running for multiple versions of the same submission.
- The number of containers started at the same time by the `RebuildSubmissions` method must be limited to some reasonable number (5 or 10?).

0xf8f8ff:
Resolves #532.

### DIFF
--- a/web/rebuild.go
+++ b/web/rebuild.go
@@ -1,15 +1,15 @@
 package web
 
 import (
+	"sync"
 	"time"
 
 	pb "github.com/autograde/quickfeed/ag"
 	"github.com/autograde/quickfeed/ci"
 	"github.com/gosimple/slug"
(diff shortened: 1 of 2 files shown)


meling:
I've been thinking a bit about the rate-limiting issue. We should have some counter in `docker.go` that keeps track of all running containers, not only those due to the rebuilding. On the other hand, maybe a better approach would be to monitor the CPU load and keep the load below some limit, but I haven't looked into how we might do this.

meling:
It is not clear how many containers we would want to run in parallel, but I left `maxContainers = 10` for now. Merging.

PR #535, opened 2021-09-05
Fixed multiple checked answers sometimes counting as correct

JosteinLindhom:
This PR fixes an issue where multiple checked answers would count as correct if the first checked answer is the correct answer.

Examples that gave points:
```
Question 1:
- [x] a) // Correct answer
- [x] b)
- [x] c)
- [x] d)

Question 2:
- [ ] a) 
- [ ] b)
- [x] c) // Correct answer
- [x] d)
```

### DIFF
--- a/kit/exercise/multiple_choice.go
+++ b/kit/exercise/multiple_choice.go
@@ -19,6 +19,7 @@ var (
 )
 
 // ParseMarkdownAnswers returns a map of the answers found in the given answer file.
+// Only a single answer is allowed per question
 func ParseMarkdownAnswers(answerFile string) (map[int]string, error) {
 	md, err := ioutil.ReadFile(answerFile)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


kit/exercise/multiple_choice.go
```diff
@@ -28,14 +28,19 @@ func ParseMarkdownAnswers(answerFile string) (map[int]string, error) {
...
 	for _, line := range strings.Split(string(md), "\n") {
 		if qNumRegExp.MatchString(line) {
 			qNum := qNumRegExp.ReplaceAllString(line, "$1")
 			// ignore error since regular expression ensure it is already a number
 			currentQ, _ = strconv.Atoi(qNum)
 		}
-		_, found := answerMap[currentQ]
-		if !found && currentQ != -1 && selectionRegExp.MatchString(line) {
+		if !multipleAnswers[currentQ] && currentQ != -1 && selectionRegExp.MatchString(line) {
+			if _, found := answerMap[currentQ]; found {
```
meling, line 39: Add comment here: `// if found is true, multiple answers are given for the same question`

Maybe also point out in a second doc line to the `ParseMarkdownAnswers` function:

```text
Only a single answer per question is allowed.
```

meling (changes requested):

meling:
Good catch! I knew about this problem, but didn't fix it when I rewrote this a while back... Laziness... Anyway, thanks! Will merge when you add the comments.

PR #554, opened 2021-10-26
Support scoring Benchmark functions

### LINKED ISSUE #553: Add support for running and scoring benchmarks
Currently, we can hack it using code like this:
```go
	s, err := sh.Output("go test -v -run none -bench BenchmarkAGWordCount")
	if err != nil {
		sc.Fail()
		t.Fatal(err)
	}
```

However, we should add support for passing `testing.TB` to:
- `score.Print()`
- `score.registry.Add()`
- `score.registry.AddSub()`
- `score.PanicHandler()`

meling:
- Allow testing.TB instead of testing.T in Score API
- Allow also testing.B in Add and AddSub methods

Fixes #553.

### DIFF
--- a/kit/score/registry.go
+++ b/kit/score/registry.go
@@ -107,13 +107,19 @@ func testName(testFn interface{}) string {
 	}
 	name := runtime.FuncForPC(reflect.ValueOf(testFn).Pointer()).Name()
 	name = lastElem(name)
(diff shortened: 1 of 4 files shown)


meling:
Closing this because it isn't possible to do it like this. We have to shell out and parse the output for benchmarks. See issue description.

Issue #570, opened 2022-01-15
GORM Update from struct with zero value fields may not work correctly


meling:
As was discovered today, and also previously, GORM does not update fields with zero value, e.g., bool = false. There are two workarounds for this problem, as mentioned in [this comment](https://github.com/quickfeed/quickfeed/issues/569#issuecomment-1013729572).

We should examine all uses of GORM `Update(s)` and check that updates are correctly stored in the database.

Also we should use the [`Select`](https://gorm.io/docs/update.html#Update-Selected-Fields) approach to ensure that zero value fields of the struct are updated. Using `Select` is preferred because it will allow us to rename the struct fields cleanly, which would not be the case with the `map[string]interface{}` approach, where each field name of the struct would be a string that a type-safe rename would not find.

0xf8f8ff:
So after two days of struggling with gorm there are two major conclusions:

1. Using `Select("*").Updates(...)` is dangerous. For example, 

```
Model(&qf.Enrollment{}).Where(~conditions~).Select("*").Updates(Updates(&qf.Enrollment{State: enrol.State, Status: enrol.Status, LastActivityDate: enrol.LastActivityDate}))
```

will set all the fields not mentioned in `Updates` (for example, `UserID` and `CourseID`) to default values. Currently there is no way to ensure we are always sending query structs where all the fields are set to the relevant values (most often then not fields not relevant to the method are left empty).

The alternative would be using `Select` with field names like `Select("State", "Status", "LastActivityDate")`, but it defeats the purpose as there is still dependency on strings representing field names which we would like to avoid.

Another possibility is fetching the original record inside the database method before updating. Then we could update only the changed fields and use this struct as a query in `Updates` if we don't mind an extra database lookup.

2. Database method `UpdateEnrollments` is used from many places and works correctly only because of this gorm's "feature" of not updating fields to their defaul values with a simple `Updates`.

For example, when updating activity date with `UpdateEnrollment(query)`, the query has empty fields for enrollment `Status` and `State`. Without this gorm's "feature" both these fields would be reset to default values. I don't know how safe it is because if gorm suddenly changes this behaviour the impact on the database will be huge.

However, rewriting `UpdateEnrollment` with `Select` requires changes to all the methods that call it (in most cases an additional call to the database to fetch the original enrollments to ensure ther eare no fiels unintentionally set to defaul values).

meling:
[...]

Issue #569, opened 2022-01-15
Review.Ready bool field fails to reset


meling:
When trying to cycle the `Ready` field, the returned value will remain true, even if set to false.

Running the code in `TestCreateUpdateReview` shows the problem:
```go
	review.Score = 90
	review.Ready = true
	updateSubmission(t, db, review)

	review.Score = 50
	review.Ready = false
	updateSubmission(t, db, review)
```
After setting `Ready = true` and then setting it back to `false` it will remain `true`, even though the `Score = 50` was updated correctly.
```text
go test -v -run TestCreateUpdateReview
=== RUN   TestCreateUpdateReview
sub 2: false, score: 95
sub 2: true, score: 90
sub 2: true, score: 50
```
This is a known GORM problem; [this](https://stackoverflow.com/questions/56653423/gorm-doesnt-update-boolean-field-to-false) seems to provide a solution.

meling:
For future reference. By default, GORM will not update zero value fields. See GORM documentation [here](https://gorm.io/docs/update.html#Updates-multiple-columns) and [specifically the Select approach here](https://gorm.io/docs/update.html#Update-Selected-Fields).

We should use the `Select` approach to ensure that zero value fields of the struct are updated. Using `Select` is preferred because it will allow us to rename the struct fields cleanly, which would not be the case with the `map[string]interface{}` approach, where each field name of the struct would be a string that a type-safe rename would not find.

PR #502, opened 2021-08-16
Redesigned frontend

JosteinLindhom:
This pull request implements a redesigned frontend.

TODOs:

- [x] Manual Review
- [x] Release
- [x] Promote / Demote / Reject course members
- [x] Display more user information in the members view
- [x] Implement rebuild
- [ ] General visual improvements

This PR includes both the current and redesigned frontend. Running QuickFeed enables a user to visit both (given that both are webpacked), but the default is the current frontend.

Fixes #569.

PR #573, opened 2022-01-17
Fix default values not updating in the database on UpdateReview

0xf8f8ff:
Gorm does not update default values in the database if Updates uses a struct, changed `UpdateReview` to use a map instead.


meling:
@0xf8f8ff Ups... I forgot to tell you that I've already fixed this issue in the frontend PR #502. See also issues #569 and #570. We should use the `Select()` method instead of the `map[string]interface{}` because this gives better type safety (can rename fields without the risk of forgetting to update the strings in these update commands.)

PR #583, opened 2022-01-24
Keep approved status for submissions with manual review

### LINKED ISSUE #582: Approved status of a submission resetting on a new push to the -labs repo
Approved status of a submissions for an assignment with manual review resets back to `Not approved` after a new push to the student's repo.

0xf8f8ff:
Resolves #582.

### DIFF
--- a/web/hooks/github.go
+++ b/web/hooks/github.go
@@ -190,6 +190,7 @@ func (wh GitHubWebHook) recordSubmissionWithoutTests(rData *ci.RunData) {
 		return
 	}
 	newSubmission := &pb.Submission{
+		ID:           newest.ID,
 		AssignmentID: rData.Assignment.ID,
 		BuildInfo: &score.BuildInfo{
 			BuildDate: time.Now().Format(pb.TimeLayout),
@@ -200,7 +201,8 @@ func (wh GitHubWebHook) recordSubmissionWithoutTests(rData *ci.RunData) {
 		CommitHash: rData.CommitID,
 		UserID:     rData.Repo.UserID,
 		GroupID:    rData.Repo.GroupID,
-		Status:     rData.Assignment.IsApproved(newest, newest.GetScore()),
+		Status:     newest.GetStatus(),
+		Released:   newest.GetReleased(),
 	}
 	if err := wh.db.CreateSubmission(newSubmission); err != nil {
 		wh.logger.Errorf("Failed to save submission for user %s, assignment %d: %v", rData.JobOwner, rData.Assignment.ID, err)
--- a/web/hooks/github.go
+++ b/web/hooks/github.go
@@ -13,6 +13,7 @@ import (
 	"github.com/autograde/quickfeed/log"
 	"github.com/google/go-github/v35/github"
 	"go.uber.org/zap"
+	"gorm.io/gorm"
 )
 
 // GitHubWebHook holds references and data for handling webhook events.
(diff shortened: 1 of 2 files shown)


web/hooks/github.go
```diff
@@ -177,23 +178,35 @@ func (wh GitHubWebHook) runAssignmentTests(assignment *pb.Assignment, repo *pb.R
...
 // recordSubmissionWithoutTests saves a new submission without running any tests
 // for a manually graded assignment.
-func (wh GitHubWebHook) recordSubmissionWithoutTests(data *ci.RunData) {
+func (wh GitHubWebHook) recordSubmissionWithoutTests(rData *ci.RunData) {
+	submissionQuery := &pb.Submission{
+		AssignmentID: rData.Assignment.GetID(),
+		UserID:       rData.Repo.GetUserID(),
+		GroupID:      rData.Repo.GetGroupID(),
+	}
+	newest, err := wh.db.GetSubmission(submissionQuery)
```
meling, line 187: Does `newest` have the new score? It is confusing where the score comes from. Or maybe what I don't quite understand is why `recordSubmissionWithoutTests` is triggered based on a push event, rather than a gRPC call to update the benchmark/manual review.

meling:
Could you add a test to simulate this? So that we know that the logic remains correct across other changes.

web/hooks/github.go
0xf8f8ff: `newest` is the student's previous submission in the database (if any), it can have a score > 0 (set from a manual review), it can already be approved as well. However, some students push updates to the same assignment after approval, and now the submission updates, the approved status disappears and the score gets reset to zero. This is what I'm trying to fix.

`recordSubmissionWithoutTests` only triggers on a push event: it creates a submission for this lab in the database (skipping all the steps related to the tests). Only after that it is possible to create a review (a review always "belongs" to a certain submission by a remote ID).

0xf8f8ff:
> Could you add a test to simulate this? So that we know that the logic remains correct across other changes.

Will do.

web/hooks/github.go
0xf8f8ff: Added a test, but I feel like it does not belong in the `github_test.go` file since it doesn't test anything hook-related directly and is only concerned with the database behavior. Should I move it to a separate file?

meling: Ok. I think I now understand. I suspect that I might forget this again by the next time I look at this code, so it would be nice to have a comment in the code to explain this.

I too think the `recordSubmissionWithoutTests` function and its test doesn't belong here. Please check my recent PR #588, where I've tried to decouple `ci.RunTests` from `ci.RecordResults` (these are now methods on the `RunData` struct). I think we could do less in `ci.RecordResults` and some how make it do the same as your revised `recordSubmissionWithoutTests`, eliminating the latter function (reusing `ci.RecordResults`). We would have to move out some of the code from the this function, e.g., having to do with slip days updating and the rebuild stuff. Please take a look and let me know if there are any synergies to be had here.

0xf8f8ff: Yes, I'm on it. Found another review-related bug while testing #588 (which exists in the main, so unrelated to the PR), but the rest seems to be working fine.

meling: I've refactored #588 to have `ci.RecordResults` support both manual review submissions and test run submissions. I think I've updated most of the relevant code, except I haven't migrated the test case that you wrote on this PR. Could you do that and push it to #588, please. And let me know if there are any issues with my approach?

0xf8f8ff: Updated/migrated the test. No issues found.

meling:
Closing in favor of #589.

PR #589, opened 2022-01-25
Refactored CI RunTests and RecordResults

### LINKED ISSUE #588: Refactor RunTests and RecordResults with less coupling
Current implementation of `ci.RunTests` calls `runTests` and `recordResults` as local functions in `ci` package and all these functions all need access to the `RunData` struct type. We should refactor these functions into methods on `RunData` and allow `RecordResults` to be called independently to record other results that are not due to running tests, e.g., for manually reviewed exercises (see #583). Furthermore, this refactor will also help to decouple functionality and make things more flexible, such as the ability to clone repositories before running tests from within quickfeed itself, rather than cloning as part of the `run.sh` script.

meling:
- Converted funcs to methods on RunData type
- Added run.sh script to testdata and load into Script
- Docker cleanup and simplification
- Cleanup Docker tests; added TestDockerPull
- Replace logger.Errorf with return fmt.Errorf
- Moved timeout code to separate method
- Separated RunTests and RecordResults
- Return pb.Submission from RecordResults
- Log notice about error not being a test failure

Fixes #588.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -14,7 +14,7 @@ jobs:
         uses: golangci/golangci-lint-action@v2
         with:
           version: latest
-          skip-go-installation: true
+          # skip-go-installation: true
           skip-pkg-cache: true
           skip-build-cache: true
           args: --disable errcheck --timeout 5m
--- a/ag/assignment.go
+++ b/ag/assignment.go
@@ -1,6 +1,7 @@
 package ag
 
 import (
+	context "context"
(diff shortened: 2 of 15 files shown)


0xf8f8ff (commented):
Everything looks fine. Didn't test things related to the docker scripts in depth, but the rest seems to work as expected.

0xf8f8ff (commented):
Everything looks fine. Migrated the test for RecordResults in case of submissions for manual review assignments from #583.

PR #595, opened 2022-01-30
Adds support for processing task-*.md files and GitHub issue creation

### LINKED ISSUE #302: Quickfeed should create issues on student accounts
Based on source files in markdown format, e.g. representing multiple choice or other type of Questions or other Tasks to be performed, Quickfeed should create an `Issue` on GitHub that can be automatically checked by Quickfeed once "submitted" by the student.

Students can "submit" answers to Questions by sending a Comment to the corresponding Issue. Quickfeed can already analyze multiple choice questions in markdown format. We could have other types questions that can be answered inline in comments. Similarly Tasks can be defined as an Issue, that students should close once they have completed a task.

meling:
This partially implements issue creation and syncing of tasks / exercises. It needs to be revised before merging.

Fixes #302

- Added github Enhancement plan markdown file
- Added small markdown fixies
- Added recommended changes in enhancement doc file
- fixed enhancement file
- Added Tasks creation
- Added git task creation and sync
- fixed test case
- Fixed git tasks file finder
- Moved issue design doc to doc/design-docs
- Some fixes and initial restructuring
- Revised Assignment and Task messages in proto file
- Refactored to use new readTestsRepositoryContent
- Added scripts folder to testdata/tests
- Added missing newlines
- Improved tests for run.sh scripts
- Renamed matchAll to matchAny
- Deprecated assignmentid in assignment.yml
- Replaced assignmentid with order in assignment.yml
- Added sort tasks by title to ensure consistency
- Added words to spell check list
- Added documentation for task-*.md files
- Fixed typo


scm/github.go
```diff
@@ -689,6 +689,108 @@ func (s *GithubSCM) GetUserScopes(ctx context.Context) *Authorization {
...
+			GitError: err,
+		}
+	}
+	s.logger.Debugf("CreateRepository: done creating issue %s", opt.Title)
+
+	return toIssue(issue), nil
+}
+
+// GetRepoIssue implements the SCM interface
+func (s *GithubSCM) GetRepoIssue(ctx context.Context, opt *IssueOptions) (*Issue, error) {
```
JosteinLindhom, line 722: `IssueOptions` looks very similar to `RepositoryOptions`.

You could do this:
```go
func (s *GithubSCM) GetRepoIssue(ctx context.Context, issueNumber int, opt *RepositoryOptions) (*Issue, error) {
	if !opt.valid() { // Valid as long as either opt.ID > 0 || (opt.Path != "" && opt.Owner != "")
		...
	}
	issue, _, err := s.client.Issues.Get(ctx, opt.Owner, opt.Path, issueNumber)
```

And similarly for `GetRepoIssues`, but omitting `issueNumber int`

JosteinLindhom: Doing this, and as mentioned below, you can remove `IssueOptions` altogether from `scm.go`.

scm/github_test.go
```diff
@@ -114,3 +114,123 @@ func TestCreateHook(t *testing.T) {
...
+	issueNumber := 20
+	// Creating new Client
+	s, err := scm.NewSCMClient(zap.NewNop().Sugar(), "github", accessToken)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ctx := context.Background()
+
+	opt1 := &scm.IssueOptions{
```
JosteinLindhom, line 219: By doing changes mentioned previously, this can be simplified to

```go
	opt := &scm.CreateIssueOptions{
		Organization: qfTestOrg,
		Repository:   repo,
		Title:        title,
		Body:         body,
	}

	_, err = s.EditRepoIssue(ctx, issueNumber, opt)
```

scm/github.go
JosteinLindhom: This also requires some small adjustments in the tests.

scm/github.go
```diff
@@ -689,6 +689,108 @@ func (s *GithubSCM) GetUserScopes(ctx context.Context) *Authorization {
...
+	var issues []*Issue
+	for _, issue := range issueList {
+		issues = append(issues, toIssue(issue))
+	}
+
+	return issues, nil
+}
+
+// EditRepoIssue implements the SCM interface
+func (s *GithubSCM) EditRepoIssue(ctx context.Context, opt *IssueOptions, isu *CreateIssueOptions) (*Issue, error) {
```
JosteinLindhom, line 766: Similarly to `GetRepoIssue` and `GetRepoIssues` you can remove `opt *IssueOptions` here and rather use `issueNumber int`.

database/database.go
```diff
@@ -129,6 +129,23 @@ type Database interface {
...
+	// DeleteTask deletes specified task
+	DeleteTask(task *pb.Task) error
+	// UpdateTasks updates slice of tasks
+	UpdateTasks(tasks []*pb.Task) (err error)
+	// GetTasks gets tasks based on query
+	GetTasks(query *pb.Task) ([]*pb.Task, error)
+	// CreateIssues creates a batch of issues
+	CreateIssues(issues []*pb.Issue) error
+	// UpdateIssues updates a batch of issues
+	UpdateIssues(issues []*pb.Issue) (err error)
```
JosteinLindhom, line 146: These errors don't need to be named

assignments/assignments.go
```diff
@@ -49,6 +52,13 @@ func UpdateFromTestsRepo(logger *zap.SugaredLogger, db database.Database, course
 		return
 	}
 	logger.Debugf("Assignments for %s successfully updated from '%s' repo", course.GetCode(), pb.TestsRepo)
+
+	ctx := context.Background()
+	err = handleTasks(ctx, db, scm, course, getTasksFromAssignments(ctx, assignments))
+	if err != nil {
+		logger.Errorf("Failed to Create tasks on '%s' repository: %v", pb.TestsRepo, err)
```
meling, line 59: `Create` --> `create`

assignments/assignments.go
```diff
@@ -49,6 +52,13 @@ func UpdateFromTestsRepo(logger *zap.SugaredLogger, db database.Database, course
 		return
 	}
 	logger.Debugf("Assignments for %s successfully updated from '%s' repo", course.GetCode(), pb.TestsRepo)
+
+	ctx := context.Background()
```
meling, line 56: Move this `ctx` line early in this method; above `fetchAssignments` and use it also in `fetchAssignments` instead of `context.Background()`. We typically use the same context for all calls that we may need to cancel or timeout... although we don't do that here at the moment.

assignments/assignments.go
```diff
@@ -49,6 +52,13 @@ func UpdateFromTestsRepo(logger *zap.SugaredLogger, db database.Database, course
...
+	err = handleTasks(ctx, db, scm, course, getTasksFromAssignments(ctx, assignments))
```
meling, line 57: Several things about `getTasksFromAssignments`:
- does not use/need `ctx`
- we should pass `assignments` into `handleTasks` and extract the tasks therein.
- that way we avoid calling the method here.

assignments/walk_tests_repo.go
```diff
@@ -0,0 +1,156 @@
...
+				defaultScriptTemplate = string(contents)
+			}
+
+		case dockerfile:
+			courseDockerfile = string(contents)
+		}
+
+		if match(filepath.Base(path), taskFilePattern) {
+			assignment := assignmentsMap[assignmentName]
+			// Probably a better way of acquiring task name
```
meling, line 101: Try this:
```go
filepath.Join(assignmentName, filepath.Base(path))
```

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	"context"
+	"fmt"
+	"path/filepath"
+
+	pb "github.com/autograde/quickfeed/ag"
+	"github.com/autograde/quickfeed/database"
+	"github.com/autograde/quickfeed/scm"
+)
+
+// Things to do:
```
meling, line 14: Please use the marker `TODO(name): <task to do>` since this allows easier to search for TODO issues in code.

assignments/tasks_test.go
```diff
@@ -0,0 +1,539 @@
...
+	if err != nil {
+		return db, cleanup, err
+	}
+
+	// Create repositories with issues
+	n := 2
+	for _, repo := range repos {
+		var user *pb.User
+		// Might not even be necessary to handle repos differently in these tests.
+		dbRepo := &pb.Repository{}
```
meling, line 64: As the linter tells this allocation is never used, it is just overwritten. The correct way to do this is:
```go
		var dbRepo *pb.Repository
```

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+
+// Creates an issue on specified repository.
+func createScmIssue(c context.Context, sc scm.SCM, course *pb.Course, repo *pb.Repository, task *pb.Task) (*scm.Issue, error) {
+	issueOptions := &scm.CreateIssueOptions{
+		Organization: course.Name,
+		Repository:   filepath.Base(repo.GetHTMLURL()), // Needs to be of type "tests", not "https://github.com/qf101/tests". This is a very hacky solution. pb.Repository should probably have a field "Name" that is set upon creation.
+		Title:        task.Title,
+		Body:         task.Body,
+	}
+	issue, err := sc.CreateIssue(c, issueOptions)
```
meling, line 59: Here you can return directly.
```go
	return sc.CreateIssue(c, issueOptions)
```

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+		Body:         issue.Body,
+	}
+	return sc.EditRepoIssue(c, int(issue.IssueNumber), issueOptions)
+}
+
+// Creates an issue on specified repository.
+func createScmIssue(c context.Context, sc scm.SCM, course *pb.Course, repo *pb.Repository, task *pb.Task) (*scm.Issue, error) {
+	issueOptions := &scm.CreateIssueOptions{
+		Organization: course.Name,
+		Repository:   filepath.Base(repo.GetHTMLURL()), // Needs to be of type "tests", not "https://github.com/qf101/tests". This is a very hacky solution. pb.Repository should probably have a field "Name" that is set upon creation.
```
meling, line 55: I've added a helper method to `pb.Repository` that you can use here: `repo.Name()`

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	return task, nil
+}
+
+// Updates an issue on specified repository
+func updateScmIssue(c context.Context, sc scm.SCM, course *pb.Course, repo *pb.Repository, issue *pb.Issue) (*scm.Issue, error) {
+	issueOptions := &scm.CreateIssueOptions{
+		Organization: course.Name,
+		Repository:   filepath.Base(repo.GetHTMLURL()), // Todo
```
meling, line 44: I've added a helper method to `pb.Repository` that you can use here: `repo.Name()`

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+
+	pb "github.com/autograde/quickfeed/ag"
+	"github.com/autograde/quickfeed/database"
+	"github.com/autograde/quickfeed/scm"
+)
+
+// Things to do:
+// - Repositories in database do not have a "Name"-field.
+// - Ordering of tasks (See teacher.md)
+
```
meling, line 17: Suggest to add this helper function:
```go
// taskName returns the task name as a combination of assignmentName/filename
// excluding the task- prefix and the .md suffix.
func taskName(assignmentName, basePath string) string {
	taskName := basePath[len("task-"):]
	taskName = taskName[:len(taskName)-len(".md")]
	return filepath.Join(assignmentName, taskName)
}
```

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+// newTask returns a task from markdown contents and associates it with the given assignment.
+// The provided markdown contents must contain a title specified on the first line,
+// starting with the "# " character sequence, followed by two new line characters.
+func newTask(contents []byte, assignment *pb.Assignment, name string) (*pb.Task, error) {
+	if !bytes.HasPrefix(contents, []byte("# ")) {
+		return nil, fmt.Errorf("task for assignment %s does not start with a # title marker", assignment.Name)
```
meling, line 23: This error message should be updated to reference the given task `name`.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	bodyIndex := bytes.Index(contents, []byte("\n\n"))
+	if bodyIndex == -1 {
+		return nil, fmt.Errorf("failed to find task body in %s", assignment.Name)
```
meling, line 27: This error message should be updated to reference the given task `name`.

assignments/walk_tests_repo.go
```diff
@@ -0,0 +1,156 @@
...
+			task, err := newTask(contents, assignment, assignmentName+"/"+filepath.Base(path))
```
meling, line 102: Assuming you add the `taskName()` helper function in `tasks.go` you can do this:
```go
	taskName := taskName(assignmentName, filepath.Base(path))
	task, err := newTask(contents, assignment, taskName)
```
Or better avoid passing `assignment` to `newTask` by passing instead the `uint32`: `assignment.GetOrder()`. See `tasks.go` for additional comments.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	task := &pb.Task{
+		AssignmentOrder: assignment.GetOrder(),
```
meling, line 31: Could be passed in as a `assignmentOrder uint32` instead of `pb.Assignment`.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	for _, existingTask := range existingTasks {
+		task, ok := taskMap[existingTask.Name]
+		if !ok {
+			// There exists a task in db, that is not represented by a task found in scm.
+			db.DeleteTask(existingTask)
+			continue
+		}
+		if !(task.Title == existingTask.Title && task.Body == existingTask.Body) {
```
meling, line 149: This boolean expression can be simplified using De Morgan's laws:
```go
	if task.Title != existingTask.Title || task.Body != existingTask.Body {
```
But perhaps you should move this check to a method on `Task` in the `ag` package:
```go
func (t *Task) HasChanged(new *Task) bool {
     return t.Title != new.Title || t.Body != new.Body
}
```
Should also add a simple test case for this `HasChanged()` method.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+func getTasksFromAssignments(c context.Context, assignments []*pb.Assignment) []*pb.Task {
+	tasks := []*pb.Task{}
+	for _, assignment := range assignments {
+		tasks = append(tasks, assignment.Tasks...)
+	}
+
+	return tasks
+}
+
+func handleTasks(c context.Context, db database.Database, s scm.SCM, course *pb.Course, tasks []*pb.Task) error {
```
meling, line 89: This is doing two different things and should be decoupled:

- Make new database method `SynchronizeTasks(course, tasks)` or `SynchronizeAssignmentTasks(course, tasks)`. This method should do the `GetAssignmentsByCourse()` and the loop over assignments and update the tasks as you do in the current `synchronizeTasks()` function below. Note that this method will not need the context.Context parameter since that's not used in the database.
- A separate method to synchronize the issues... (haven't got a specific suggestion at the moment... maybe soon, but it should only deal with updating the issues on the repos.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		return err
+	}
+
+	assignments, err := db.GetAssignmentsByCourse(course.GetID(), false)
+	if err != nil {
+		return err
+	}
+
+	// Loops through all assignments
```
meling, line 102: Unnecessary comment.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	for _, assignment := range assignments {
+		err := synchronizeTasks(c, db, assignment, tasks)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Loops through all student repos
```
meling, line 110: Better comment would be:
```go
// Synchronize issues on all student repositories.
```
(However, perhaps this could be a top-level comment for the new method described above.)

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	for _, repo := range repos {
+		if !repo.IsStudentRepo() {
+			continue
+		}
+
+		err = synchronizeIssues(c, db, course, s, repo, tasks)
```
meling, line 116: Should be moved into the if statement:
```go
if err = sync... ; err != nil {
```

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// synchronizeTasks synchronizes tasks in the database, with the ones of given assignment.
+// Returns a slice of tasks as they appear in the database.
+func synchronizeTasks(c context.Context, db database.Database, assignment *pb.Assignment, tasks []*pb.Task) error {
```
meling, line 127: Convert this function into a database method as suggested above.
- `context.Context` not used in this function.

meling: btw: when moving it to database package, remember to bring along any tests.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	taskMap := make(map[string]*pb.Task)
+	for _, task := range tasks {
+		if task.AssignmentOrder == assignment.Order {
+			taskMap[task.Name] = task
+		}
+	}
+
+	existingTasks, err := db.GetTasks(&pb.Task{AssignmentID: assignment.GetID()})
+	if err != nil {
+		return err
```
meling, line 139: Check if this database lookup returns an error if there are no tasks associated with `assignment.GetID()`. Typically, such lookups would return `gorm.ErrRecordNotFound`. Suggest to add test for such scenarios and adjust this check accordingly. That is, typically it would be something like this:

```go
	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
		return fmt.Errorf("failed to get tasks for assignment %d: %w", assignment.GetID(), err)
	}
```

assignments/tasks.go
meling: Moreover, you could write this more like a database transaction that does multiple steps... Look at `CreateSubmission()` in `gormdb_submission.go` for an example.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	for _, existingTask := range existingTasks {
+		task, ok := taskMap[existingTask.Name]
+		if !ok {
+			// There exists a task in db, that is not represented by a task found in scm.
+			db.DeleteTask(existingTask)
```
meling, line 146: Should use `_ = db.DeleteTask ...` to indicate that we ignore the error explicitly.

assignments/tasks.go
meling: By making a new `SynchronizeTasks()` database method, we can actually remove these three methods from the `Database` interface:
```go
	// CreateTasks creates slice of tasks
	CreateTasks(tasks []*pb.Task) error
	// DeleteTask deletes specified task
	DeleteTask(task *pb.Task) error
	// UpdateTasks updates slice of tasks
	UpdateTasks(tasks []*pb.Task) error
```
Each of these are used only once in connection with the sync operation. Thus, they can be made unexported and not part of the interface.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+		delete(taskMap, existingTask.Name)
+	}
+
+	// Only tasks that there is no existing record of remains
+	for _, task := range taskMap {
+		tasksToBeCreated = append(tasksToBeCreated, task)
+	}
+
+	err = db.CreateTasks(tasksToBeCreated)
```
meling, line 163: These two methods `CreateTasks` and `UpdateTasks` can probably be combined into the transaction I mentioned in another comment.

Otherwise, if you opt to not do a transaction, please make the calls internal to the `if err := db.... ; err != nil`.

Actually, the last method can simply `return db.Update ...`

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+// synchronizeIssues synchronizes database and scm with issues based on tasks found
+func synchronizeIssues(c context.Context, db database.Database, course *pb.Course, s scm.SCM, repo *pb.Repository, tasks []*pb.Task) error {
+	issuesToBeCreated := []*pb.Issue{}
+	issuesToBeUpdated := []*pb.Issue{}
+	tasksMap := make(map[string]*pb.Task)
+	for _, task := range tasks {
+		tasksMap[task.Name] = task
+	}
+
+	// Loops through existing issues on repo.
```
meling, line 184: Revise comment??:
```go
// Find existing issues to be updated
```

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+		Name:         task.Name,
+		Title:        task.Title,
+		Body:         task.Body,
+	}
+	return issue, nil
+}
+
+// Following is Oje code (placement might be temporary):
+
+func getTasksFromAssignments(c context.Context, assignments []*pb.Assignment) []*pb.Task {
```
meling, line 80: Seems like the return type from this would be better suited if it was a `map[string]*pb.Task`, which is what is used by the synchronize functions, avoiding the extra loops therein.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	issue, err := sc.CreateIssue(c, issueOptions)
+	if err != nil {
+		return nil, err
+	}
+	return issue, nil
+}
+
+// This is more of a converter function, and cannot currently return an error. Should probably be renamed or something.
+func createIssue(c context.Context, repo *pb.Repository, task *pb.Task, scmIssue *scm.Issue) (*pb.Issue, error) {
```
meling, line 67: No need for the `context.Context` parameter.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	// Loops through existing issues on repo.
+	for _, issue := range repo.Issues {
+		task, ok := tasksMap[issue.Name]
+		if !ok {
+			// What should happen if task does not exist for issue?
+			continue
+		}
+		// Might be necessary to check if the issue is closed here, otherwise we might end up editing already closed issues.
+		// If this becomes a necessity, then the issue data-record must have a closed (true/false) field, or a status field.
+		if !(task.Title == issue.Title && task.Body == issue.Body) {
```
meling, line 193: Can reuse the `task.HasChanged()` method

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	err = db.UpdateIssues(issuesToBeUpdated)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Only used for testing
```
meling, line 237: Should be moved to the relevant `_test.go` file, or deleted if no longer needed. Ideally, the "real" function should also be used for testing. E.g., if you need to test without actually hitting GitHub you can use the FakeSCM thing.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	err = db.UpdateTasks(tasksToBeUpdated)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// synchronizeIssues synchronizes database and scm with issues based on tasks found
+func synchronizeIssues(c context.Context, db database.Database, course *pb.Course, s scm.SCM, repo *pb.Repository, tasks []*pb.Task) error {
```
meling, line 176: Ideally, we should avoid updating the database and scm (GitHub) in the same method... Still thinking about better decoupling here. Meanwhile, could you add a doc comment above that explains what is stored in the database and why that is needed.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	issuesToBeCreated := []*pb.Issue{}
+	issuesToBeUpdated := []*pb.Issue{}
+	tasksMap := make(map[string]*pb.Task)
```
meling, line 179: Could instead pass in the `taskMap` directly instead of `[]*pb.Tasks`

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+
+	return nil
+}
+
+// synchronizeTasks synchronizes tasks in the database, with the ones of given assignment.
+// Returns a slice of tasks as they appear in the database.
+func synchronizeTasks(c context.Context, db database.Database, assignment *pb.Assignment, tasks []*pb.Task) error {
+	tasksToBeCreated := []*pb.Task{}
+	tasksToBeUpdated := []*pb.Task{}
+	taskMap := make(map[string]*pb.Task)
```
meling, line 130: Could instead pass in the `taskMap` directly instead of `[]*pb.Tasks`... it is not totally clear to me if you need the check for `assignment.Order`? If needed, you can still iterate over the map I guess.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	return issue, nil
+}
+
+// This is more of a converter function, and cannot currently return an error. Should probably be renamed or something.
+func createIssue(c context.Context, repo *pb.Repository, task *pb.Task, scmIssue *scm.Issue) (*pb.Issue, error) {
+	issue := &pb.Issue{
+		RepositoryID: repo.ID,
+		IssueNumber:  uint64(scmIssue.IssueNumber),
+		Name:         task.Name,
```
meling, line 71: Shouldn't the `Issue` type instead just reference `Task.ID` with a field `TaskID` which is declared a foreign key. Otherwise you will store the `task.Body` for every student repository in the database, which could get big.

assignments/tasks.go
```diff
@@ -0,0 +1,284 @@
...
+
+	// Updates issues on scm.
+	for _, issue := range issuesToBeUpdated {
+		_, err := updateScmIssue(c, s, course, repo, issue)
+		if err != nil {
+			return err
+		}
+	}
+
+	err := db.CreateIssues(issuesToBeCreated)
```
meling, line 225: The `CreateIssue` and `UpdateIssue` database methods are only used once; meaning it might be better to just combine them somehow into a single method to avoid bloating the database interface. Recall also error handling and returning error as explained for the other sync func above.

meling (commented):
Seems I cannot request changes myself, since I created the PR. But please see the comments and fix.

ag/task.go
```diff
@@ -0,0 +1,6 @@
+package ag
+
+// HasChanged returns true if a task has a different body or title than another.
```
meling, line 3: `... if task t has a ... than the new task`

ag/task_test.go
```diff
@@ -0,0 +1,68 @@
...
+		Title:           "This is the original task",
+		Body:            "Description description",
+		Name:            "lab1/1",
+	}
+
+	// Comparing the original task with itself.
+	wantResult := false
+	gotResult := originalTask.HasChanged(originalTask)
+
+	if diff := cmp.Diff(wantResult, gotResult, protocmp.Transform()); diff != "" {
```
meling, line 26: It is a bit overkill to use `cmp.Diff` on bool results since it is easy to compare with plain Go:

```go
if want != got {
  t.Errorf("%v.HasChanged(%v) = %t, expected %t", originalTask, originalTask, got, want)
}
```
Note that the two inputs may print too much, so maybe that's where you would want to compare the two tasks instead of the bool values... Not sure what is best here... I guess what allows to see why it might fail.

This type of test might lend itself to so called [table-driven tests](https://github.com/golang/go/wiki/TableDrivenTests) because you have multiple different task examples to compare; which could be defined separately. Then use something like this in the Go test struct:

```go
inTask1, inTask2 pb.Task
want bool
```

meling: I thinks perhaps you can simplify the input tasks to the bare minimum to test the method in question, namely only define tasks with various configurations of the `Title` and `Body` fields, and ignore all the other fields... Thus the output using the `%v` verb would not be too verbose. And you can set the title and body fields to be short strings to avoid verbosity in the output.

assignments/assignments.go
```diff
@@ -21,26 +21,30 @@ import (
 // UpdateFromTestsRepo updates the database record for the course assignments.
 func UpdateFromTestsRepo(logger *zap.SugaredLogger, db database.Database, course *pb.Course) {
 	logger.Debugf("Updating %s from '%s' repository", course.GetCode(), pb.TestsRepo)
-	s, err := scm.NewSCMClient(logger, course.GetProvider(), course.GetAccessToken())
+	scm, err := scm.NewSCMClient(logger, course.GetProvider(), course.GetAccessToken())
+	ctx := context.Background()
```
meling, line 25: Move this line after the err check.

assignments/assignments.go
```diff
@@ -49,6 +53,12 @@ func UpdateFromTestsRepo(logger *zap.SugaredLogger, db database.Database, course
 		return
 	}
 	logger.Debugf("Assignments for %s successfully updated from '%s' repo", course.GetCode(), pb.TestsRepo)
+
+	err = handleTasks(ctx, db, scm, course, assignments)
```
meling, line 57: Move into if statement.

assignments/assignments.go
```diff
@@ -21,26 +21,30 @@ import (
...
+
 	if dockerfile != "" && dockerfile != course.Dockerfile {
 		course.Dockerfile = dockerfile
 		if err := db.UpdateCourse(course); err != nil {
 			logger.Debugf("Failed to update Dockerfile for course %s: %s", course.GetCode(), err)
 			return
 		}
 	}
+
+	// Does not store tasks, since they need to be handled by HandleTasks
```
meling, line 47:
```go
// Does not store tasks associated with assignments; tasks are handled separately by handleTasks below
```

assignments/tasks.go
```diff
@@ -0,0 +1,152 @@
...
+func newTask(contents []byte, assignmentOrder uint32, name string) (*pb.Task, error) {
+	if !bytes.HasPrefix(contents, []byte("# ")) {
+		return nil, fmt.Errorf("task with name: %s, does not start with a # title marker", name)
+	}
+	bodyIndex := bytes.Index(contents, []byte("\n\n"))
+	if bodyIndex == -1 {
+		return nil, fmt.Errorf("failed to find task body in task: %s", name)
+	}
+
+	task := &pb.Task{
```
meling, line 34: Can return directly here:
```go
return &pb.Task{
	AssignmentOrder: assignmentOrder,
	Title:           string(contents[2:bodyIndex]),
	Body:            string(contents[bodyIndex+2:]),
	Name:            name,
}, nil
```

database/gormdb_tasks.go
```diff
@@ -0,0 +1,127 @@
...
+
+import (
+	"errors"
+	"fmt"
+
+	pb "github.com/autograde/quickfeed/ag"
+	"gorm.io/gorm"
+)
+
+// TODO(Meling): Methods such as GetTasks and CreateTasks are not necessary, except for in tests. They therefore need to be a part of the interface, even though they are not actually used.
```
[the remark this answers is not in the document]
oleespe: If I make `gormdb_tasks_test.go` part of the `database` package I can no longer use `qtest.TestDB()` to create a test database (this leads to an import cycle anyways). How do you want me to proceed?

assignments/tasks.go
```diff
@@ -0,0 +1,158 @@
...
+	// Creating issues in database, based on issues created on scm.
+	return db.CreateIssues(createdIssues)
+}
+
+// createIssues creates issues on scm based on repository, course and tasks. Returns created issues.
+func createIssues(ctx context.Context, sc scm.SCM, course *pb.Course, repo *pb.Repository, tasks []*pb.Task) ([]*pb.Issue, error) {
+	createdIssues := []*pb.Issue{}
+	for _, task := range tasks {
+		issueOptions := &scm.CreateIssueOptions{
+			Organization: course.Name,
```
meling, line 113: Is this correct? The `course.Name` is the long name of the course, e.g., `Operating Systems and Systems Programming`, whereas `course.Tag` would typically be `DAT320` (capitalized). There is also `course.GetOrganizationPath()` that might be more accurate for this one...

oleespe: I don't know... IssueOptions was formatted like this when I started on this project. When running tests on the `qf101` org, `course.Name` is set to this value. Given what you said, I suppose `course.Tag` or `course.GetOrganizationPath()` should be used instead.

oleespe: According to the github api it should be as follows: `/repos/{owner}/{repo}/issues`, where `owner` is what `course.Name` sets atm

meling: Right, so `owner` in this context is the organization (like `dat320-2022`) or if a personal user (like `meling`), whereas `repo` is the various repos under said organization/user. Hence, this should be `course.GetOrganizationPath()`, which corresponds to this line in the proto file:

```proto
    string organizationPath        = 9;   // The organization's SCM name, e.g. uis-dat520-2020.
```

PR #602, opened 2022-02-03
Minor fixes to the redesigned frontend

JosteinLindhom:
- Made links open in a new window rather
- Added helpers for user, user repo, and group repo links
- Add alert to Dashboard
- Fixed secondary progress bar displaying it should not (display only if secondaryProgress > 0)
- Changed default behaviour of navbar & course cards to display when VISIBLE, and hide when HIDDEN
- Clone course to avoid course being mutated in state
- Redirect to dashboard if getAllCourseSubmissions fail
- Implemented a button that shows a spinner when awaiting functions calls
- Moved login button to the top rather than bottom of sidebar
- Display correct slipdays for selected enrollment in results view
- Missing imports
- Added action to approve all enrollments
- Remove enrollments successfully rejected from state
- Added alerthandler for editCourse
- Added specific alert in server returns an unauthenticated error
- Fixed check for whether user is course creator or not
- missing imports
- Disabled input field for group name if group is approved in old frontend
- Missing import

### DIFF
--- a/dev/src/Helpers.ts
+++ b/dev/src/Helpers.ts
@@ -238,22 +238,25 @@ export const defaultYear = (date: Date): number => {
     return (date.getMonth() <= 11 && date.getDate() <= 31) && date.getMonth() > 10 ? (date.getFullYear() + 1) : date.getFullYear()
 }
 
-export const generateSubmissionRows = (links: UserCourseSubmissions[], cellGenerator: (s: SubmissionLink) => RowElement, groupName?: boolean, assignmentID?: number): Row[] => {
+export const generateSubmissionRows = (links: UserCourseSubmissions[], cellGenerator: (s: SubmissionLink, e?: Enrollment) => RowElement, groupName?: boolean, assignmentID?: number): Row[] => {
     const state = useAppState()
+    const course = state.courses.find(c => c.getId() === state.activeCourse)
     return links?.map((link) => {
         const row: Row = []
         if (link.enrollment && link.user) {
-            row.push({ value: link.user.getName(), link: `https://github.com/${link.user.getLogin()}` })
+            const url = course ? userRepoLink(course, link.user) : userLink(link.user)
+            row.push({ value: link.user.getName(), link: url })
             groupName && row.push(link.enrollment.getGroup()?.getName() ?? "")
         } else if (link.group) {
-            row.push(link.group.getName())
+            const data: RowElement = course ? { value: link.group.getName(), link: groupRepoLink(course, link.group) } : link.group.getName()
+            row.push(data)
         }
         if (link.submissions) {
             for (const submissionLink of link.submissions) {
                 if (state.review.assignmentID > 0 && submissionLink.getAssignment()?.getId() != state.review.assignmentID) {
                     continue
                 }
-                row.push(cellGenerator(submissionLink))
+                row.push(cellGenerator(submissionLink, link.enrollment))
             }
         }
         return row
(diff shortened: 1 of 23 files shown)


dev/src/Helpers.ts
```diff
@@ -276,6 +279,32 @@ export const generateAssignmentsHeader = (base: RowElement[], assignments: Assig
     return base
 }
 
+const slugify = (str: string): string => {
```
meling, line 282: We shouldn't need to call this in the frontend. That is, we should instead make sure that students can only type allowed characters in the group name, and thus their github group repo name can be used directly on Line 306.

dev/src/Helpers.ts
```diff
@@ -276,6 +279,32 @@ export const generateAssignmentsHeader = (base: RowElement[], assignments: Assig
...
+    return `https://github.com/${user.getLogin()}`
+}
+
+export const userRepoLink = (course: Course, user: User): string => {
+    return `https://github.com/${course.getOrganizationpath()}/${user.getLogin()}-labs`
+}
+
+export const groupRepoLink = (course: Course, group: Group): string => {
+    course.getOrganizationpath()
+    return `https://github.com/${course.getOrganizationpath()}/${slugify(group.getName())}`
```
meling, line 306: Should not need to slugify if the database returns proper group names.

dev/src/components/ReviewPage.tsx
```diff
@@ -7,18 +7,24 @@ import ReviewForm from "./manual-grading/ReviewForm"
...
     const actions = useActions()
     const courseID = getCourseID()
-    const isCourseCreator = state.courses[courseID].getCoursecreatorid() === state.self.getId()
+    const history = useHistory()
     const [groupView, setGroupView] = useState<boolean>(false)
 
     useEffect(() => {
         if (!state.courseSubmissions[courseID]) {
-            actions.getAllCourseSubmissions(courseID)
+            // Redirect to home if the call to get all submissions fails
```
meling, line 22: Will that help? Does it mean the user must login again?

dev/src/overmind/actions.tsx
```diff
@@ -185,19 +186,36 @@ export const updateEnrollment = async ({ actions, effects }: Context, { enrollme
...
+            }
         } else {
             // If unsuccessful, alert user
             actions.alertHandler(response)
         }
     }
 }
 
+export const updateEnrollments = async ({ state, actions, effects }: Context, courseID: number): Promise<void> => {
+    if (confirm("Do you really want to approve all students?")) {
```
meling, line 208: Maybe: `Please confirm that you want to approve all students?`

dev/src/overmind/actions.tsx
```diff
@@ -616,7 +650,10 @@ export const isAuthorizedTeacher = async ({ effects }: Context): Promise<boolean
 }
 
 export const alertHandler = ({ state }: Context, response: IGrpcResponse<unknown>): void => {
-    if (response.status.getCode() >= 0) {
+    if (response.status.getCode() === StatusCode.UNAUTHENTICATED) {
+        // The user is not logged in.
+        state.alerts.push({ text: "Your session is not valid. Please refresh this page and log in to continue.", color: Color.RED })
```
meling, line 655: Does this mean were logged in, but the session became invalid... If so, then
```text
Your session has expired. Please reload page to log in.
```
On the other hand, if we get here, would it not be possible to trigger a reload of the page? I guess then we should instead just say:
```text
Your session has expired. Please log in again.
```

meling (changes requested):
Looks good to me! Minor changes needed. But could you also check the deepsource.io warnings if there are some that are important to fix.

dev/src/components/ReviewPage.tsx
JosteinLindhom: After changing the behaviour of `success(response)` to reload the page if the user session is expired/invalid, this is no longer necessary.

PR #607, opened 2022-02-11
Frontend sorting

JosteinLindhom:
This PR implements sorting and filtering of submissions in both the Results and Review view.

You can sort and filter submissions per assignment, or for all assignments.

Submissions can be sorted by status and score
Rows can be filtered by teacher status and whether a submission is approved or not.


dev/src/overmind/state.tsx
```diff
@@ -187,7 +207,96 @@ export const state: State = {
...
+                    filteredSubmissions = filteredSubmissions.filter(submission => {
+                        return submission.enrollment ? !isTeacher(submission.enrollment) : false
+                    })
+                    break
+                case "approved":
+                    // approved filters all entries where all assignments have been approved
+                    filteredSubmissions = filteredSubmissions.filter(submission => {
+                        if (rootState.review.assignmentID > 0) {
+                            // If a specific assignment is selected, filter by that assignment
+                            const sub = submission.submissions?.find(submission => submission.getAssignment()?.getId() === rootState.review.assignmentID)?.getSubmission()
```
meling, line 241: Does this return a submission `sub` that is then used as a boolean below? Maybe also use an extra line for the predicate evaluation.

dev/src/overmind/state.tsx
```diff
@@ -187,7 +207,96 @@ export const state: State = {
...
+            }
+        }
+
+        const m = state.sortAscending ? -1 : 1
+        const sortedSubmissions = Object.values(filteredSubmissions).sort((a, b) => {
+            let subA: Submission | undefined
+            let subB: Submission | undefined
+            if (rootState.review.assignmentID > 0) {
+                // If a specific assignment is selected, sort by that assignment
+                subA = a.submissions?.find(submission => submission.getAssignment()?.getId() === rootState.review.assignmentID)?.getSubmission()
```
meling, line 262: These are similar to Line 241. Would it make sense to use a helper or something for this?

dev/src/overmind/state.tsx
```diff
@@ -187,7 +207,96 @@ export const state: State = {
...
+                        const sA = subA?.getScore()
+                        const sB = subB?.getScore()
+                        if (sA !== undefined && sB !== undefined) {
+                            return m * (sB - sA)
+                        } else if (sA !== undefined) {
+                            return -m
+                        }
+                        return m
+                    }
+                    const aSubs = a.submissions?.reduce((acc, cur) => { return acc + (cur.hasSubmission() ? (cur.getSubmission() as Submission).getScore() : 0) }, 0) ?? 0
```
meling, line 278: Quite long line. And a bit tricky to see what's going on. Especially at the end, with many zero values. Same comment for Lines 287-288. Maybe a helper function can alleviate the problem.

dev/src/components/ReviewPage.tsx
```diff
@@ -65,10 +68,11 @@ const ReviewPage = (): JSX.Element => {
                 <div className={state.review.assignmentID >= 0 ? "col-md-4" : "col-md-6"}>
                     <Search placeholder={"Search by name ..."} >
                         <Button type={ButtonType.BUTTON}
-                            text={groupView ? "View by group" : "View by student"}
-                            onclick={() => { setGroupView(!groupView); actions.review.setAssignmentID(-1) }}
+                            text={groupView ? "View by student" : "View by group"}
```
meling, line 71: Maybe the string `View by ` can be moved first and concatenate with `groupView ? "student" : "group"`

BTW: Is it correct that `groupView === true` should display `student`?

JosteinLindhom: I will concatenate the string.

Regarding `groupView === true -> student`: I'm not entirely sure what makes the most sense.

The change makes the button tell the user what will happen when s/he clicks the button. The previous behaviour told the user what s/he was currently viewing.

i.e. the button now communicates the resulting action of clicking it, rather than the current state.

meling: Good point, didn’t think about that

dev/src/overmind/state.tsx
JosteinLindhom: Agreed. Will add helpers for this and other areas you mention.

dev/src/overmind/state.tsx
JosteinLindhom: The conditional checks the truthy value of `sub` and short circuits if `.find(...)` returns `undefined`, i.e. it does not exist.

As `sub` can only be of the type `undefined` or `Submission`, this is shorthand for `sub !== undefined && isApproved(sub)`.

meling (approved):

PR #609, opened 2022-02-22
Accept repository invitations on user's behalf

### LINKED ISSUE #540: Remove unnecessary GitHub invitations
An invitation to the `info` repository is being sent to the enrolled students, but it is unnecessary because the repository is public and everyone in the course organization will already have read access to this repo.

meling:
First stab at automatically accepting invitations.

Fixes #540

### DIFF
--- a/database/gormdb_repository.go
+++ b/database/gormdb_repository.go
@@ -52,18 +52,6 @@ func (db *GormDB) GetRepositories(query *pb.Repository) ([]*pb.Repository, error
 	return repos, nil
 }
 
-func (db *GormDB) GetRepositoryByType(userID uint64, orgID uint64, repoType pb.Repository_Type) (*pb.Repository, error) {
-	var repo pb.Repository
-	query := &pb.Repository{OrganizationID: orgID, RepoType: repoType}
-	if repoType == pb.Repository_USER {
-		query.UserID = userID
-	}
-	if err := db.conn.First(&repo, query).Error; err != nil {
-		return nil, err
-	}
-	return &repo, nil
-}
-
 // DeleteRepositoryByRemoteID deletes repository by provider's ID
 func (db *GormDB) DeleteRepositoryByRemoteID(rid uint64) error {
 	repo, err := db.GetRepositoryByRemoteID(rid)
--- a/database/gormdb_repository.go
+++ b/database/gormdb_repository.go
@@ -52,6 +52,18 @@ func (db *GormDB) GetRepositories(query *pb.Repository) ([]*pb.Repository, error
 	return repos, nil
 }
 
+func (db *GormDB) GetRepositoryByType(userID uint64, orgID uint64, repoType pb.Repository_Type) (*pb.Repository, error) {
+	var repo pb.Repository
+	query := &pb.Repository{OrganizationID: orgID, RepoType: repoType}
+	if repoType == pb.Repository_USER {
+		query.UserID = userID
+	}
+	if err := db.conn.First(&repo, query).Error; err != nil {
+		return nil, err
+	}
+	return &repo, nil
+}
+
 // DeleteRepositoryByRemoteID deletes repository by provider's ID
 func (db *GormDB) DeleteRepositoryByRemoteID(rid uint64) error {
 	repo, err := db.GetRepositoryByRemoteID(rid)
--- a/scm/github.go
+++ b/scm/github.go
@@ -689,6 +689,46 @@ func (s *GithubSCM) GetUserScopes(ctx context.Context) *Authorization {
 	return &Authorization{Scopes: gitScopes}
 }
 
+// GetRepositoryInvites implements the SCM interface
+func (s *GithubSCM) GetRepositoryInvites(ctx context.Context, opt *RepositoryOptions) ([]*github.RepositoryInvitation, error) {
+	if !opt.valid() {
+		return nil, ErrMissingFields{
+			Method:  "GetRepositoryInvites",
+			Message: fmt.Sprintf("%+v", opt),
+		}
+	}
+
+	invites, _, err := s.client.Repositories.ListInvitations(ctx, opt.Owner, opt.Path, nil)
+	if err != nil {
+		return nil, ErrFailedSCM{
+			Method:   "GetRepositoryInvites",
+			GitError: fmt.Errorf("failed to list repository invitations for organization %s and repository %s: %w", opt.Owner, opt.Path, err),
+			Message:  fmt.Sprintf("failed to list repository invitations for organization %s and repository %s", opt.Owner, opt.Path),
+		}
+	}
+	return invites, nil
+}
+
+// AcceptRepositoryInvite implements the SCM interface
+func (s *GithubSCM) AcceptRepositoryInvite(ctx context.Context, opt *RepositoryInvitationOptions) error {
+	if !opt.valid() {
+		return ErrMissingFields{
+			Method:  "AcceptRepositoryInvite",
+			Message: fmt.Sprintf("%+v", opt),
+		}
+	}
+
+	_, err := s.client.Users.AcceptInvitation(ctx, int64(opt.InvitationID))
+	if err != nil {
+		return ErrFailedSCM{
+			Method:   "AcceptRepositoryInvite",
+			GitError: fmt.Errorf("failed to accept repository invitation %d: %w", opt.InvitationID, err),
+			Message:  fmt.Sprintf("failed to accept repository invitation %d", opt.InvitationID),
+		}
+	}
+	return nil
+}
+
 func toRepository(repo *github.Repository) *Repository {
 	return &Repository{
 		ID:      uint64(repo.GetID()),
--- a/scm/github.go
+++ b/scm/github.go
@@ -690,9 +690,9 @@ func (s *GithubSCM) GetUserScopes(ctx context.Context) *Authorization {
 }
 
 // GetRepositoryInvites implements the SCM interface
-func (s *GithubSCM) GetRepositoryInvites(ctx context.Context, opt *RepositoryInvitationOptions) ([]*RepositoryInvitation, error) {
+func (s *GithubSCM) AcceptRepositoryInvites(ctx context.Context, opt *RepositoryInvitationOptions) error {
 	if !opt.valid() {
-		return nil, ErrMissingFields{
+		return ErrMissingFields{
 			Method:  "GetRepositoryInvites",
 			Message: fmt.Sprintf("%+v", opt),
 		}
@@ -700,43 +700,26 @@ func (s *GithubSCM) GetRepositoryInvites(ctx context.Context, opt *RepositoryInv
 
 	invites, _, err := s.client.Users.ListInvitations(ctx, &github.ListOptions{})
 	if err != nil {
-		return nil, ErrFailedSCM{
+		return ErrFailedSCM{
 			GitError: fmt.Errorf("failed to fetch GitHub repository invitations: %w", err),
 			Method:   "GetRepositoryInvites",
 			Message:  "failed to fetch GitHub repository invitations",
 		}
 	}
 
-	scmInvites := make([]*RepositoryInvitation, 0)
 	for _, invite := range invites {
 		if invite.Repo.Owner.GetLogin() != opt.Owner || invite.Invitee.GetLogin() != opt.Login {
 			// Ignore unrelated invites
 			continue
 		}
-		scmInvite := &RepositoryInvitation{
-			ID:   invite.GetID(),
-			Repo: invite.GetRepo().GetName(),
-		}
-		scmInvites = append(scmInvites, scmInvite)
-	}
-	return scmInvites, nil
-}
 
-// AcceptRepositoryInvite implements the SCM interface
-func (s *GithubSCM) AcceptRepositoryInvite(ctx context.Context, opt *RepositoryInvitation) error {
-	if !opt.valid() {
-		return ErrMissingFields{
-			Method:  "AcceptRepositoryInvite",
-			Message: fmt.Sprintf("%+v", opt),
-		}
-	}
-
-	_, err := s.client.Users.AcceptInvitation(ctx, opt.ID)
-	if err != nil {
-		return ErrFailedSCM{
-			Method:   "AcceptRepositoryInvite",
-			GitError: fmt.Errorf("failed to accept repository invitation %d: %w", opt.ID, err),
-			Message:  fmt.Sprintf("failed to accept repository invitation %d", opt.ID),
+		_, err := s.client.Users.AcceptInvitation(ctx, invite.GetID())
+		if err != nil {
+			return ErrFailedSCM{
+				GitError: fmt.Errorf("failed to accept GitHub repository invitation: %w", err),
+				Method:   "GetRepositoryInvites",
+				Message:  "failed to accept GitHub repository invitation",
+			}
 		}
 	}
 	return nil
--- a/scm/github.go
+++ b/scm/github.go
@@ -701,9 +701,9 @@ func (s *GithubSCM) GetRepositoryInvites(ctx context.Context, opt *RepositoryInv
 	invites, _, err := s.client.Users.ListInvitations(ctx, &github.ListOptions{})
 	if err != nil {
 		return nil, ErrFailedSCM{
-			GitError: fmt.Errorf("failed to get GitHub repository invitations: %w", err),
+			GitError: fmt.Errorf("failed to fetch GitHub repository invitations: %w", err),
 			Method:   "GetRepositoryInvites",
-			Message:  fmt.Sprintf("failed to get GitHub repository invitations"),
+			Message:  "failed to fetch GitHub repository invitations",
 		}
 	}
 
--- a/scm/github.go
+++ b/scm/github.go
@@ -690,7 +690,7 @@ func (s *GithubSCM) GetUserScopes(ctx context.Context) *Authorization {
 }
 
 // GetRepositoryInvites implements the SCM interface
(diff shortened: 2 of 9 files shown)


database/gormdb_repository.go
```diff
@@ -52,6 +52,18 @@ func (db *GormDB) GetRepositories(query *pb.Repository) ([]*pb.Repository, error
 	return repos, nil
 }
 
+func (db *GormDB) GetRepositoryByType(userID uint64, orgID uint64, repoType pb.Repository_Type) (*pb.Repository, error) {
```
meling, line 55: Remove; see comment above.

web/courses.go
```diff
@@ -4,10 +4,12 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"strings"
 	"time"
 
 	pb "github.com/autograde/quickfeed/ag"
 	"github.com/autograde/quickfeed/scm"
+	"github.com/google/go-github/v35/github"
```
meling, line 12: We should avoid leaking `github` specific API outside the `scm` package.

scm/scm.go
```diff
@@ -62,6 +63,10 @@ type SCM interface {
 	RemoveMember(context.Context, *OrgMembershipOptions) error
 	// Lists all authorizations for authenticated user.
 	GetUserScopes(context.Context) *Authorization
+	// Lists all pending repository invites.
+	GetRepositoryInvites(context.Context, *RepositoryOptions) ([]*github.RepositoryInvitation, error)
```
meling, line 67: This should use a custom `scm`-specific return type to avoid leaking GitHub-specific types out of the `scm` package.

web/courses.go
```diff
@@ -442,6 +444,10 @@ func (s *AutograderService) enrollStudent(ctx context.Context, sc scm.SCM, enrol
 		if err := s.db.CreateRepository(&userRepo); err != nil {
 			return err
 		}
+
+		// repoTypes are the repositories we wish to accept invites for on behalf of the user
```
meling, line 448: Simplify comment and split long lines to make this more readable.
```go
		// repository types to accept invites for, on behalf of the user
		repoTypes := []pb.Repository_Type{
			pb.Repository_ASSIGNMENTS,
			pb.Repository_COURSEINFO,
			pb.Repository_USER,
		}
```

web/courses.go
```diff
@@ -533,6 +539,61 @@ func (s *AutograderService) setLastApprovedAssignment(submission *pb.Submission,
...
+			// and we only want to accept the invite for the
+			// user we are enrolling.
+			if *invite.Invitee.Login != user.Login {
+				continue
+			}
+			invites = append(invites, invite)
+		}
+	}
+
+	if !(len(invites) > 0) {
```
meling, line 570: This is more clear:
```go
if len(invites) == 0 {
```

web/courses.go
```diff
@@ -533,6 +539,61 @@ func (s *AutograderService) setLastApprovedAssignment(submission *pb.Submission,
...
+
+	user, err := s.db.GetUser(user.ID)
+	if err != nil {
+		s.logger.Errorf("Failed to get user %d: %v", user.ID, err)
+		return
+	}
+	// We need the user access token to accept invitations on their behalf
+	accessToken, err := user.GetAccessToken("github")
+	if err != nil {
+		fmt.Errorf("Failed to get access token for user %s: %v", user.Login, err)
```
meling, line 582: This would need a return in front of it... but see my comment above.

web/courses.go
```diff
@@ -533,6 +539,61 @@ func (s *AutograderService) setLastApprovedAssignment(submission *pb.Submission,
...
+			}
+			invites = append(invites, invite)
+		}
+	}
+
+	if !(len(invites) > 0) {
+		return
+	}
+
+	user, err := s.db.GetUser(user.ID)
```
meling, line 574: The `pb.User` object is already passed in as an argument to the `acceptRepositoryInvites` method; so this database lookup is unnecessary.

web/courses.go
```diff
@@ -533,6 +539,61 @@ func (s *AutograderService) setLastApprovedAssignment(submission *pb.Submission,
...
+	if err != nil {
+		s.logger.Errorf("Failed to get user %d: %v", user.ID, err)
+		return
+	}
+	// We need the user access token to accept invitations on their behalf
```
meling, line 579: We don't need this either. We already pass the `sc` object as input to the `acceptRepositoryInvites` method as well. That is, we only need to query for the access token in one place and we usually pass the `sc` object to those methods that need to talk to GitHub. We typically use this pattern:
```go
	usr, scm, err := s.getUserAndSCMForCourse(ctx, in.GetCourseID())
```
in various methods in `autograder_service.go`.

web/courses.go
```diff
@@ -533,6 +539,61 @@ func (s *AutograderService) setLastApprovedAssignment(submission *pb.Submission,
...
+	accessToken, err := user.GetAccessToken("github")
+	if err != nil {
+		fmt.Errorf("Failed to get access token for user %s: %v", user.Login, err)
+		return
+	}
+	sc, ok := s.scms.GetSCM(accessToken)
```
meling, line 585: We also don't need this as explained above.

web/courses.go
```diff
@@ -533,6 +539,61 @@ func (s *AutograderService) setLastApprovedAssignment(submission *pb.Submission,
...
+	if !ok {
+		s.logger.Errorf("GetSCM failed: could not get SCM for user: %s", user.Login)
+		return
+	}
+	for _, invite := range invites {
+		if err := sc.AcceptRepositoryInvite(ctx, &scm.RepositoryInvitationOptions{InvitationID: uint64(*invite.ID)}); err != nil {
+			s.logger.Errorf("Failed to accept repository invite for user %s for repository %s: %v", user.Login, invite.Repo, err)
```
meling, line 592: Error message is a bit long, what about:
```go
			s.logger.Errorf("Failed to invite user %s to repository %s: %v", user.Login, invite.Repo, err)
```

scm/scm.go
```diff
@@ -214,3 +219,8 @@ type Team struct {
 type Authorization struct {
 	Scopes []string
 }
+
+// RepositoryInvitation represents a repository invitation.
```
meling, line 223: Suggest to rename this:
```go
// RepositoryInvitation represents a repository invitation.
type RepositoryInvitation struct {
	ID   int64
	Repo string
}
```
Note also that it is better to use `int64` here to avoid type casting; since it is only used in this one place anyway. (I don't recall why we kept these IDs as `uint64`)
Also, I added the `Repo` field to allow printing the error message... but it does not seem to be necessary for the logic of the function.

web/courses.go
```diff
@@ -533,6 +539,61 @@ func (s *AutograderService) setLastApprovedAssignment(submission *pb.Submission,
 	return s.db.UpdateEnrollment(query)
 }
 
+// acceptRepositoryInvites tries to accept repository invitations for the given repository types on behalf of the given user.
+func (s *AutograderService) acceptRepositoryInvites(ctx context.Context, sc scm.SCM, user *pb.User, course *pb.Course, repoTypes []pb.Repository_Type) {
+	invites := []*github.RepositoryInvitation{}
```
meling, line 544: If you rename as I suggest in `scm.go`, you can do:
```go
	invites := make([]*scm.RepositoryInvitation, 0)
```
See other comments below.

web/courses.go
```diff
@@ -533,6 +539,61 @@ func (s *AutograderService) setLastApprovedAssignment(submission *pb.Submission,
...
+			continue
+		}
+		for _, invite := range repoInvites {
+			// Some repositories contain more than one invite
+			// and we only want to accept the invite for the
+			// user we are enrolling.
+			if *invite.Invitee.Login != user.Login {
+				continue
+			}
+			invites = append(invites, invite)
```
meling, line 566: This would become:
```go
			repoInvite := &scm.RepositoryInvitation{
				ID:   *invite.ID,
				Repo: invite.Repo.String(),
			}
			invites = append(invites, repoInvite)
```

web/courses.go
```diff
@@ -533,6 +539,61 @@ func (s *AutograderService) setLastApprovedAssignment(submission *pb.Submission,
...
+		fmt.Errorf("Failed to get access token for user %s: %v", user.Login, err)
+		return
+	}
+	sc, ok := s.scms.GetSCM(accessToken)
+	if !ok {
+		s.logger.Errorf("GetSCM failed: could not get SCM for user: %s", user.Login)
+		return
+	}
+	for _, invite := range invites {
+		if err := sc.AcceptRepositoryInvite(ctx, &scm.RepositoryInvitationOptions{InvitationID: uint64(*invite.ID)}); err != nil {
```
meling, line 591: Again, assuming the rename in `scm.go` and the above tweaks, you can do this directly:
```go
		if err := sc.AcceptRepositoryInvite(ctx, invite); err != nil {
```

web/courses.go
meling, line 544: If you rename as I suggest in `scm.go`, you can do:
```go
	invites := make([]*scm.RepositoryInvitation, 0)
```
See other comments below.

web/courses.go
```diff
@@ -533,6 +539,61 @@ func (s *AutograderService) setLastApprovedAssignment(submission *pb.Submission,
...
+	for _, repoType := range repoTypes {
+		repo, err := s.db.GetRepositoryByType(user.ID, course.OrganizationID, repoType)
+		if err != nil {
+			s.logger.Error("Failed to get repo: %s", err)
+			continue
+		}
+		// TODO: pb.Repository should contain Path
+		path := strings.Split(repo.GetHTMLURL(), "/")
+
+		repoInvites, err := sc.GetRepositoryInvites(ctx, &scm.RepositoryOptions{ID: repo.RepositoryID, Owner: course.OrganizationPath, Path: path[len(path)-1]})
```
meling, line 554: I'm a bit uncomfortable with the way you extract the path here. I'm not completely sure what it should be without testing things myself, but I see two solutions:
1. Add a `GetPath()` method in `ag/repo.go` to extract the path, you can add one that is similar to the `GetTestURL()`; this uses `LastIndex` instead of split, which is more efficient.
2. Another alternative is to only pass the RepoID and use the `GetRepository` method within `GetRepositoryInvites`. This is obviously less efficient, since it will talk with GitHub.

web/courses.go
```diff
@@ -533,6 +539,61 @@ func (s *AutograderService) setLastApprovedAssignment(submission *pb.Submission,
...
+		if err != nil {
+			s.logger.Errorf("Failed to get repository invites for %v: %s - %s", repo.RepoType, user.Login, err)
+			continue
+		}
+		for _, invite := range repoInvites {
```
meling, line 559: Suggest to move this loop into the `GetRepositoryInvites`; you only need to pass the `user.Login`. This would simplify the code here.

meling:
@JosteinLindhom Since I created this PR I cannot request changes via the pull request thing... weird. Anyway, I've reviewed it. Take a look at my suggestions.

web/courses.go
JosteinLindhom: I don't like it either. I was thinking that perhaps adding `Path` to the `pb.Repository` type and storing it in the database would be a good way to do this.

```go
// courses_new.go
// line 79:84
// repo scm.Repository
dbRepo := pb.Repository{
        OrganizationID: org.ID,
        RepositoryID:   repo.ID,
        HTMLURL:        repo.WebURL,
        RepoType:       pb.RepoType(path),
}
```
The `scm.Repository` already has Path, so we would just need to update the various functions that save repositories to the database.

Storing the `Path` is useful as we avoid having to extract anything from a string, and we already have to do a database lookup for the repository.

Then we could just do this:
```go
repoInvites, err := sc.GetRepositoryInvites(ctx, &scm.RepositoryOptions{ID: repo.RepositoryID, Owner: course.OrganizationPath, Path: repo.Path})
```

The use cases for this is obviously for this PR, in addition to, I believe, @oleespe's PR & Tasks work where having the path would be helpful.

This would not work for repositories already created (courses running this semester) without updating them in some way.

Alternatively I will go for your first suggested solution.

meling: Not sure we need to store it db, since it seems to contain the same info as HTMLURL minus some extra stuff... So I prefer to not add redundant data to the db. Note that, we may reconsider this later, e.g., after the semester, but it would require a more in-depth review of the existing usages of HTMLURL and Path etc. to see if some of these things can be harmonized and simplified. 

Note also that the doc comment for the `scm.RepositoryOptions` type; it says that you only need `ID` or both `Owner` and `Path`. That is, not all three. If you look at the other uses of this option, you'll see that id only `ID` is used we will first lookup the `Owner` and `Path` and use them to make the actual call to GitHub. (Exception: `GetRepository` can actually lookup the repository directly using only ID)

Suggest to go with my first suggestion for now.

web/courses.go
JosteinLindhom: I do not think there is a way around fetching the SCM of the enrollee. Accepting repository invitations has to be done using the access token of the invitee. The `sc` object passed in is that of the teacher, and does not have permission to accept invites for another user.

web/courses.go
JosteinLindhom: The user we get here:
```go
course, user := enrolled.GetCourse(), enrolled.GetUser()
```
does not come preloaded with `RemoteIdentity` (from `GetEnrollmentByCourseAndUser`). We could preload `RemoteIdentity` to avoid this lookup, but I need to check if that would have some unintended side-effects.

meling: That's strange because the code does seem to preload it:
```go
func (db *GormDB) GetUser(userID uint64) (*pb.User, error) {
	var user pb.User
	if err := db.conn.Preload("RemoteIdentities").First(&user, userID).Error; err != nil {
		return nil, err
	}
	return &user, nil
}
```

scm/github.go
```diff
@@ -689,6 +689,59 @@ func (s *GithubSCM) GetUserScopes(ctx context.Context) *Authorization {
...
+// GetRepositoryInvites implements the SCM interface
+func (s *GithubSCM) GetRepositoryInvites(ctx context.Context, opt *RepositoryInvitationOptions) ([]*RepositoryInvitation, error) {
+	if !opt.valid() {
+		return nil, ErrMissingFields{
+			Method:  "GetRepositoryInvites",
+			Message: fmt.Sprintf("%+v", opt),
+		}
+	}
+
+	invites, _, err := s.client.Users.ListInvitations(ctx, &github.ListOptions{})
```
meling, line 701: One potential issue with this implementation... what if a student (user) has unrelated invitations that he doesn't want to accept. Will the code below simply accept all invitations, including those unrelated to the user's QuickFeed courses?? We should test this somehow, to understand the behavior of what is returned from `ListInvitations`.

web/courses.go
meling: Looks like the new solution to use `s.getSCM()` for the user reasonable. Although I wish we didn't have to specify the provider here... But that's another issue.

web/courses.go
```diff
@@ -533,6 +535,34 @@ func (s *AutograderService) setLastApprovedAssignment(submission *pb.Submission,
...
+	userSCM, err := s.getSCM(ctx, user, "github")
+	if err != nil {
+		s.logger.Errorf("Failed to get SCM for user %d: %v", user.ID, err)
+		return
+	}
+	opts := &scm.RepositoryInvitationOptions{
+		Login: user.Login,
+		Owner: course.GetOrganizationPath(),
+	}
+	invites, err := userSCM.GetRepositoryInvites(ctx, opts)
```
meling, line 554: Suggest that we combine `GetRepositoryInvites` and `AcceptRepositoryInvite` into one method called `AcceptRepositoryInvites` that does both functionalities. I don't think we need the `Get` method for anything else. That way we reduce the methods in the `scm` interface as well.

meling (commented):
Again, can't formally request changes since I created the PR, but please check my suggestions @JosteinLindhom.

scm/github.go
JosteinLindhom: From my testing I found that we could filter out unrelated invitations by looking at the owner of the repository invite.

```go
	for _, invite := range invites {
		if invite.Repo.Owner.GetLogin() != opt.Owner || invite.Invitee.GetLogin() != opt.Login {
			// Ignore unrelated invites
			continue
		}
                ...
        }
```

Where the repository in `invite.Repo` looks like this:

```go
// invite.Repo
github.Repository{
    ...
    Owner: github.User{
        Login:"quickfeed-test-org", 
        ...
        HTMLURL:"https://github.com/quickfeed-test-org"
        Type:"Organization"
        ...
    }
    ...
}
```

Example of an unrelated invite:

```go
// Unrelated invite to a repo created by an individual
github.Repository{
    ...
    Owner: github.User{
        Login:"JosteinLindhom", 
        HTMLURL:"https://github.com/JosteinLindhom", 
        Type:"User"
        ...
    }
    ...
}
```

My understanding of how this works is that a repository can be owned by either an individual, or an organization. For our course repositories the owner will be an organization, whose `Login` will be equal to `course.GetOrganizationPath()`.

scm/scm.go
```diff
@@ -214,3 +216,8 @@ type Team struct {
 type Authorization struct {
 	Scopes []string
 }
+
+type RepositoryInvitationOptions struct {
```
meling, line 220: Would be nice to document what is the content of the two fields, cf. the other comment regarding the same.

scm/github.go
```diff
@@ -689,6 +689,42 @@ func (s *GithubSCM) GetUserScopes(ctx context.Context) *Authorization {
...
+
+	for _, invite := range invites {
+		if invite.Repo.Owner.GetLogin() != opt.Owner || invite.Invitee.GetLogin() != opt.Login {
+			// Ignore unrelated invites
+			continue
+		}
+
+		_, err := s.client.Users.AcceptInvitation(ctx, invite.GetID())
+		if err != nil {
+			return ErrFailedSCM{
```
meling, line 718: This error could now report which specific repository for which the invitation failed; I guess that could be useful to at least log in the system log.

PR #616, opened 2022-03-08
Added unique index to disallow courses with same code, year

JosteinLindhom:
- Add unique constraint for (course.code, course.year)
- Add check to see if a course violates unique constraint
- Add test to verify that courses with same code, year cannot be added
- Update tests where unique constraint was violated

### DIFF
--- a/database/gormdb_course.go
+++ b/database/gormdb_course.go
@@ -17,6 +17,9 @@ func (db *GormDB) CreateCourse(courseCreatorID uint64, course *pb.Course) error 
 	var courses int64
 	if err := db.conn.Model(&pb.Course{}).Where(&pb.Course{
 		OrganizationID: course.OrganizationID,
+	}).Or(&pb.Course{
+		Code: course.Code,
+		Year: course.Year,
 	}).Count(&courses).Error; err != nil {
 		return err
 	}
--- a/database/gormdb_test.go
+++ b/database/gormdb_test.go
@@ -177,9 +177,9 @@ func TestGormDBGetCourses(t *testing.T) {
 	defer cleanup()
 
 	admin := qtest.CreateFakeUser(t, db, 10)
-	c1 := &pb.Course{OrganizationID: 1}
-	c2 := &pb.Course{OrganizationID: 2}
-	c3 := &pb.Course{OrganizationID: 3}
+	c1 := &pb.Course{OrganizationID: 1, Year: 1}
+	c2 := &pb.Course{OrganizationID: 2, Year: 2}
+	c3 := &pb.Course{OrganizationID: 3, Year: 3}
 	qtest.CreateCourse(t, db, admin, c1)
 	qtest.CreateCourse(t, db, admin, c2)
 	qtest.CreateCourse(t, db, admin, c3)
@@ -333,10 +333,10 @@ func TestGormDBGetCoursesByUser(t *testing.T) {
 	defer cleanup()
 
 	admin := qtest.CreateFakeUser(t, db, 1)
-	c1 := &pb.Course{OrganizationID: 1}
-	c2 := &pb.Course{OrganizationID: 2}
-	c3 := &pb.Course{OrganizationID: 3}
-	c4 := &pb.Course{OrganizationID: 4}
+	c1 := &pb.Course{OrganizationID: 1, Year: 1}
+	c2 := &pb.Course{OrganizationID: 2, Year: 2}
+	c3 := &pb.Course{OrganizationID: 3, Year: 3}
+	c4 := &pb.Course{OrganizationID: 4, Year: 4}
 	qtest.CreateCourse(t, db, admin, c1)
 	qtest.CreateCourse(t, db, admin, c2)
 	qtest.CreateCourse(t, db, admin, c3)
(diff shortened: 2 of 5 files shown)


database/gormdb_test.go
```diff
@@ -177,9 +177,9 @@ func TestGormDBGetCourses(t *testing.T) {
 	defer cleanup()
 
 	admin := qtest.CreateFakeUser(t, db, 10)
-	c1 := &pb.Course{OrganizationID: 1}
-	c2 := &pb.Course{OrganizationID: 2}
-	c3 := &pb.Course{OrganizationID: 3}
+	c1 := &pb.Course{OrganizationID: 1, Year: 1}
```
meling, line 180: Suggest to add also the code, like this:
```go
	c1 := &pb.Course{OrganizationID: 1, Code: "DAT101", Year: 1}
	c2 := &pb.Course{OrganizationID: 2, Code: "DAT101", Year: 2}
	c3 := &pb.Course{OrganizationID: 3, Code: "DAT101", Year: 3}
```

database/gormdb_test.go
```diff
@@ -333,10 +333,10 @@ func TestGormDBGetCoursesByUser(t *testing.T) {
 	defer cleanup()
 
 	admin := qtest.CreateFakeUser(t, db, 1)
-	c1 := &pb.Course{OrganizationID: 1}
-	c2 := &pb.Course{OrganizationID: 2}
-	c3 := &pb.Course{OrganizationID: 3}
-	c4 := &pb.Course{OrganizationID: 4}
+	c1 := &pb.Course{OrganizationID: 1, Year: 1}
```
meling, line 336:
```go
	c1 := &pb.Course{OrganizationID: 1, Code: "DAT101", Year: 1}
	c2 := &pb.Course{OrganizationID: 2, Code: "DAT101", Year: 2}
	c3 := &pb.Course{OrganizationID: 3, Code: "DAT101", Year: 3}
	c4 := &pb.Course{OrganizationID: 4, Code: "DAT101", Year: 4}
```

database/gormdb_test.go
```diff
@@ -379,10 +379,10 @@ func TestGormDBGetCoursesByUser(t *testing.T) {
 	}
 
 	wantCourses := []*pb.Course{
-		{ID: c1.ID, OrganizationID: 1, CourseCreatorID: admin.ID, Provider: "fake", Enrolled: pb.Enrollment_PENDING},
-		{ID: c2.ID, OrganizationID: 2, CourseCreatorID: admin.ID, Provider: "fake", Enrolled: pb.Enrollment_NONE},
-		{ID: c3.ID, OrganizationID: 3, CourseCreatorID: admin.ID, Provider: "fake", Enrolled: pb.Enrollment_STUDENT},
-		{ID: c4.ID, OrganizationID: 4, CourseCreatorID: admin.ID, Provider: "fake", Enrolled: pb.Enrollment_NONE},
+		{ID: c1.ID, OrganizationID: 1, Year: 1, CourseCreatorID: admin.ID, Provider: "fake", Enrolled: pb.Enrollment_PENDING},
```
meling, line 382:
```go
		{ID: c1.ID, OrganizationID: 1, Code: "DAT101", Year: 1, CourseCreatorID: admin.ID, Provider: "fake", Enrolled: pb.Enrollment_PENDING},
		{ID: c2.ID, OrganizationID: 2, Code: "DAT101", Year: 2, CourseCreatorID: admin.ID, Provider: "fake", Enrolled: pb.Enrollment_NONE},
		{ID: c3.ID, OrganizationID: 3, Code: "DAT101", Year: 3, CourseCreatorID: admin.ID, Provider: "fake", Enrolled: pb.Enrollment_STUDENT},
		{ID: c4.ID, OrganizationID: 4, Code: "DAT101", Year: 4, CourseCreatorID: admin.ID, Provider: "fake", Enrolled: pb.Enrollment_NONE},
```

database/gormdb_course.go
```diff
@@ -17,6 +17,9 @@ func (db *GormDB) CreateCourse(courseCreatorID uint64, course *pb.Course) error
 	var courses int64
 	if err := db.conn.Model(&pb.Course{}).Where(&pb.Course{
 		OrganizationID: course.OrganizationID,
+	}).Or(&pb.Course{
```
meling, line 20: We need to document what this does. Suggest to add a comment to the godoc line, e.g., something like:
```go
// CreateCourse creates a new course if user with given ID is admin, enrolls user as course teacher.
// The provided course must have a unique (GitHub) OrganizationID not already associated with existing course.
// Similarly, the course must have a unique course code and year.
```
Please tell me if this is not accurate; feel free to improve it.

database/gormdb_test.go
```diff
@@ -751,6 +751,54 @@ func TestGormDBUpdateCourse(t *testing.T) {
 	}
 }
 
+func TestGormDBCourseUniqueContraint(t *testing.T) {
```
meling, line 754: Please move this test to a new file `gormdb_course_test.go`. (We would like to move more of the tests in `gormdb_test.go` to other files to make it easier to find, instead of one very long file...)

database/gormdb_test.go
```diff
@@ -751,6 +751,54 @@ func TestGormDBUpdateCourse(t *testing.T) {
...
+	// Test that a course with the same code and year cannot be created.
+	db, cleanup := qtest.TestDB(t)
+	defer cleanup()
+	wantCourse := &pb.Course{
```
meling, line 758: This test should probably be expanded a bit to test multiple scenarios.
- That we cannot create two courses with the same `OrganizationID` (maybe it can be done in another separate test, but it is also fine to do it as part of this test; it has to do with uniqueness.)
- That we cannot create two courses with the same `Code`-`Year` -- this was already supported.

database/gormdb_test.go
```diff
@@ -751,6 +751,54 @@ func TestGormDBUpdateCourse(t *testing.T) {
...
+	}
+
+	remoteID := &pb.RemoteIdentity{Provider: wantCourse.Provider, RemoteID: 10, AccessToken: "token"}
+	admin := qtest.CreateUserFromRemoteIdentity(t, db, remoteID)
+
+	if err := db.CreateCourse(admin.ID, wantCourse); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := db.CreateCourse(admin.ID, course); err == nil {
```
meling, line 782: Suggest to move the comment outside to clarify what is expected to happen. (General English language tip: Whenever you see yourself writing _this_, stop and think if you can be more specific.)
And check that we get the expected error type.
```go
// CreateCourse should fail because the unique constraint (course.code, course.year) is violated
if err := db.CreateCourse(admin.ID, course); err != nil && !errors.Is(err, database.ErrCourseExists) {
	t.Fatal(err)
}
```

database/gormdb_test.go
```diff
@@ -751,6 +751,54 @@ func TestGormDBUpdateCourse(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if diff := cmp.Diff(wantCourse, gotCourse, protocmp.Transform()); diff != "" {
+		t.Errorf("course mismatch (-want +got):\n%s", diff)
+	}
+
+	// Now create a course with same code but different year
+	course.Year = 2018
+	if err := db.CreateCourse(admin.ID, course); err != nil {
```
meling, line 796: Suggest to move comment above the actual action instead of inside the if (where it would end up if it didn't succeed):
```go
	// CreateCourse should succeed because the unique constraint (course.code, course.year) is not violated
```

meling (changes requested):
This looks solid. Most a few minor documentation issues and if you could add a test to also prove that we cannot create two courses with the same `OrganizationID`, with otherwise "legal"/different values for the code and year. (Not sure if we already have such a test though, so could be useful to check that first...)

database/gormdb_test.go
JosteinLindhom: I moved this test, along with all course related tests I found to a new file `gormdb_course_test.go`.

meling (approved):

PR #620, opened 2022-03-10
Harmonize Rebuild RPCs

JosteinLindhom:
This PR harmonizes the `RebuildSubmission` and `RebuildSubmissions` rpc functions into one, `RebuildSubmissions`.

Due to the function now returning `Void`, it has become harder to dynamically update the frontend upon successful rebuild. 
Implementation of submission streaming can solve this for the redesigned frontend. 
In the old frontend the user has to click the rebuilt submission again after a successful rebuild to view the updated submission..

Alternatively, the ability to fetch a single submission would be sufficient. Fetching a single submission could be added to the `GetSubmissions` function, in a manner not unlike the harmonization of `RebuildSubmissions`.

- Updated old frontend to support harmonized rebuild rpc
- Updated redesigned frontend to support harmonized rebuild rpc
- Removed RebuildSubmission and modified rebuild message to use oneof
- Added helper method and updated validation method
- Combined RebuildSubmissions & RebuildSubmission
- Updated rebuild tests


JosteinLindhom:
> Seems to not compile @JosteinLindhom

Seems like I did not push recompiled proto files. 😬

public/src/components/teacher/Results.tsx
```diff
@@ -171,18 +171,20 @@ export class Results extends React.Component<IResultsProps, IResultsState> {
...
 
     private async rebuildSubmission(): Promise<boolean> {
-        const currentSubmission = this.state.selectedSubmission;
+        const currentSubmission = this.state.selectedSubmission
         if (currentSubmission && currentSubmission.submission) {
-            const ans = await this.props.onSubmissionRebuild(currentSubmission.assignment.getId(), currentSubmission.submission.id);
+            const ans = await this.props.onSubmissionRebuild(currentSubmission.assignment.getId(), currentSubmission.submission.id)
             if (ans) {
-                currentSubmission.submission = ans;
+                // TODO: This no longer works because the updated submission
```
meling, line 178: Couldn't this be fixed by simply fetching the submission using `GetSubmission`? That would at least keep the functionality as it was.

dev/src/overmind/actions.tsx
```diff
@@ -454,11 +454,13 @@ export const setActiveAssignment = ({ state }: Context, assignmentID: number): v
...
 
 /** Rebuilds the currently active submission */
-export const rebuildSubmission = async ({ state, effects }: Context): Promise<void> => {
+export const rebuildSubmission = async ({ state, actions, effects }: Context): Promise<void> => {
     if (state.currentSubmission && state.selectedAssignment) {
         const response = await effects.grpcMan.rebuildSubmission(state.selectedAssignment.getId(), state.activeSubmission)
-        if (response.data) {
-            state.currentSubmission = response.data
+        if (success(response)) {
+            // TODO: Alerting is temporary due to the fact that the server no longer returns the updated submission.
```
meling, line 461: Couldn't this be fixed by simply fetching the submission using `GetSubmission` once the gRPC rebuild call returned successfully? That would at least preserve the functionality as it was.

web/autograder_service.go
```diff
@@ -605,37 +605,39 @@ func (s *AutograderService) UpdateSubmission(ctx context.Context, in *pb.UpdateS
...
-		return nil, ErrInvalidUserInfo
-	}
-	if !s.isTeacher(usr.ID, in.GetCourseID()) {
-		s.logger.Error("RebuildSubmissions failed: user is not teacher")
-		return nil, status.Error(codes.PermissionDenied, "only teachers can rebuild all submissions")
-	}
-	if err := s.rebuildSubmissions(in); err != nil {
-		s.logger.Errorf("RebuildSubmissions failed: %v", err)
-		return nil, status.Error(codes.InvalidArgument, "failed to rebuild submissions")
+// TODO: Access policy: any for SubmissionID type ?
```
meling, line 611: How about this?:
```go
// Access policy: Teacher of CourseID or student / group owner of SubmissionID.
```

dev/src/overmind/actions.tsx
JosteinLindhom: We actually don't have a gRPC function to fetch a single submission currently. I could add a gRPC function for this.

meling: Okay. Hmmm. Can we use `GetSubmissions`? We would need `CourseID` and one of `UserID` or `GroupID` at the call site. Another option could be to actually return a `Submission` from the `RebuildSubmissions` method instead of `Void`. If we are rebuilding all we could return an empty `Submission`. I prefer the first, since it would be a bit strange to return a single submission...

web/autograder_service.go
```diff
@@ -605,37 +605,39 @@ func (s *AutograderService) UpdateSubmission(ctx context.Context, in *pb.UpdateS
...
+func (s *AutograderService) RebuildSubmissions(ctx context.Context, in *pb.RebuildRequest) (*pb.Void, error) {
+	// RebuildType can be either SubmissionID or CourseID, but not both.
+	switch in.GetRebuildType().(type) {
+	case *pb.RebuildRequest_SubmissionID:
+		if !s.isValidSubmission(in.GetSubmissionID()) {
+			s.logger.Errorf("ApproveSubmission failed: submitter has no access to the course")
```
meling, line 617: `RebuildSubmission failed: ...`

web/autograder_service.go
```diff
@@ -605,37 +605,39 @@ func (s *AutograderService) UpdateSubmission(ctx context.Context, in *pb.UpdateS
...
+			return nil, status.Error(codes.PermissionDenied, "submitter has no course access")
+		}
+		_, err := s.rebuildSubmission(in)
+		if err != nil {
+			s.logger.Errorf("RebuildSubmission failed: %v", err)
+			return nil, status.Error(codes.InvalidArgument, "failed to rebuild submission")
+		}
+	case *pb.RebuildRequest_CourseID:
+		usr, err := s.getCurrentUser(ctx)
```
meling, line 626: Should this:
```go
		usr, err := s.getCurrentUser(ctx)
		if err != nil {
			s.logger.Errorf("RebuildSubmissions failed: authentication error: %v", err)
		}
```
be moved to the top of the method? 

It doesn't look likt the `isValidSubmission` checks that actual user is a valid user.

dev/src/overmind/actions.tsx
meling: Maybe it is best to leave this as is for now, and implement #625 separately to avoid that this PR becomes too big. Can you create an issue for these TODO items, so we don't forget them.

web/autograder_service.go
meling: Revising my own comment. Based on what we wrote for `GetSubmissions`, we should say something like this:
```go
// Access policy:
// Current User if Owner of submission,
// Current User if member of group for group submission,
// Teacher of CourseID.
```

JosteinLindhom: Note that neither the old or redesigned frontend has implemented student access for rebuilding submissions.

Currently the `s.isValidSubmission` checks:
- If the submission belongs to an enrollment with status `Student` or higher
- If the submission belongs to a group that belongs to the course.

However, the submission checked is retrieved from the database based on the received submission ID in the received request.

With this in mind, a student could in theory trigger a gRPC call to `RebuildSubmissions` with a request containing an arbitrary submission ID. As long as the conditions listen above are fulfilled, this could trigger a rebuild of a submission not belonging to the student.

Either we:
- Allow students to rebuild their own submissions, requiring minor modifications to `isValidSubmission`
    - Minor modifications to the frontend is required as well
- Disallow students from rebuilding their own submissions. Done by simply moving the `isTeacher` check to the top.

meling: Great. You are paying attention. We should disallow students to rebuild since this feature was available before and was misused when or our tests had bugs etc.

PR #622, opened 2022-03-11
Clean up make submission results functions used by GetSubmissionsByCourse

### LINKED ISSUE #621: Clean up makeResults functions
In web/courses.go there are three select operations to fetch submissions / results:
- all
- group
- individual

There are some common pattern for these three methods that warrant extracting out a private function that can be used by all three and make each of them more clean and readable. These functions are also not tested sufficiently via the exported functions; should also add a test for a more complete view of these three functions.

meling:
- Trivial makeResults simplification; no need to append
- Use pb.SubmissionsForCourseRequest_ALL instead of default
- Make Assignment.CloneWithoutSubmissions robust
- Added ByUser and ByGroup predicate methods to Submission
- Typo
- Added TestGetSubmissionsByCourse
- Disabled output from TestDeleteGroup
- Refactor make{All|Individual|Group}Results functions

Fixes #621

### DIFF
--- a/ag/assignment.go
+++ b/ag/assignment.go
@@ -3,6 +3,8 @@ package ag
 import (
 	context "context"
 	"time"
+
+	"google.golang.org/protobuf/proto"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


meling:
@JosteinLindhom If you could test this with your database set up, to check that we get ALL, GROUP, and INDIVIDUAL submission results correctly over to the frontend.

Issue #617, opened 2022-03-09
Combine UpdateEnrollment and UpdateEnrollments


meling:
To reduce duplication we should consider to combine these two methods into one.

PR #624, opened 2022-03-17
Combined UpdateEnrollments & UpdateEnrollment

JosteinLindhom:
This PR combines the functions UpdateEnrollment and UpdateEnrollments (#617)

The new proto message, UpdateEnrollmentsRequest, contains only a single `oneof`, with possible values `uint64 courseID` or `Enrollment`

I'm not sure using `oneof` is optimal here, or whether always supplying courseID, and having Enrollment be optional is a less verbose solution.

- Combined UpdateEnrollment and UpdateEnrollments
- Updated tests using UpdateEnrollment
- Recompiled proto files


meling:
> I'm not sure using `oneof` is optimal here, or whether always supplying courseID, and having Enrollment be optional is a less verbose solution.

I agree that it looks like it would be better to always include `courseID` because the backend needs it anyway.

meling:
@JosteinLindhom Having looked into this code a bit more, I think perhaps a better option would be to replace the `UpdateEnrollmentsRequest` message type with `Enrollments` (plural) which becomes a wrapper around `[]*pb.Enrollments`. The reason is that we only use `actions.updateEnrollments(courseID)` in one place (`Members.tsx`). Here we also have access to `pending` which is of type `Enrollment[]`, which should trivially convert to `Enrollments` to be passed to the actual gRPC call. 

Similarly, the calls to the single valued `updateEnrollments(enrollment)` can also trivially be converted to a single element `Enrollments` message.

This way, we only have one method and we don't need special handling for a course-based accept all pending.

The only potential issue is performance in that we need to submit more data to accept all pending students; however, this operation is only performed once per student per course, so I doubt it will be a problem. Perhaps we need to zero out some of the fields to make them take less space over the wire, but again don't think it will be necessary unless we observe long delays.

PS: I considered some other options as well, but I think the above is the more reasonable design choice.

ag/validation.go
```diff
@@ -237,3 +237,14 @@ func (c *GradingCriterion) IsValid() bool {
 func (r *CourseUserRequest) IsValid() bool {
 	return r.CourseCode != "" && r.UserLogin != "" && r.CourseYear > 2019
 }
+
+func (m *Enrollments) IsValid() bool {
+	isValid := true
+	for _, e := range m.Enrollments {
+		if !e.IsValid() {
+			isValid = false
+			break
```
meling, line 246: No need to set `isValid = false` and `break` here. You can simply `return false`.

Actually, you don't even need the `isValid` variable because if the for-loop exits without having returned, `isValid` will be true, and thus you only need to `return m.HasCourseID()`.

ag/enrollment.go
```diff
@@ -86,3 +86,25 @@ func (m *Enrollment) IsTeacher() bool {
...
+func (m *Enrollments) GetCourseID() uint64 {
+	if len(m.GetEnrollments()) == 0 {
+		return 0
+	}
+	return m.Enrollments[0].GetCourseID()
+}
+
+// HasCourseID checks all enrollments have the same Course ID
+func (m *Enrollments) HasCourseID() bool {
+	if len(m.GetEnrollments()) == 0 {
```
meling, line 100: I like to be on the safe side and call `m.GetEnrollments()` only once instead of three times (I'm not sure if the compiler is always able to optimize these calls...). Same applies to `GetCourseID()` above.

That is: 
```go
enrollments := m.GetEnrollments()
if len(enrollments) == 0 {
  return false
}
courseID := enrollments[0].GetCourseID()
for _, e := enrollments {
  ...
```

web/autograder_service.go
```diff
@@ -241,58 +241,37 @@ func (s *AutograderService) CreateEnrollment(_ context.Context, in *pb.Enrollmen
...
 	}
-	err = s.updateEnrollments(ctx, scm, in.GetCourseID())
-	if err != nil {
-		s.logger.Errorf("UpdateEnrollments failed: %v", err)
-		if contextCanceled(ctx) {
-			return nil, status.Error(codes.FailedPrecondition, ErrContextCanceled)
+
+	for _, enrollment := range in.GetEnrollments() {
+		if s.isCourseCreator(enrollment.CourseID, enrollment.UserID) {
+			s.logger.Errorf("UpdateEnrollment failed: user %s attempted to demote course creator", user.GetName())
```
meling, line 260: `UpdateEnrollments` (missing s)

web/autograder_service.go
```diff
@@ -241,58 +241,37 @@ func (s *AutograderService) CreateEnrollment(_ context.Context, in *pb.Enrollmen
...
+			return nil, status.Error(codes.PermissionDenied, "course creator cannot be demoted")
 		}
-		if ok, parsedErr := parseSCMError(err); ok {
-			return nil, parsedErr
+
+		err = s.updateEnrollment(ctx, scm, user.GetLogin(), enrollment)
```
meling, line 264: This line can be moved into the if-statement since it only returns an `err`

web/courses.go
```diff
@@ -98,7 +98,7 @@ func (s *AutograderService) updateEnrollment(ctx context.Context, sc scm.SCM, cu
 	case pb.Enrollment_NONE:
 		return s.rejectEnrollment(ctx, sc, enrollment)
 
-	case pb.Enrollment_STUDENT:
+	case pb.Enrollment_STUDENT, pb.Enrollment_PENDING:
```
meling, line 101: Are you sure this is correct behavior? If we receive a request from the frontend (or other client) with status pending, then we should enroll student? I know that we set the student to pending when we do `CreateEnrollment`, and that we read back the enrollment on Line 88. 

I guess my question is: will it happen that the frontend generates `PENDING` enrollment request, when we want to update it to student? I could imagine that we should instead just ignore such requests, like this:

```go
	case pb.Enrollment_PENDING:
		return nil
```

web/courses_test.go
```diff
@@ -484,67 +504,38 @@ func TestPromoteDemoteRejectTeacher(t *testing.T) {
...
-		CourseID: course.ID,
-		Status:   pb.Enrollment_NONE,
-	}); err == nil {
+	teacher_enrollment.Status = pb.Enrollment_NONE
+	if _, err := ags.UpdateEnrollments(ctx, request); err == nil {
 		t.Error("expected error: 'only course creator can change status of other teachers'")
 	}
 
 	// teacher demotes student1, must succeed
+	student1_enrollment.Status = pb.Enrollment_STUDENT
```
meling, line 535: In Go we don't like underscore ;-) `student1Enrollment` and `teacherEnrollment` should do it...

web/courses_test.go
```diff
@@ -467,13 +467,33 @@ func TestPromoteDemoteRejectTeacher(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// student1 attempts to promote student2 to teacher, must fail
-	ctx := qtest.WithUserContext(context.Background(), student1)
-	if _, err := ags.UpdateEnrollment(ctx, &pb.Enrollment{
+	student1_enrollment := &pb.Enrollment{
```
meling, line 470: Avoid underscore in Go.

web/courses.go
meling: I've revised and pushed code that hopefully makes the logic more sensible and a bit more decoupled.

PR #628, opened 2022-03-27
Quickfeed Security: Redesigned Authentication and Authorization Architecture (Master's Thesis 2022)

0xf8f8ff:
A work in progress.


ag/ag.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.27.1
-// 	protoc        v3.19.4
+// 	protoc        v3.17.3
```
meling, line 4: Please update your protobuf installation to avoid potential issues due to older version.

ag/ag.proto
```diff
@@ -65,22 +65,23 @@ message Groups {
...
-    uint64 organizationID          = 8;
-    string organizationPath        = 9;  // The organization's SCM name, e.g. uis-dat520-2020.
-    uint32 slipDays                = 10;
-    string dockerfile              = 11;
-    Enrollment.UserStatus enrolled = 12 [(go.field) = { tags: 'gorm:"-"' }];
-
-    repeated Enrollment enrollments = 13;
-    repeated Assignment assignments = 14;
-    repeated Group groups           = 15;
+    uint64 ID = 1;
```
meling, line 68: Could you install `clang-format` in `/usr/bin`? (see the `.vscode/settings.json`) This would allow you to format the `ag.proto` the same as me, avoiding that the field numbers gets reflowed...

public/package.json
```diff
@@ -14,7 +14,7 @@
     "@types/google-protobuf": "^3.7.4",
     "@types/react": "^16.9.56",
     "@types/react-dom": "^16.9.9",
-    "google-protobuf": "^3.14.0",
+    "google-protobuf": "^3.20.0-rc.2",
     "grpc-web": "^1.3.0",
```
meling, line 18: We now use 1.3.1 in master.

web/config/app.go
```diff
@@ -0,0 +1,98 @@
+package config
```
meling, line 1: Feels like this file belongs in the `scm` package as `github_app.go`

web/config/app.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// AppClient creates client for the Quickfeed GitHub Application
+// This client can only access the metadata of the Application itself
+// To access organizations via GitHub API we need to derive an installation client
+// from this Application client for each course organization
+func NewApp() (*GithubApp, error) {
+	config := NewAppConfig()
+	if !config.Valid() {
+		return nil, fmt.Errorf("Error configuring GitHub App: %+v", config)
```
meling, line 50: Error strings should not be capitalized. Check also others: Line 54, 58 and so on...

web/config/app.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+	if installationID == 0 {
+		return nil, fmt.Errorf("Installation not found for organization %s", courseOrg)
+	}
+	install, err := ghApp.App.InstallationConfig(strconv.Itoa(int(installationID)))
+	if err != nil {
+		return nil, fmt.Errorf("Error configuring github client for installation: %s", err)
+	}
+	return gh.NewClient(install.Client(ctx)), err
```
meling, line 97: err -> nil

web/config/app.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// Creates a new scm client with access to the course organization
+func (ghApp *GithubApp) NewInstallationClient(ctx context.Context, courseOrg string) (*gh.Client, error) {
+	resp, err := ghApp.App.Client().Get(InstallationsAPI)
+	if err != nil {
+		return nil, fmt.Errorf("Cannot get installations for App %s: %s", ghApp.Config.AppID, err)
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body) // response body is []byte
```
meling, line 73: Since Go 1.16, we should use `io.ReadAll()`. Do we need the comment? Feels like this is well-known and easy to uncover if unsure.

web/config/app.go
```diff
@@ -0,0 +1,98 @@
...
+// AppClient creates client for the Quickfeed GitHub Application
+// This client can only access the metadata of the Application itself
+// To access organizations via GitHub API we need to derive an installation client
+// from this Application client for each course organization
+func NewApp() (*GithubApp, error) {
+	config := NewAppConfig()
+	if !config.Valid() {
+		return nil, fmt.Errorf("Error configuring GitHub App: %+v", config)
+	}
+	appKey, err := key.FromFile(appKeyPath)
```
meling, line 52: Do we need to read the key material from file, or can it be read from ENV var? I see it is currently a pem file and not a token... so perhaps it is necessary.

web/config/app.go
```diff
@@ -0,0 +1,98 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"strconv"
+
+	"github.com/beatlabs/github-auth/app"
+	"github.com/beatlabs/github-auth/key"
+	gh "github.com/google/go-github/v35/github"
```
meling, line 14: Should we upgrade to `github.com/google/go-github/v43/github`?? Not sure it is necessary, and what additional features are supported... didn't check. Perhaps some additional github apps features are supported...

web/config/app.go
meling: Would also be nice to have a `TestGitHubApp` thing in `github_app_test.go`

web/auth/jwt.go
```diff
@@ -0,0 +1,49 @@
...
+	UserID  uint64                              `json:"user_id"`
+	Admin   bool                                `json:"admin"`
+	Courses map[uint64]pb.Enrollment_UserStatus `json:"courses"`
+}
+
+type TokenManager struct {
+	TokensToUpdate []uint64
+}
+
+func (tm *TokenManager) JWTUpdateRequired(claims *Claims) bool {
```
meling, line 23: Would be nice to have proper tests for these methods.

web/autograder_service.go
```diff
@@ -21,18 +22,20 @@ type AutograderService struct {
 	logger *zap.SugaredLogger
 	db     database.Database
 	scms   *auth.Scms
-	bh     BaseHookOptions
+	// bh     BaseHookOptions
```
meling, line 25: Better to remove these lines instead of commenting. Should be easy to see the diff.

web/config/app.go
```diff
@@ -0,0 +1,98 @@
...
+	if err := json.Unmarshal(body, &installations); err != nil {
+		return nil, fmt.Errorf("Error unmarshalling installation response: %s", err)
+	}
+	var installationID int64
+	for _, inst := range installations {
+		log.Println("Checking installation ", *inst.Account.Login)
+		if *inst.Account.Login == courseOrg {
+			log.Println("Found installation for ", courseOrg)
+			log.Println("Installation ID is ", *inst.ID)
+			installationID = *inst.ID
```
meling, line 87: Perhaps you could `break` out of the for-loop when found.

web/config/app.go
```diff
@@ -0,0 +1,98 @@
...
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body) // response body is []byte
+	if err != nil {
+		return nil, fmt.Errorf("Error reading installation response: %s", err)
+	}
+	var installations []*gh.Installation
+	if err := json.Unmarshal(body, &installations); err != nil {
+		return nil, fmt.Errorf("Error unmarshalling installation response: %s", err)
+	}
+	var installationID int64
```
meling, line 81: Perhaps this code could be extracted into a function: 

```go
func findApp(courseOrg, installations) (id string, err error)
```
The `id` can then be used directly in `InstallationConfig()` without `strconv.Itoa()`

web/config/config.go
```diff
@@ -0,0 +1,92 @@
+package config
+
+import (
+	"os"
+
+	"github.com/autograde/quickfeed/internal/rand"
+)
+
+const (
```
meling, line 9: Maybe these `consts` could be organized into separate `const` sections or in separate files.

meling (commented):
I've commented on some of the most obvious things.

web/auth/jwt.go
```diff
@@ -18,22 +17,22 @@ type Claims struct {
...
 
 // TODO: probably most of these methods (and struct fields) can be changed to unexported
 
 // JWTUpdateRequired returns element index if the user ID is in the list,
 // otherwise returns -1 to indicate that JWT update is not needed for this user
-func (tm *TokenManager) JWTUpdateRequired(claims *Claims) uint64 {
-	index := -1
-	for i, tokenID := range tm.TokensToUpdate {
-		if claims.UserID == tokenID {
+func (tm *TokenManager) JWTUpdateRequired(claims *Claims) int {
```
meling, line 27: I would prefer to avoid the `JWT` prefix as well as unexporting them (if not needed outside the package).

Also, I think this method should return a `bool` instead since the caller can easily find the claims's `UserID` via `claims.UserID`.

web/auth/jwt.go
```diff
@@ -56,39 +55,35 @@ func (tm *TokenManager) JWTUpdateClaims(db database.Database, userID uint64) (*C
...
 // Update will remove a user ID at the index from the list
-func (tm *TokenManager) Remove(index uint64) {
+func (tm *TokenManager) Remove(index uint64, db database.Database) {
 	updatedTokenList := append(tm.TokensToUpdate[:index], tm.TokensToUpdate[index+1:]...)
 	tm.TokensToUpdate = updatedTokenList
 }
 
 // Add adds a new UserID to the list of users who require an updated JWT token
-func (tm *TokenManager) Add(userID uint64) {
+func (tm *TokenManager) Add(userID uint64, logger *zap.SugaredLogger, db database.Database) {
```
meling, line 66: Maybe the `TokenManager` struct should have a copy of the `db` and `logger` so that you don't need to pass these to all methods.

web/auth/jwt.go
```diff
@@ -56,39 +55,35 @@ func (tm *TokenManager) JWTUpdateClaims(db database.Database, userID uint64) (*C
...
 	}
-	tm.TokensToUpdate = append(tm.TokensToUpdate, userID)
-}
-
-// String creates a string of all IDs in the list to store in the database
-func (tm *TokenManager) String() string {
-	return strings.Trim(strings.Join(strings.Split(fmt.Sprint(tm.TokensToUpdate), " "), ","), "[]")
+	tokenQuery := &pb.UpdateTokenRecord{UserID: userID}
+	if err := db.CreateTokenRecord(tokenQuery); err != nil {
+		logger.Errorf("error updating token record in the database: %w", err)
```
meling, line 75: When logging errors in methods/functions we must do:
- either return after logging the error or
- add a comment explaining why we don't return
- another option is always to return the `err`

Also:
- The capitalization rule does not apply when logging. 
- The message could be: `Failed to update JWT token in database: %v`
- Recall that `%w` does not apply to printing/logging, only if you use `fmt.Errorf("... : %w)` to construct a new error with added context that you return to the caller.

web/auth/jwt.go
meling: Note that I've added a helper in `log/logger.go`, mainly to be used from tests like so:
```go
logger := log.Zap(false)
sugaredLogger := log.Zap(false).Sugar()
```
The `verbose` flag turns logging on/off. For tests we typically only want to enable logging when we get test failures to better understand why there is a problem...

web/auth/jwt.go
```diff
@@ -56,39 +55,35 @@ func (tm *TokenManager) JWTUpdateClaims(db database.Database, userID uint64) (*C
...
+	}
+	// TODO(vera): remove, testing if ID is getting updated, must be tested in the corresponding test
+	logger.Debugf("Token query after update in the database: %+v", tokenQuery)
+	tm.TokensToUpdate = append(tm.TokensToUpdate, tokenQuery)
 }
 
 // UpdateFromDatabase splits string of user IDs stored in the database into a list
```
meling, line 82: Comment not up-to-date anymore

web/auth/jwt.go
```diff
@@ -56,39 +55,35 @@ func (tm *TokenManager) JWTUpdateClaims(db database.Database, userID uint64) (*C
...
-		var intID int
-		intID, err := strconv.Atoi(id)
-		if err != nil {
-			fmt.Printf("Error updating token manager from database: failed converting ID %s to int: %s", id, err)
-		}
-		integerIDs[i] = uint64(intID)
+func (tm *TokenManager) UpdateFromDatabase(logger *zap.SugaredLogger, db database.Database) {
+	tokens, err := db.GetTokenRecords()
+	if err != nil {
+		logger.Errorf("cannot fetch token to update from the database: %w", err)
```
meling, line 86: Should probably return after logging here; or return the error for logging at the call site.

Log message could be: `Failed to update JWT tokens from database: %v`

web/auth/jwt_test.go
```diff
@@ -1,6 +1,67 @@
 package auth_test
 
-import "testing"
+import (
+	"testing"
 
-func TestUpdateJWT(t *testing.T) {
+	"github.com/autograde/quickfeed/ag"
```
meling, line 6: Imported twice causing the linter to complain. For now I prefer to use `pb` since that's what we use in other places.

In the future, I think I want to move the `ag` package to a new `proto` package, which seems to be a common place for proto files... but not sure... I would require a pretty big PR that touches many packages...

web/auth/jwt_test.go
```diff
@@ -1,6 +1,67 @@
...
+		Courses: make(map[uint64]pb.Enrollment_UserStatus, 0),
+	}
+
+	idNoUpdate := manager.JWTUpdateRequired(&claimsNoUpdate)
+	if idNoUpdate != -1 {
+		t.Errorf("expected index -1 (update not required), got %d", idNoUpdate)
+	}
+	idToUpdate := manager.JWTUpdateRequired(&claimsToUpdate)
+	if idToUpdate != int(claimsToUpdate.UserID) {
+		t.Errorf("expected user ID %d, got %d", claimsToUpdate.UserID, idToUpdate)
```
meling, line 46: Please adopt the standard approach for [reporting test failures](https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures).

ag/ag.proto
```diff
@@ -297,6 +298,14 @@ message Organizations {
     repeated Organization organizations = 1;
 }
 
+// TODO(vera): rename this message to smth more descriptive
+// UpdateTokenRecord stores ID of a user whose token requires an update
+// next time the user logs in
+message UpdateTokenRecord {
```
meling, line 304: Could we instead add a field to the `User` type to indicate whether or not a user's JWT token is valid?

We have database methods that can return a single user or a slice of users:
- `GetUser(userID uint64) *pb.User`
- `GetUsers(userIDs ...uint64) []*pb.User`

database/database.go
```diff
@@ -128,6 +129,13 @@ type Database interface {
 	// DeleteRepository deletes repository for the given remote provider's ID.
 	DeleteRepository(remoteID uint64) error
 
+	// GetTokenRecord returns all current update token records.
```
meling, line 132: Consider if these are necessary if we could avoid the whole `UpdateTokenRecord` type by adding a field to the database?

ag/ag.pb.go
0xf8f8ff: Done. The latest version is 3.20.0, should we move to it instead?

web/config/app.go
0xf8f8ff: Fixed.

web/config/app.go
0xf8f8ff: Removed.

web/config/app.go
0xf8f8ff: Upgraded to v43.

web/auth/jwt.go
0xf8f8ff: Tests added.

web/config/config.go
0xf8f8ff: Moved to a separate file.

web/auth/jwt.go
0xf8f8ff: Done.

web/config/app.go
0xf8f8ff: GitHub wants a pem file as an app key, but I think that a path to the file should be read from ENV.

ag/ag.proto
0xf8f8ff: Agree that this is a better solution. Updated.

web/auth/jwt_test.go
0xf8f8ff: Yeah, double importing was an accident. Fixed.

web/auth/jwt.go
0xf8f8ff: Fixed.

web/config/app.go
0xf8f8ff: Moved, added test

go.mod
```diff
@@ -6,42 +6,47 @@ require (
 	github.com/360EntSecGroup-Skylar/excelize v1.4.1
 	github.com/alta/protopatch v0.5.0
 	github.com/autograde/quickfeed/kit v0.6.0
-	github.com/docker/docker v20.10.12+incompatible
+	github.com/beatlabs/github-auth v0.0.0-20220228165532-228567f612ea
+	github.com/docker/docker v20.10.14+incompatible
+	github.com/golang-jwt/jwt v3.2.2+incompatible
 	github.com/google/go-cmp v0.5.7
 	github.com/google/go-github/v35 v35.3.0
```
meling, line 13: We should make sure to not depend on both versions.

web/hooks/github.go
```diff
@@ -47,11 +47,31 @@ func (wh GitHubWebHook) Handle(w http.ResponseWriter, r *http.Request) {
...
 		wh.handlePush(e)
+	case *github.InstallationEvent:
+		wh.logger.Debug(log.IndentJson(e))
+		wh.handleInstallation(e)
 	default:
 		wh.logger.Debugf("Ignored event type %s", github.WebHookType(r))
 	}
 }
 
+func (wh GitHubWebHook) handleInstallation(payload *github.InstallationEvent) {
```
meling, line 58: Could you add a brief explainer for when this will be called. Looks like there is one app per course. Things like this would be helpful to draw a mental picture for reads of the code.

web/hooks/github.go
```diff
@@ -47,11 +47,31 @@ func (wh GitHubWebHook) Handle(w http.ResponseWriter, r *http.Request) {
...
+	wh.logger.Debugf("Received app installation event with ID %s for organization %s with ID %d",
+		payload.Installation.ID, *payload.Sender.Login, payload.Installation.Account.ID)
+
+	// TODO: check if course for such organization name exists in the database
+	course, err := wh.db.GetCourse(uint64(*payload.Installation.Account.ID), false)
```
meling, line 63: We should probably use the payload getter methods (I think they protect better against nil panics), similar to what we do in `handlePush` below.

meling: Like these:
```go
	payload.GetInstallation().GetID()
	payload.GetInstallation().GetAccount().GetID()
	payload.GetSender().GetLogin()
```

web/hooks/github.go
```diff
@@ -47,11 +47,31 @@ func (wh GitHubWebHook) Handle(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		wh.logger.Errorf("Failed to get course for app installation with organization %s (ID %d)")
+		return
+	}
+	course.InstallationID = uint64(*payload.Installation.ID)
```
meling, line 68: If you make the `InstallationID` be `int64` in the proto file, you would avoid this type casting.

main.go
```diff
@@ -75,17 +74,18 @@ func main() {
...
+	// 	Secret:  os.Getenv("WEBHOOK_SECRET"),
+	// }
 
 	runner, err := ci.NewDockerCI(logger)
 	if err != nil {
 		log.Fatalf("failed to set up docker client: %v\n", err)
 	}
 	defer runner.Close()
 
+	// TODO: find and replace (if possible) all occasions where this tiken is used
```
meling, line 88: This is used only by the `cmd/approvelist` command and the `helpbot` (separate repo). See `NewQuickFeed()` and `getSubmissions` in `approvelist`.

meling: There is also some commented code in `GetSubmissionsByCourse` that should be removed when this function can be used from `cmd/approvelist` without auth problems.

ag/ag.proto
```diff
@@ -73,14 +74,15 @@ message Course {
...
-    string organizationPath        = 9;  // The organization's SCM name, e.g. uis-dat520-2020.
-    uint32 slipDays                = 10;
-    string dockerfile              = 11;
-    Enrollment.UserStatus enrolled = 12 [(go.field) = { tags: 'gorm:"-"' }];
-
-    repeated Enrollment enrollments = 13;
-    repeated Assignment assignments = 14;
-    repeated Group groups           = 15;
+    string organizationPath        = 9;   // The organization's SCM name, e.g. uis-dat520-2020.
+    uint64 installationID          = 10;  // Each organization has a unique GitHub App installation ID
```
meling, line 78: Suggest to use `int64` instead to avoid casting later. See other comment.

web/hooks/github.go
0xf8f8ff: Actually, we don't need to listen for an installation event when creating a course, I'll remove this method when the alternative is finished.

ag/ag.proto
0xf8f8ff: Done.

ag/ag.proto
```diff
@@ -26,9 +26,10 @@ message User {
     string email     = 5;
     string avatarURL = 6;
     string login     = 7;
+    bool updateToken = 8;
 
-    repeated RemoteIdentity remoteIdentities = 8;
-    repeated Enrollment enrollments          = 9;
+    repeated RemoteIdentity remoteIdentities = 9;
+    repeated ag.Enrollment enrollments       = 10;
```
meling, line 32: Why add `ag.` here?

assignments/tasks_test.go
```diff
@@ -125,7 +125,7 @@ func TestHandleTasks(t *testing.T) {
 	accessToken := scm.GetAccessToken(t)
 
 	logger := log.Zap(false).Sugar()
-	scm, err := scm.NewSCMClient(logger, "github", accessToken)
+	scm, err := scm.NewSCMClient(logger, scm.GetTestClient(t, qfTestOrg), "github", accessToken)
```
meling, line 128: Should this use instead (since this is a `_test.go` file):
```go
qtest.TestSCMClient(ctx, t, qfTestOrg, "github", accessToken)
```

meling: Should `TestSCMClient` accept a `logger` as input or have a way to enable it internally for local logging... during testing

ci/run_tests_test.go
```diff
@@ -54,16 +54,15 @@ func testRunData(qfTestOrg, userName, accessToken, scriptTemplate string) *ci.Ru
 	return runData
 }
 
+// TODO(vera): needs update (GetUserName needs client based on a personal access token)
```
meling, line 57: Is this test fixable?

ci/run_tests_test.go
```diff
@@ -83,22 +82,14 @@ func TestRunTests(t *testing.T) {
 	t.Logf("%+v\n", results)
 }
 
+// TODO(vera): needs update (GetUserName needs client based on a personal access token)
```
meling, line 85: Is this test fixable?

internal/qtest/test_helper.go
```diff
@@ -48,6 +49,12 @@ func TestDB(t *testing.T) (database.Database, func()) {
 	}
 }
 
+// TestConfig creates a configuration for test instance of the AutograderService
+func TestConfig(t *testing.T) *config.Config {
```
meling, line 53: The `t` argument is not used yet. Maybe it will, but as is it is not necessary.

Also the doc comment should say: `... returns a configuration for testing the AutograderService.`

main.go
```diff
@@ -73,19 +75,13 @@ func main() {
...
-		Secret:  os.Getenv("WEBHOOK_SECRET"),
-	}
-
 	runner, err := ci.NewDockerCI(logger)
 	if err != nil {
 		log.Fatalf("failed to set up docker client: %v\n", err)
 	}
 	defer runner.Close()
 
+	// TODO(vera): find and replace (if possible) all occasions where this token is used
```
meling, line 84: We only use the token-based approach with `helpbot` and `approvelist`. It would be great to have the `cmd/approvelist` command line tool be updated as an example of how we would also have to update the `helpbot`.

main.go
```diff
@@ -61,7 +63,7 @@ func main() {
 		dbFile   = flag.String("database.file", "qf.db", "database file")
 		public   = flag.String("http.public", "public", "path to content to serve")
 		httpAddr = flag.String("http.addr", ":8081", "HTTP listen address")
-		grpcAddr = flag.String("grpc.addr", ":9090", "gRPC listen address")
+		// grpcAddr = flag.String("grpc.addr", ":9090", "gRPC listen address")
```
meling, line 66: Remove commented lines if no longer needed.

main.go
```diff
@@ -94,28 +90,86 @@ func main() {
...
 
-	agService := web.NewAutograderService(logger, db, scms, bh, runner)
-	go web.New(agService, *public, *httpAddr)
+	serverConfig := config.NewConfig(*baseURL, *public, *httpAddr)
 
-	lis, err := net.Listen("tcp", *grpcAddr)
+	githubApp, err := scm.NewApp()
 	if err != nil {
-		log.Fatalf("failed to start tcp listener: %v\n", err)
+		log.Fatalf("failed to start GitHub app: %v\n", err)
```
meling, line 97: No need for `\n` when using `Fatalf`

main.go
```diff
@@ -94,28 +90,86 @@ func main() {
...
+		RedirectURL:  serverConfig.Endpoints.CallbackURL,
 	}
-	go func() {
-		if err := httpServer.ListenAndServe(); err != nil {
-			log.Fatal("Unable to start a http server.")
-		}
-	}()
+	tokenManager, err := auth.NewTokenManager(db, config.TokenExpirationTime, serverConfig.Secrets.TokenSecret, *httpAddr)
+	if err != nil {
+		log.Fatalf("failed to make token manager: %v\n", err)
```
meling, line 108: No need for `\n` when using `Fatalf`

main.go
meling: Log msg should be: `Failed to`

main.go
meling: Log msg should be: `Failed to`

main.go
```diff
@@ -94,28 +90,86 @@ func main() {
 		log.Println("Added application token")
 	}
 
-	agService := web.NewAutograderService(logger, db, scms, bh, runner)
-	go web.New(agService, *public, *httpAddr)
+	serverConfig := config.NewConfig(*baseURL, *public, *httpAddr)
```
meling, line 93: Maybe use `cfg` instead of long `serverConfig`

scm/app.go
```diff
@@ -0,0 +1,124 @@
...
+// To access organizations via GitHub API we need to derive an installation client
+// from this Application client for each course organization
+func NewApp() (*GithubApp, error) {
+	config := newAppConfig()
+	if !config.Valid() {
+		return nil, fmt.Errorf("error configuring GitHub App: %+v", config)
+	}
+	appKey, err := key.FromFile(config.keyPath)
+	if err != nil {
+		return nil, fmt.Errorf("error reading key from file: %s", err)
```
meling, line 65: Here we should use `%w` instead of `%s` to wrap the error.

scm/app.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	appClientConfig, err := app.NewConfig(config.appID, appKey)
+	if err != nil {
+		return nil, fmt.Errorf("error creating GitHub application client: %s", err)
```
meling, line 69: Here we should use `%w` instead of `%s` to wrap the error.

scm/app.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return &GithubApp{
+		config: config,
+		App:    appClientConfig,
+		scms:   NewScms(),
+	}, nil
+}
+
+// Creates a new scm client with access to the course organization
```
meling, line 78: Remember to follow Go practice when documenting:
```go
// NewInstallationClient returns a new scm client with access to the course organization.
```

scm/app.go
```diff
@@ -0,0 +1,124 @@
...
+func (ghApp *GithubApp) NewInstallationClient(ctx context.Context, courseOrg string) (*github.Client, error) {
+	resp, err := ghApp.App.Client().Get(InstallationsAPI)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching installations for GitHub app %s: %s", ghApp.config.appID, err)
```
meling, line 82: Here we should use `%w` instead of `%s` to wrap the error. (same for Lines 87, 91 and 105)...

scm/app.go
```diff
@@ -0,0 +1,124 @@
...
+// GetIDs returns app client ID and secret to be used in auth flow
+func (ghApp *GithubApp) GetID() (string, string) {
+	return ghApp.config.clientID, ghApp.config.secret
+}
+
+func (ghApp *GithubApp) GetUserURL() string {
+	return GitHubUserAPI
+}
+
+// TODO(vera): update and move to a file with test helpers
```
meling, line 119: Already done?

scm/app.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil, fmt.Errorf("cannot find GitHub app installation for organization %s", courseOrg)
+	}
+	install, err := ghApp.App.InstallationConfig(strconv.Itoa(int(installationID)))
+	if err != nil {
+		return nil, fmt.Errorf("error configuring github client for installation: %s", err)
+	}
+	return github.NewClient(install.Client(ctx)), nil
+}
+
+// GetIDs returns app client ID and secret to be used in auth flow
```
meling, line 110:
```go
// GetID returns ...
```
Remember that doc comments should also end with a punctuation: `.`

scm/env_helper.go
```diff
@@ -33,3 +36,21 @@ func GetWebHookServer(t *testing.T) string {
 	}
 	return serverURL
 }
+
+// Creates organization based GitHub app client to use in tests. If organization name is empty,
+// uses the default test organization name
+func GetTestClient(t *testing.T, org string) *github.Client {
```
meling, line 42: Test helpers should be in `_test.go` files. You should also call `t.Helper()` on the first line in test helpers.

scm/github.go
```diff
@@ -22,14 +24,21 @@ type GithubSCM struct {
 }
 
 // NewGithubSCMClient returns a new Github client implementing the SCM interface.
-func NewGithubSCMClient(logger *zap.SugaredLogger, token string) *GithubSCM {
-	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
-	client := github.NewClient(oauth2.NewClient(context.Background(), ts))
+func NewGithubSCMClient(logger *zap.SugaredLogger, client *github.Client, token string) (*GithubSCM, error) {
```
meling, line 27: Suggest to split this function into two:
```go
func NewGithubAppClient(logger, client)
func NewGithubTokenClient(logger, token)
```

scm/github.go
```diff
@@ -709,7 +721,7 @@ func (s *GithubSCM) CreateIssue(ctx context.Context, opt *CreateIssueOptions) (*
 	if err != nil {
 		return nil, ErrFailedSCM{
 			Method:   "CreateIssue",
-			Message:  fmt.Sprintf("failed to create Issue %s, make sure all the details are correct ", opt.Title),
+			Message:  fmt.Sprintf("failed to create issue (repository %s, organization %s), make sure all the details are correct ", opt.Repository, opt.Organization),
```
meling, line 724: Should there be a space at the end of this string?

Also: it is not so clear what "all the details" refers to. Could this be clarified?

meling: PS: The log message on 719 should also be fixed... I guess this is from @oleespe's PR.

scm/github_app_test.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/autograde/quickfeed/log"
+	"github.com/autograde/quickfeed/scm"
+)
+
+func TestGitHubApp(t *testing.T) {
+	testOrg := scm.GetTestOrganization(t)
```
meling, line 12: Should call `t.Helper()`

scm/scm.go
```diff
@@ -77,10 +78,10 @@ type SCM interface {
 }
 
 // NewSCMClient returns a new provider client implementing the SCM interface.
-func NewSCMClient(logger *zap.SugaredLogger, provider, token string) (SCM, error) {
+func NewSCMClient(logger *zap.SugaredLogger, client *github.Client, provider, token string) (SCM, error) {
```
meling, line 81: We should avoid passing a GitHub-specific argument to this constructor. Instead we should create the `github.Client` inside this function or in the `NewGithubSCMClient` function. I think this would actually simplify things, making `AdminService` and `AutograderService` contain just the `SCM` interface instead of `scms.GithubApp` concrete type.

web/access_control.go
```diff
@@ -131,15 +118,17 @@ func (s *AutograderService) isCourseCreator(courseID, userID uint64) bool {
 	return course.GetCourseCreatorID() == userID
 }
 
+// TODO(vera): these two methods can be repurposed to return course-related scm client for most methods
```
meling, line 121: We should probably separate this into two methods. One for course-specific scm and one for personal access token-based scm for accepting invitations. That way, we could probably returns a more narrow interface for the access token scm with only the relevant methods. That is, we should probably define a new interface for it.

Checked: This method may no longer be necessary, since the current user is only used for access control in `UpdateCourse`.

web/access_control.go
```diff
@@ -150,11 +139,7 @@ func (s *AutograderService) getUserAndSCM(ctx context.Context, provider string)
 // All errors are logged, but only a single error is returned to the client.
 // This is a helper method to facilitate consistent treatment of errors and logging.
 func (s *AutograderService) getUserAndSCMForCourse(ctx context.Context, courseID uint64) (*pb.User, scm.SCM, error) {
```
meling, line 141: Turns out this is now just a forwarding to `getUserAndSCM`, which is used in:
- UpdateEnrollments
- UpdateGroup
- DeleteGroup
- IsEmptyRepo

To check if user is teacher, which should also be done in access control before calling these methods.

web/admin_service.go
```diff
@@ -0,0 +1,161 @@
...
+	"go.uber.org/zap"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+type AdminService struct {
+	logger *zap.SugaredLogger
+	db     database.Database
+	app    *scms.GithubApp
```
meling, line 23: Should this be an `scm.Admin` interface that only expose the SCM methods that the `AdminService` needs??

web/admin_service.go
```diff
@@ -0,0 +1,161 @@
...
+func (s *AdminService) getSCM(courseID uint64) (scm.SCM, error) {
+	sc, ok := s.app.GetSCM(courseID)
+	if ok {
+		return sc, nil
+	}
+	return nil, errors.New("no SCM found")
+}
+
+// TODO(vera): this method must work for any service, all services must share same scm list
+// MakeSCMClients creates a new scm client (GitHub app installation based client) for each course in the database.
```
meling, line 132: Should this be moved to the AdminService constructor? Since it should be called at start and shouldn't be called multiple times... if I understood correctly.

web/auth/auth.go
```diff
@@ -178,59 +184,49 @@ func sessionData(session *sessions.Session) string {
...
+func OAuth2Login(logger *zap.SugaredLogger, db database.Database, config oauth2.Config) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != "GET" {
+			logger.Errorf("GitHub login failed: request method %s", r.Method)
+			http.Redirect(w, r, "/", http.StatusUnauthorized)
 		}
+		// TODO(vera): adapt to use with other providers if needed
+		provider := "github"
+
+		// TODO(vera): make sure teacher suffix no longer necessary
```
meling, line 196: That would be fantastic!

web/auth/auth.go
```diff
@@ -178,59 +184,49 @@ func sessionData(session *sessions.Session) string {
...
-			logger.Error(err.Error())
-			return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+func OAuth2Login(logger *zap.SugaredLogger, db database.Database, config oauth2.Config) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != "GET" {
+			logger.Errorf("GitHub login failed: request method %s", r.Method)
+			http.Redirect(w, r, "/", http.StatusUnauthorized)
 		}
+		// TODO(vera): adapt to use with other providers if needed
+		provider := "github"
```
meling, line 194: If teacher suffix below is no longer necessary, then we can also remove `provider` here and instead use details from `config` for logging, which should indicate whether or not it is GitHub.

web/auth/auth.go
```diff
@@ -178,59 +184,49 @@ func sessionData(session *sessions.Session) string {
...
-		r := c.Request()
-
-		provider, err := gothic.GetProviderName(r)
-		if err != nil {
-			logger.Error(err.Error())
-			return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+func OAuth2Login(logger *zap.SugaredLogger, db database.Database, config oauth2.Config) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != "GET" {
+			logger.Errorf("GitHub login failed: request method %s", r.Method)
```
meling, line 190: Maybe we should remove `GitHub` here and instead use details from `config` for logging.

web/auth/auth.go
```diff
@@ -178,59 +184,49 @@ func sessionData(session *sessions.Session) string {
...
-		return c.Redirect(http.StatusTemporaryRedirect, url)
+		logger.Debugf("Redirecting to %s to perform authentication; AuthURL: %v", provider, config.Endpoint.AuthURL)
+		http.Redirect(w, r, config.Endpoint.AuthURL, http.StatusTemporaryRedirect)
 	}
 }
 
 // OAuth2Callback handles the callback from an oauth2 provider.
-func OAuth2Callback(logger *zap.SugaredLogger, db database.Database, scms *Scms) echo.HandlerFunc {
-	return func(c echo.Context) error {
+func OAuth2Callback(logger *zap.SugaredLogger, db database.Database, config oauth2.Config, app *scm.GithubApp, tokens *TokenManager, secret string) http.HandlerFunc {
```
meling, line 217: This method was very long (170 lines). Please break into smaller helper functions.

web/auth/auth.go
```diff
@@ -178,59 +184,49 @@ func sessionData(session *sessions.Session) string {
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != "GET" {
+			logger.Errorf("GitHub login failed: request method %s", r.Method)
+			http.Redirect(w, r, "/", http.StatusUnauthorized)
```
meling: According to the docs for `http.Redirect()`:
*The provided code should be in the 3xx range and is usually StatusMovedPermanently, StatusFound or StatusSeeOther.*

So it is not clear if `http.Redirect()` is the correct approach for 401 `http.StatusUnauthorized`.

Also, when redirecting, does it make sense to continue executing or should the method return.

Lastly, we do a lot of 401 redirects and error logging. I added a helper `unauthorized()` for this... see latest commit.

web/auth/auth.go
```diff
@@ -477,11 +461,11 @@ func extractState(r *http.Request, key string) (redirect string, teacher bool) {
...
 }
 
-func extractSessionCookie(w *echo.Response) string {
+func extractToken(w http.ResponseWriter) string {
 	// Helper function that extracts an outgoing session cookie.
 	outgoingCookies := w.Header().Values(OutgoingCookie)
 	for _, cookie := range outgoingCookies {
-		if c := strings.Split(cookie, "="); c[0] == SessionKey {
+		if c := strings.Split(cookie, "="); c[0] == "auth" {
 			token := strings.Split(cookie, ";")[0]
```
meling, line 475: Can return directly here.

web/auth/auth.go
```diff
@@ -477,11 +461,11 @@ func extractState(r *http.Request, key string) (redirect string, teacher bool) {
 	return url[1:], teacher
 }
 
-func extractSessionCookie(w *echo.Response) string {
+func extractToken(w http.ResponseWriter) string {
 	// Helper function that extracts an outgoing session cookie.
 	outgoingCookies := w.Header().Values(OutgoingCookie)
 	for _, cookie := range outgoingCookies {
-		if c := strings.Split(cookie, "="); c[0] == SessionKey {
+		if c := strings.Split(cookie, "="); c[0] == "auth" {
```
meling, line 474: Go 1.18 has a nice [strings.Cut](https://pkg.go.dev/strings#Cut) function that would be perfect for this.

web/auth/interceptors/access_control.go
```diff
@@ -0,0 +1,134 @@
...
+	pb "github.com/autograde/quickfeed/ag"
+	"github.com/autograde/quickfeed/web/auth"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+// "user" role implies that user attempts to access information about himself.
```
meling, line 16: We should avoid string comparison for the roles; it will be faster and less error prone to use something like this:
```go
type (
	role  int
	roles []role
)

const (
	// user role implies that user attempts to access information about himself.
	user role = iota
	// group role implies that the user is a course student + a member of the given group.
	group
	// student role implies that the user is enrolled in the course with any role.
	student
	// teacher: user enrolled in the course with teacher status.
	teacher
	// owner is the creator of a manual review.
	owner
	// admin is the ...
	admin
)

var access = map[string]roles{
	"UpdateUser":           {user, admin},
	"GetEnrollmentsByUser": {user, admin},
}
```

web/auth/interceptors/access_control.go
```diff
@@ -0,0 +1,134 @@
...
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		// get JWT from context
+		meta, ok := metadata.FromIncomingContext(ctx)
+		if !ok {
+			logger.Sugar().Errorf("Access control failed: missing metadata")
+			return nil, status.Errorf(codes.Unauthenticated, "access denied")
+		}
+		logger.Sugar().Debugf("Request %s has metadata: %+v", info.FullMethod, meta) // tmp
+		// get claims from jwt
+		for _, c := range meta.Get("cookie") {
```
meling, line 65: There is a helper function `auth.extractToken` that does more or less the same thing. Didn't check if we can or should export that function, but we should at least avoid many nested loops and ifs -- it makes the code difficult to read.

web/auth/interceptors/access_control.go
```diff
@@ -0,0 +1,134 @@
...
+					if ok {
+						for _, role := range roles {
+							switch role {
+							// TODO(vera): refactor case handlers
+							case "user":
+								// Methods that can be accessed by the owner of the UserID can also be accessed by admin.
+								// Skip other checks if user has admin role.
+								// TODO(vera): should better let it continue and check the next role
+								// instead of checking for admin here
+								if !claims.Admin {
```
meling, line 91: This check and the subsequent switch on the method looks awkward (special cases). I'm not quite sure I understand this check.

web/auth/interceptors/access_control.go
meling:
```go
					roles, ok := access[method]
					if ok {
						for _, role := range roles {
							switch role {
							case user:
							case group:
							case teacher:
							case admin:
							}
						}
					}
```

web/auth/interceptors/access_control.go
```diff
@@ -0,0 +1,134 @@
...
+			return nil, status.Errorf(codes.Unauthenticated, "access denied")
+		}
+		logger.Sugar().Debugf("Request %s has metadata: %+v", info.FullMethod, meta) // tmp
+		// get claims from jwt
+		for _, c := range meta.Get("cookie") {
+			fields := strings.Fields(c)
+			for _, field := range fields {
+				logger.Sugar().Debugf("metadata field: %s", field) // tmp
+				if strings.Contains(field, "auth") {
+					token := strings.Split(field, "=")[1]
```
meling, line 70: Could use the new `strings.Cut()` from Go 1.18.

web/auth/interceptors/interceptors.go
```diff
@@ -0,0 +1 @@
+package interceptors
```
meling, line 1: Delete?

web/autograder_service.go
```diff
@@ -11,28 +11,30 @@ import (
...
 )
 
 // AutograderService holds references to the database and
 // other shared data structures.
 type AutograderService struct {
 	logger *zap.SugaredLogger
 	db     database.Database
-	scms   *auth.Scms
-	bh     BaseHookOptions
+	app    *scms.GithubApp
```
meling, line 25: Should not be GitHub-specific.

web/auth/interceptors/tokens.go
```diff
@@ -0,0 +1,60 @@
...
+		if !ok {
+			logger.Sugar().Errorf("Token validation failed: missing metadata")
+			return nil, status.Errorf(codes.Unauthenticated, "access denied")
+		}
+		logger.Sugar().Debugf("Validate: Request %s has metadata: %+v", info.FullMethod, meta) // tmp
+		for _, c := range meta.Get("cookie") {
+			fields := strings.Fields(c)
+			for _, field := range fields {
+				logger.Sugar().Debugf("metadata field: %s", field) // tmp
+				if strings.Contains(field, "auth") {
```
meling, line 27: Go 1.18 `strings.Cut()`

web/auth/interceptors/tokens.go
```diff
@@ -0,0 +1,60 @@
...
+
+func ValidateToken(logger *zap.Logger, tokens *auth.TokenManager) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		meta, ok := metadata.FromIncomingContext(ctx)
+		if !ok {
+			logger.Sugar().Errorf("Token validation failed: missing metadata")
+			return nil, status.Errorf(codes.Unauthenticated, "access denied")
+		}
+		logger.Sugar().Debugf("Validate: Request %s has metadata: %+v", info.FullMethod, meta) // tmp
+		for _, c := range meta.Get("cookie") {
```
meling, line 23: Use helper func. See other [comment here.](https://github.com/quickfeed/quickfeed/pull/628/files#r860093460)

web/auth/interceptors/tokens.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/autograde/quickfeed/web/auth"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+func ValidateToken(logger *zap.Logger, tokens *auth.TokenManager) grpc.UnaryServerInterceptor {
```
meling, line 15: Seems like we only use the `SugaredLogger`, so we should pass that in instead. (Same for all inceptors I think)

web/auth/interceptors/token_updater.go
```diff
@@ -0,0 +1,79 @@
...
+	"strings"
+	"time"
+
+	pb "github.com/autograde/quickfeed/ag/types"
+	"github.com/autograde/quickfeed/web/auth"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+)
+
+var methods = []string{
```
meling, line 15: If you make this be a `map[string]struct{}` you can initialize it as:
```go
var methods = map[string]struct{} {
 "UpdateUser": struct{}{},
 "CreateCourse": struct{}{},
...
}
```
Then you don't need to do `sort.Strings(methods)` and `sort.SearchStrings()`... etc

meling:
Closing this as the majority of the functionality has been merged via smaller PRs.

PR #634, opened 2022-04-01
Adds version check for grpc-web

### LINKED ISSUE #629: Add check for correct proto versions
We need to update the `grpc-web` version in these places:
- `dev/package.json`
- `public/package.json`
- `Makefile`

We currently support a workflow that allows to compile the proto files and UI frontend code using:
```sh
% make proto
% make ui
```

As part of this workflow, we should add a check to ensure that all versions are updated before compiling the UI.

Unfortunately, it doesn't appear that `protoc-gen-grpc-web` has a `-v` or `--version` flag that can be checked.

We should also check these versions:
```sh
% protoc --version
libprotoc 3.19.4
% protoc-gen-go --version
protoc-gen-go v1.27.1
% protoc-gen-go-grpc --version
protoc-gen-go-grpc 1.2.0
% protoc-gen-grpc-web --version
protoc-gen-grpc-web: Unknown option: --version
```

meling:
- Added proto_path options to settings.json
- Added grpcweb version check before the ui target
- Added check for latest release version of grpc-web
- Added alternative install instructions for grpc-web

Fixes #629

### DIFF
--- a/.vscode/quickfeed-words.txt
+++ b/.vscode/quickfeed-words.txt
@@ -85,12 +85,13 @@ Operativsystemer
 opsys
 orgs
 outform
-Overmind
 overmind
+Overmind
 paxos
 pgrep
 pkgs
 privkey
+proto
 protobuf
 protoc
 protoimpl
--- a/.vscode/quickfeed-words.txt
+++ b/.vscode/quickfeed-words.txt
@@ -82,6 +82,7 @@ ngrok
 npmtools
 npmwebpack
 NTNU
+oneof
 Operativsystemer
 opsys
 orgs
--- a/.vscode/quickfeed-words.txt
+++ b/.vscode/quickfeed-words.txt
@@ -45,6 +45,7 @@ Grafana
 Grafana's
 groupname
 grpc
+grpcpb
 grpcweb
 Hein
 Hein's
@@ -88,6 +89,7 @@ outform
 overmind
 Overmind
 paxos
+pbgo
 pgrep
 pkgs
 privkey
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,6 +8,12 @@
     "breadcrumbs.enabled": true,
     "protoc": {
         "compile_on_save": false,
+        "options": [
+            "--proto_path=${workspaceRoot}",
+            "--proto_path=${workspaceRoot}/kit",
+            "--proto_path=`go list -m -f {{.Dir}} google.golang.org/protobuf`",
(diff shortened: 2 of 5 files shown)


meling:
@JosteinLindhom @0xf8f8ff @oleespe @Synnems @Padogen Could you all try out this PR on your local machine to see if there are any problems? The main thing to try is to run `make version-check`, and if that works it would be interesting to see if `make proto` works. I think this should work on Windows WSL, but probably not on pure Windows PowerShell, but feel free to let me know if it actually works. It is not a goal to make it work under Windows, but it should work under WSL.

The goal of this PR is to check that the locally installed dev tools you are using is of the correct version.

PS: If you have a newer version of some of these dev tools, then it should not give a warning, but should compile new versions of the `.pb.go` files... essentially forcing others to upgrade.

JosteinLindhom:
Both `make version-check` and `make proto` work on WSL.

In Windows `make version-check` (and ` go run .\cmd\vercheck\main.go`)  fails due to `protoc-gen-go.exe` not having any arguments `--version`, or other variants to show version.

```ps
PS quickfeed> protoc-gen-go --version
protoc-gen-go.exe: unknown argument "--version" (this program should be run by protoc, not directly)
```

I have not been able to find a solution for this issue however.

meling:
Ok. That's interesting. Then either you have an old version installed, or it is from another codebase than [this one](https://github.com/protocolbuffers/protobuf-go/blob/master/cmd/protoc-gen-go/main.go). It looks like you can download the latest windows version from [here](https://github.com/protocolbuffers/protobuf-go/releases/tag/v1.28.0). Although, I typically use `make devtools`, that should do the same.

Note that we haven't upgraded to the latest version; maybe I'll do that as part of this PR.

meling:
Could you try to add this code to the `vercheck` tool on after Line 53:
```go
	if strings.Contains(s, `unknown argument "--version"`) {
		log.Println(s)
		log.Fatalf("Your installed %s version is too old. Please update to the latest version.", tool)
	}
```

meling:
The previous code probably didn't work; @JosteinLindhom could you try with the new version. Feel free to play around with the vercheck code.

JosteinLindhom:
I did indeed have an older version where the `--version` flag was not implemented. Updating to the one you linked works.

Looking for `unknown argument "--version"` does not seem to work, as `sh.Output(tool + "--version")` returns an empty string, with an error `exit status 1`.

meling:
It is often a bit confusing where errors are being sent. I have updated again. @JosteinLindhom does this version work better?

JosteinLindhom:
The lastest commit fixed the issue I had in Windows.

Issue #604, opened 2022-02-07
[Manual reviews]: Release all not working, Approve all is working "sometimes"?


0xf8f8ff:
The manual review feature `Release` > `Release all` seems to do nothing at the moment.

`Approve all` is a mystery. It works perfectly fine when I test it locally, but does not approve everyone when used for DAT310 labs on qf. Can't find any error messages in logs.

PR #637, opened 2022-04-06
Release approve all

JosteinLindhom:
This PR fixes an issue where the `Release all` functionality did nothing (#604).

When a review is updated, the related submission and its score is only updated if the submission has been `released`. This means that unreleased submissions will have a score of 0. 

```go
if submission.Released {
	// Updated review will most probably have a new score. Update the submission score as well
	// for submissions with released review.
	submission.Score = review.Score
	if err := s.db.UpdateSubmission(submission); err != nil {
		return nil, err
	}
```

The release all functionality looks at *submission* scores, and releases all those above the given minimum score limit.

```go
func (db *GormDB) UpdateSubmissions(courseID uint64, query *pb.Submission) error {
        return db.conn.
		Model(query).
		Where("assignment_id = ?", query.AssignmentID).
		Where("score >= ?", query.Score).
		Updates(&pb.Submission{
			Status:   query.Status,
			Released: query.Released,
		}).Error
}
```

The fix I attempt in this PR is to update the submission score if it differs from the computed score of an updated review.

```go
// Update the submission's score if the review score has changed.
if submission.Score != review.Score {
	submission.Score = review.Score
	if err := s.db.UpdateSubmission(submission); err != nil {
		return nil, err
	}
}
```

This introduces another problem: students can see their submission score before the submission being released.
To avoid this, I added a helper method to the `Submissions` type that will remove the `score`, `status`, and `reviews` of an unreleased submission that has `len(reviews) > 0` reviews.

```go
// autograder_service.go GetSubmissions(...)
// If the user is not a teacher, remove score and reviews from submissions that are not released.
if !s.isTeacher(usr.ID, in.CourseID) {
	submissions.Clean()
}
```

```go
// Clean removes any score or reviews from the submission if it is not released.
// This is to prevent users from seeing the score or reviews of a submission that has not been released.
func (s *Submissions) Clean() {
	for _, submission := range s.Submissions {
		// Released submissions, or submissions with no reviews need no cleaning.
		if submission.GetReleased() || len(submission.GetReviews()) == 0 {
			continue
		}
		// Remove any score, status, or reviews if the submission is not released.
		submission.Score = 0
		submission.Status = Submission_NONE
		submission.Reviews = nil
	}
}
```

This way we no longer send (and have to hide) unreleased reviews to students in the frontend, and we do not leak the `score` or `status` of the submission before releasing it.

### DIFF
--- a/ag/submission.go
+++ b/ag/submission.go
@@ -34,3 +34,18 @@ func (s *Submission) ByUser(userID uint64) bool {
 func (s *Submission) ByGroup(groupID uint64) bool {
 	return s.GetUserID() == 0 && s.GetGroupID() > 0 && s.GetGroupID() == groupID
 }
+
+// Clean removes any score or reviews from the submission if it is not released.
+// This is to prevent users from seeing the score or reviews of a submission that has not been released.
+func (s *Submissions) Clean() {
+	for _, submission := range s.Submissions {
+		// Released submissions, or submissions with no reviews need no cleaning.
+		if submission.GetReleased() || len(submission.GetReviews()) == 0 {
+			continue
+		}
+		// Remove any score, status, or reviews if the submission is not released.
+		submission.Score = 0
+		submission.Status = Submission_NONE
+		submission.Reviews = nil
+	}
+}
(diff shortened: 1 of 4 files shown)


0xf8f8ff (approved):
This solves both problems (mass release and scores) perfectly. Thanks!

meling (approved):
Excellent work @JosteinLindhom! I love tests like this one.

PR #638, opened 2022-04-07
Accepting student bug

Padogen:
- tests for members page
- fixed promote/demote and accept bug


Synnems:
Some of the new jest tests will fail with current setup. We commented out `const response = await effects.grpcMan.updateEnrollments([temp])` in updateEnrollment in actions.tsx, and used `if (true)`. This will check if the frontend works but not the backend logic, which will make the tests pass.

dev/src/components/Members.tsx
```diff
@@ -67,7 +68,7 @@ const Members = (): JSX.Element => {
             data.push(edit ? (
                 <div>
                     <i className="badge badge-primary clickable"
-                        onClick={() => actions.updateEnrollment({ enrollment: enrollment, status: isTeacher(enrollment) ? Enrollment.UserStatus.STUDENT : Enrollment.UserStatus.TEACHER })}>
+                        onClick={() => { actions.updateEnrollment({ enrollment: enrollment, status: isTeacher(enrollment) ? Enrollment.UserStatus.STUDENT : Enrollment.UserStatus.TEACHER }); setPromote(!promote) }}>
```
JosteinLindhom, line 71: Currently neither `accept` nor `promote` are used to control which text is displayed. 

Note also that the `accept` and `promote` states are function scoped. `members` is a nested function `.map(enrollment => { ... })`, which returns an array of rows representing individual members in the table. 

The issue that arises if `setPromote(!promote)` is invoked is that it will affect *all* rows. Flipping `promote` from `true -> false` will be reflected in all rows, rather than only the one clicked.

You can avoid this by moving the `approve/reject` and `promote/demote` buttons into two separate components having their own local state.

You will also have to handle the case where updating the enrollment status is unsuccessful.

However, I suggest handling the issue in the `updateEnrollment` action:

```ts
        if (success(response)) {
            // If successful, update enrollment in state with new status
            if (status == Enrollment.UserStatus.NONE) {
                // If the enrollment is rejected, remove it from state
                state.courseEnrollments[state.activeCourse] = state.courseEnrollments[state.activeCourse]?.filter(s => s.getId() != enrollment.getId())
            } else {
                const enrollment = state.courseEnrollments[state.activeCourse]?.find(s => s.getId() == temp.getId())
                if (enrollment) {
                    enrollment.setStatus(status)
                }
            }
        }
```

Line 7-10.

The reason the text does not update is that `state.courseEnrollments[courseID]` is wrapped in `json(...)` in `sortEnrollments(...)`.
The issue with `json(...)` is that is removes a proxy wrapper from our state object, which is used to tell Overmind when to trigger a re-render of components. 

Updating the enrollment passed in to `updateEnrollment` does change the state, but it will not trigger a re-render. Manually triggering a re-render of the `Members` component will update the text to reflect the desired value.
The change on line 7-10 retrieves a new reference to the enrollment, wrapped in a proxy. Modifying the status of this enrollment will automatically trigger a re-render with the correct text.

dev/src/Helpers.ts
```diff
@@ -363,3 +361,237 @@ export const sortEnrollments = (enrollments: Enrollment[], sortBy: EnrollmentSor
...
+        const users = obj.usersList.map(u => this.toUser(u))
+        group.setUsersList(users)
+
+        const enrollments = obj.enrollmentsList.map(e => this.toEnrollment(e))
+        group.setEnrollmentsList(enrollments)
+
+        return group
+    }
+
+    public static load = () => {
```
meling, line 587: This func seems to no longer be used. Will it be needed?

dev/src/Helpers.ts
```diff
@@ -363,3 +361,237 @@ export const sortEnrollments = (enrollments: Enrollment[], sortBy: EnrollmentSor
         return enrollmentCompare(a, b, sortBy, descending)
     })
 }
+
+// Class with converter functions for the different proto types
+// TODO(jostein): Move this to a separate file
```
meling, line 366: Ideally, I would prefer to move it as part of this PR and remove this TODO, if that isn't too much effort.

dev/src/__tests__/updateEnrollment.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+React.useLayoutEffect = React.useEffect
+
+describe("UpdateEnrollment", () => {
+    const mockedOvermind = createOvermindMock(config, {
+        grpcMan: new MockGrpcManager()
+    })
+    it("Pending student gets accecpted", async () => {
+        await mockedOvermind.actions.getEnrollmentsByCourse({ courseID: 2, statuses: [] })
+        //This is a user with status pending
```
meling, line 22: Please use space between comment symbol and text: `// This ...`

dev/src/__tests__/updateEnrollment.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    })
+
+    it("Demote teacher to student", async () => {
+        await mockedOvermind.actions.getEnrollmentsByCourse({ courseID: 2, statuses: [] })
+        //This is a user with status teacher
+        window.confirm = jest.fn(() => true)
+        mockedOvermind.actions.updateEnrollment({ enrollment: mockedOvermind.state.courseEnrollments[2][0], status: Enrollment.UserStatus.STUDENT })
+        expect(mockedOvermind.state.courseEnrollments[2][0].getStatus()).toEqual(2)
+    })
+    it("Promote student to teacher", async () => {
```
meling, line 35: Please use blank line between tests

dev/src/__tests__/updateEnrollment.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+        await mockedOvermind.actions.getEnrollmentsByCourse({ courseID: 1, statuses: [] })
+        //This is a user with status student
+        window.confirm = jest.fn(() => true)
+        mockedOvermind.actions.updateEnrollment({ enrollment: mockedOvermind.state.courseEnrollments[1][0], status: Enrollment.UserStatus.TEACHER })
+        expect(mockedOvermind.state.courseEnrollments[1][0].getStatus()).toEqual(3)
+
```
meling, line 41: No need for blank line here

dev/src/__tests__/updateEnrollment.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+        const wrapped = render(
+            <Provider value={mockedOvermind}>
+                <Router history={history} >
+                    <Route path="/course/:id/members" component={Members} />
+                </Router>
+            </Provider>
+        )
+        expect(wrapped.find("i").first().text()).toEqual("Demote")
+    })
+    it("If status is student, button should display promote", () => {
```
meling, line 70: Blank line

dev/src/__tests__/updateEnrollment.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { MockGrpcManager } from "../MockGRPCManager"
+
+
+React.useLayoutEffect = React.useEffect
+
+describe("UpdateEnrollment", () => {
+    const mockedOvermind = createOvermindMock(config, {
+        grpcMan: new MockGrpcManager()
+    })
+    it("Pending student gets accecpted", async () => {
```
meling, line 20: Typo: `accepted`

dev/src/__tests__/updateEnrollment.test.tsx
meling: I'm not into the frontend code... but I'm struggling to see the `user with status pending` in the code. The same applies to the other comments below. Can you clarify the comment please.

dev/src/__tests__/updateEnrollment.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+        await mockedOvermind.actions.getEnrollmentsByCourse({ courseID: 2, statuses: [] })
+        //This is a user with status pending
+        window.confirm = jest.fn(() => true)
+        mockedOvermind.actions.updateEnrollment({ enrollment: mockedOvermind.state.courseEnrollments[2][1], status: Enrollment.UserStatus.STUDENT })
```
meling, line 24: The code `mockedOvermind.state.courseEnrollments[2][1]` seems to be an enrollment. It is also used on Line 25. Could it be made into a separate variable, e.g.,:
```typescript
enrollment = mockedOvermind.state.courseEnrollments[2][1]
```
That can be used in both Lines 24 and 25.

meling: Similar for the remaining tests below.

dev/src/__tests__/updateEnrollment.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+        await expect(mockedOvermind.state.courseEnrollments[2][1].getStatus()).toEqual(2)
```
meling, line 25: What is `toEqual(2)`?? I believe there is an enum that should be used instead.

dev/src/__tests__/updateEnrollment.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    })
+})
+
+configure({ adapter: new Adapter() })
+
+describe("UpdateEnrollment in webpage", () => {
+    it("If status is teacher, button should display demote", () => {
+        const user = new User().setId(1).setName("Test User").setStudentid("6583969706").setEmail("test@gmail.com")
+        const enrollment = new Enrollment().setId(2).setCourseid(1).setStatus(3).setUser(user).setSlipdaysremaining(3).setLastactivitydate("10 Mar").setTotalapproved(0)
```
meling, line 50: Very long line. Suggest split point:
```typescript
        const enrollment = new Enrollment().setId(2).setCourseid(1).setStatus(3).setUser(user).
            setSlipdaysremaining(3).setLastactivitydate("10 Mar").setTotalapproved(0)
```
Same applies for next test on Line 72 as well.

dev/src/overmind/actions.tsx
```diff
@@ -459,6 +459,7 @@ export const setActiveFavorite = ({ state }: Context, isActive: boolean): void =
     state.showFavorites = isActive
 }
 
+
```
meling, line 462: No need for blank line here

dev/src/Helpers.ts
```diff
@@ -363,3 +361,237 @@ export const sortEnrollments = (enrollments: Enrollment[], sortBy: EnrollmentSor
...
+     */
+    public static create<T>(type: (new () => any)): T {
+        return (new type()).toObject() as T
+    }
+
+    /** The following functions will convert various kinds of objects
+     *  into their respective protobuf message types */
+    // TODO(jostein): Add converting functions for all types
+    // TODO(jostein): It would be awesome to make this generic. Not sure how to do that though.
+    public static toUser = (obj: User.AsObject): User => {
```
meling, line 384: Seems like these conversion funcs are not used yet. Might be worth to think about if we could generate this stuff. The simple fields should be easy; looks a bit more difficult for lists and maps etc.

dev/src/__tests__/updateEnrollment.test.tsx
Padogen: status refers to the userStatus in class Enrollment. 
`  export enum UserStatus { 
    NONE = 0,
    PENDING = 1,
    STUDENT = 2,
    TEACHER = 3,
  }`

JosteinLindhom (approved):

meling:
This seems to overlap with the Selenium PR. Or did I miss something? @JosteinLindhom

JosteinLindhom:
> This seems to overlap with the Selenium PR. Or did I miss something? @JosteinLindhom

Everything in this PR is included in #640. Only difference is `updateEnrollment.test.ts` being is slightly modified in #640.

PR #639, opened 2022-04-19
Pull request support for group repositories

oleespe:
This PR should implement the following features:
- When groups create a PR, QF checks if this PR is linked to an existing task related issue.
- It should assign reviewers to the PR.
- Some form of integration with GitHub workflows.


kit/score/registry.go
```diff
@@ -60,7 +60,11 @@ func (s *registry) PrintTestInfo(sorted ...bool) {
 // Add test with given max score and weight to the registry.
 //
 // Will panic if the test has already been registered or if max or weight is non-positive.
-func (s *registry) Add(test interface{}, taskName string, max, weight int) {
+func (s *registry) Add(test interface{}, max, weight int) {
+	s.add(testName(test), "", max, weight)
+}
+
+func (s *registry) AddWithTaskName(test interface{}, taskName string, max, weight int) {
```
meling, line 67: Add doc comment

kit/score/registry.go
```diff
@@ -69,7 +73,12 @@ func (s *registry) Add(test interface{}, taskName string, max, weight int) {
 // conjunction with MaxByName and MinByName called from within a subtest.
 //
 // Will panic if the test has already been registered or if max or weight is non-positive.
-func (s *registry) AddSub(test interface{}, taskName, subTestName string, max, weight int) {
+func (s *registry) AddSub(test interface{}, subTestName string, max, weight int) {
+	tstName := fmt.Sprintf("%s/%s", testName(test), subTestName)
+	s.add(tstName, "", max, weight)
+}
+
+func (s *registry) AddSubWithTaskName(test interface{}, taskName, subTestName string, max, weight int) {
```
meling, line 81: Add doc comment.

Suggest to swap `taskName` and `subTestName`, since the first argument is the main test "name".

kit/score/results.go
```diff
@@ -124,7 +124,15 @@ func (r *Results) Validate(secret string) error {
 // Sum returns the total score computed over the set of recorded scores.
 // The total is a grade in the range 0-100.
 // This method must only be called after Validate has returned nil.
-func (r *Results) Sum(taskName string) uint32 {
+func (r *Results) Sum() uint32 {
+	return r.sum("")
+}
+
+func (r *Results) TaskSum(taskName string) uint32 {
```
meling, line 131: Add doc comment to exported methods.

ag/course.go
```diff
@@ -23,3 +23,13 @@ func (course *Course) SetSlipDays() {
 		g.SetSlipDays(course)
 	}
 }
+
+func (course *Course) GetAllTeacherEnrollments() []*Enrollment {
```
meling, line 27: Suggest: `TeacherEnrollments()`

kit/score/results_test.go
```diff
@@ -380,25 +380,25 @@ var scoreGrades = []struct {
...
 			{TestName: "C", Score: 15, MaxScore: 15, Weight: 1},
 		},
 		out:       100,
 		wantGrade: "A",
 	},
 	{
 		in: []*score.Score{
-			{TestName: "A", Score: 05, MaxScore: 10, Weight: 1},
-			{TestName: "B", Score: 05, MaxScore: 05, Weight: 1},
+			{TestName: "A", Score: 0o5, MaxScore: 10, Weight: 1},
```
meling, line 391: I don't understand these `o` additions here? Is this some formatter thing that automatically adds these? 

The reason I added the `0` prefix was to avoid that positioning gets misaligned. That is, if we want to edit or add more test lines, it is easy to change many lines at once using multi-line editing.

kit/score/registry_example_test.go
```diff
@@ -23,13 +23,13 @@ func fibonacci(n uint) uint {
 var scoreRegistry = score.NewRegistry()
 
 func init() {
-	scoreRegistry.Add(TestFibonacciMax, len(fibonacciTests), 20)
-	scoreRegistry.Add(TestFibonacciMin, len(fibonacciTests), 20)
+	scoreRegistry.Add(TestFibonacciMax, "", len(fibonacciTests), 20)
```
meling, line 26: Shouldn't these be updated to use `AddWithTaskName()` and `AddSubWithTaskName()`

kit/score/registry.go
```diff
@@ -61,7 +61,11 @@ func (s *registry) PrintTestInfo(sorted ...bool) {
 //
 // Will panic if the test has already been registered or if max or weight is non-positive.
 func (s *registry) Add(test interface{}, max, weight int) {
-	s.add(testName(test), max, weight)
+	s.add(testName(test), "", max, weight)
+}
+
+func (s *registry) AddWithTaskName(test interface{}, taskName string, max, weight int) {
```
meling, line 67: Suggest a short version: `AddWithTask` and `AddSubWithTask`

kit/score/panic_test.go
```diff
@@ -30,12 +30,12 @@ var triangularTests = []struct {
 }
 
 func init() {
-	scores.Add(TestPanicTriangularBefore, len(triangularTests), 5)
-	scores.Add(TestPanicTriangularPanic, len(triangularTests), 5)
-	scores.Add(TestPanicTriangularAfter, len(triangularTests), 5)
-	scores.Add(TestPanicHandler, len(triangularTests), 5)
-	scores.Add(TestPanicTriangularPanicWithMsg, len(triangularTests), 5)
-	scores.Add(TestPanicHandlerWithMsg, len(triangularTests), 5)
+	scores.Add(TestPanicTriangularBefore, "", len(triangularTests), 5)
```
meling, line 33: Shouldn't these be removed again?

kit/exercise/multiple_choice_test.go
```diff
@@ -11,12 +11,12 @@ import (
 )
 
 func init() {
-	scores.Add(TestMultipleChoice0, len(tests[0].correct), 1)
-	scores.Add(TestMultipleChoice1, len(tests[1].correct), 1)
-	scores.Add(TestMultipleChoice2, len(tests[2].correct), 1)
-	scores.Add(TestMultipleChoice3, len(tests[3].correct), 1)
+	scores.Add(TestMultipleChoice0, "", len(tests[0].correct), 1)
```
meling, line 14: Remove again?

kit/score/registry_example_test.go
meling: Or rather the `""` should be removed to be compatible with the previous API, which has been reverted.

assignments/pull_requests.go
```diff
@@ -0,0 +1,98 @@
...
+
+var ErrInvalidBody = errors.New("invalid pull request body")
+
+// getLinkedIssue returns the issue number from a pull requests body.
+// E.g. 30, from the body "Fixes #30".
+func getLinkedIssue(body string) (uint64, error) {
+	if count := strings.Count(body, "#"); count != 1 {
+		return 0, ErrInvalidBody
+	}
+	_, numString, _ := strings.Cut(body, "#")
```
meling, line 24: `strings.Cut` requires Go 1.18 in `go.mod` and correspondingly in `.github/workflows/go.yml`.

PS: `golangci-lint.yml` shouldn't be updated because it doesn't yet support 1.18.

assignments/pull_requests.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/google/go-github/v35/github"
+	"go.uber.org/zap"
+)
+
+var ErrInvalidBody = errors.New("invalid pull request body")
+
+// getLinkedIssue returns the issue number from a pull requests body.
+// E.g. 30, from the body "Fixes #30".
+func getLinkedIssue(body string) (uint64, error) {
+	if count := strings.Count(body, "#"); count != 1 {
```
meling, line 21: Thinking aloud... So it is a requirement that the PR is only associated with one issue... That makes sense. However, I think the PR body could contain the `#` symbol followed by a space: `# ` indicating a markdown heading.

assignments/pull_requests.go
```diff
@@ -0,0 +1,98 @@
...
+	"strings"
+
+	pb "github.com/autograde/quickfeed/ag"
+	"github.com/autograde/quickfeed/database"
+	"github.com/autograde/quickfeed/scm"
+	"github.com/google/go-github/v35/github"
+	"go.uber.org/zap"
+)
+
+var ErrInvalidBody = errors.New("invalid pull request body")
```
meling, line 16: Suggest remove this and instead return separate, specific errors to make it easier to distinguish different problems. Such exported sentinel errors should be avoided unless you need to compare an error against the sentinel error.

assignments/pull_requests.go
```diff
@@ -0,0 +1,98 @@
...
+// getLinkedIssue returns the issue number from a pull requests body.
+// E.g. 30, from the body "Fixes #30".
+func getLinkedIssue(body string) (uint64, error) {
+	if count := strings.Count(body, "#"); count != 1 {
+		return 0, ErrInvalidBody
+	}
+	_, numString, _ := strings.Cut(body, "#")
+	issueNumber, err := strconv.Atoi(numString)
+	if err != nil {
+		return 0, ErrInvalidBody
```
meling, line 27: Could probably either return the `err` directly or wrap `err` in a `fmt.Errorf("more context...: %w", err)`

database/gormdb_course.go
```diff
@@ -164,6 +164,24 @@ func (db *GormDB) GetCoursesByUser(userID uint64, statuses ...pb.Enrollment_User
 	return courses, nil
 }
 
+// GetCourseTeachers returns a list of all teachers in a course.
+func (db *GormDB) GetCourseTeachers(query *pb.Course) ([]*pb.User, error) {
+	var course pb.Course
+	if err := db.conn.Where(query).First(&course).Error; err != nil {
+		return nil, err
+	}
+	teacherEnrollments := course.GetAllTeacherEnrollments()
```
meling, line 173: Can be called directly on line 175.

kit/score/results_test.go
oleespe: I definetly did not add these.
I recall seeing them when I first opened this file, and therefore thought they were supposed to be there.

kit/score/registry_example_test.go
oleespe: Seems I forgot to revert these when I updated the API again.
Will be changed back in next push.

assignments/pull_requests.go
oleespe: I am not sure I follow you here.
The reason I wanted the body formatted like this was so that GitHub automatically linkes the issue.
That way, the issue is closed naturally when the PR is closed.

assignments/pull_requests.go
oleespe: Part of the reason I wanted this custom error was so that I could check for it in tests.
I changed it so that I only check whether I expect an error or not.

assignments/pull_requests.go
oleespe: I also want the issue linked so that we can delete the issue db-record when the PR is correctly merged.

web/hooks/push.go
```diff
@@ -0,0 +1,259 @@
...
+		return
+	}
+	task := tasks[0]
+	// TODO(Meling): My idea is that when a teacher wants to assign a test to a specific task they will use the 'local' task name.
+	// For example, if a teacher has created the markdown file task-hello_world.md,
+	// they would do scores.AddWithTaskName(TestHelloWorld, "task-hello_world", max, weight).
+	// For the proceeding logic we need the 'global' task name however, i.e. if task-hello_world.md is in assignment1,
+	// then the full task name would be assignment1/hello_world.
+	// Do you concur with this approach?
+	// We could of course simply make teachers assign the global task name, but that seems somewhat counter-intuitive to me.
```
oleespe, line 136: While reviewing this I imagined a situation where a group pushes to two group assignments at the same time, and where the teacher has created tasks in both with the same name. E.g. groupAssignment1/task-1 and groupAssignment2/task-1. QF would detect changes in both the assignments, but given that it only scores on a per assignment basis, I think it would still work for teachers to use local task names in both.

web/hooks/push.go
```diff
@@ -0,0 +1,259 @@
...
+		for _, assignment := range assignments {
+			if assignment.IsGroupLab {
+				// only run group assignments
+				// TODO(Espeland): Need to find a way to have these tests run on the correct branch
+				results := wh.runAssignmentTests(assignment, repo, course, payload)
+
+				if !isDefaultBranch(payload) {
+					// Attempt to find the pull request for the branch, if it exists,
+					// and then assign reviewers to it, if the branch task score is higher than the assignment score limit.
+					wh.handleTaskPush(payload, results, assignment, course, repo)
```
oleespe, line 87: What happens if an assignment is graded manually? Results would be nil here.
When do we assign reviewers for a manually graded assignment?
Does manually graded simply mean that a teacher has to approve the assignment for it to pass, 
and that there are no tests? Or is it just manually approved, but with no tests?

web/hooks/push.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+}
+
+// getBranchName gets the branch name from a push event
+func getBranchName(ref string) string {
+	components := strings.Split(ref, "/")
+	return components[len(components)-1]
+}
+
+// isDefaultBranch returns true if a push event came from repository's default branch.
```
oleespe, line 256: Revise comment: returns true if a push event is for a repository's default branch.

database/gormdb_tasks.go
```diff
@@ -101,6 +101,7 @@ func (db *GormDB) SynchronizeAssignmentTasks(course *pb.Course, taskMap map[uint
 
 // CreatePullRequest creates a pull request
 func (db *GormDB) CreatePullRequest(pullRequest *pb.PullRequest) error {
+	pullRequest.Stage = pb.PullRequest_DRAFT
```
oleespe, line 104: Could use method here instead

web/hooks/pull_request.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// Currently, QF only needs to do something if the PR is approved.
+	if payload.GetReview().GetState() != "approved" {
+		wh.logger.Debug("Ignoring pull request review event for non-approved review")
+	}
+	// We make sure that the pull request is one that QF has a data record of.
+	pullRequest, err := wh.db.GetPullRequest(&pb.PullRequest{
+		ExternalRepositoryID: uint64(payload.GetRepo().GetID()),
+		Number:               uint64(payload.GetPullRequest().GetNumber()),
+	})
```
oleespe, line 26: Is this a sufficient candidate key?

web/hooks/pull_request.go
```diff
@@ -0,0 +1,198 @@
...
+	if !payload.PullRequest.GetMerged() {
+		wh.logger.Debugf("Ignoring pull request closed event for non-merged pull request #%d, in %s",
+			payload.GetPullRequest().GetNumber(), payload.GetRepo().GetFullName())
+		return
+	}
+
+	pullRequest, err := wh.db.GetPullRequest(&pb.PullRequest{
+		ExternalRepositoryID: uint64(payload.GetRepo().GetID()),
+		Number:               uint64(payload.GetPullRequest().GetNumber()),
+	})
```
oleespe, line 98: Check key.

oleespe: This is the third place where I am unsure if just ExternalRepositoryID and Number is sufficient for finding the pr.
Maybe it would be easier to add a ExternalPullRequestID field, to ensure that we get the correct PR.

web/hooks/pull_request.go
```diff
@@ -0,0 +1,198 @@
...
+	pullRequest := &pb.PullRequest{
+		ExternalRepositoryID: uint64(payload.GetRepo().GetID()),
+		TaskID:               associatedTask.GetID(),
+		IssueID:              associatedIssue.GetID(),
+		UserID:               user.GetID(),
+		SourceBranchName:     payload.GetPullRequest().GetHead().GetRef(),
+		Number:               uint64(payload.GetNumber()),
+	}
+	if !pullRequest.IsValid() {
+		wh.logger.Errorf("Failed to create pull request from event payload")
```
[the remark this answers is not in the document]
oleespe: Also, the IsValid check should probably be in the DB create function.
Should maybe rename it as well, since IsValid does not check for ID.
The method is more of a IsValidForCreation.

web/hooks/push.go
```diff
@@ -0,0 +1,259 @@
...
+// It then extracts the score from all tests belonging to that task
+// If a passing score is reached, reviewers are assigned.
+func (wh GitHubWebHook) handleTaskPush(payload *github.PushEvent, results *score.Results, assignment *pb.Assignment, course *pb.Course, repo *pb.Repository) {
+	wh.logger.Debugf("Attempting to find pull request for ref: %s, in repository: %s",
+		payload.GetRef(), payload.GetRepo().GetFullName())
+
+	pullRequest, err := wh.db.GetPullRequest(&pb.PullRequest{
+		SourceBranchName:     getBranchName(payload.GetRef()),
+		ExternalRepositoryID: uint64(payload.GetRepo().GetID()),
+	})
```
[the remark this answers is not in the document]
oleespe: It seems like you can only have one PR per branch, so this should be ok.

web/hooks/pull_request.go
oleespe: It seems like you can only have one PR per branch, so this should be ok.

web/hooks/pull_request.go
oleespe: It seems like you can only have one PR per branch, so this should be ok.

web/hooks/pull_request.go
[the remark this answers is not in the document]
oleespe: Moved it, and renamed it to Validate

ag/group.go
```diff
@@ -32,3 +32,15 @@ func (g *Group) SetSlipDays(c *Course) {
 		e.SetSlipDays(c)
 	}
 }
+
+// GetUserSubset returns a list of all users in a group, except the one with the given userID.
+func (g *Group) GetUserSubset(userID uint64) []*User {
```
meling, line 37: Better name is: `GetUsersExcept(userID...`

ag/ag.proto
```diff
@@ -199,6 +199,23 @@ message Issue {
...
+        DRAFT   = 1;
+        REVIEW  = 2;
+        APPROVED = 3;
+    }
+    uint64 ID                   = 1;
+    uint64 externalRepositoryID = 3; // Represents the external repository ID
+    uint64 taskID               = 4; // foreign key
+    uint64 issueID              = 5; // foreign key
+    uint64 userID               = 6; // The user who created this PR
+    string sourceBranchName     = 7; // The source branch for this pull request
```
meling, line 214: `sourceBranch`

ag/ag.proto
```diff
@@ -199,6 +199,23 @@ message Issue {
...
 
+message PullRequest {
+    enum Stage {
+        NONE    = 0;
+        DRAFT   = 1;
+        REVIEW  = 2;
+        APPROVED = 3;
+    }
+    uint64 ID                   = 1;
+    uint64 externalRepositoryID = 3; // Represents the external repository ID
```
meling, line 210: `repositoryID`

web/hooks/push.go
```diff
@@ -0,0 +1,260 @@
...
+		LastActivityDate: time.Now().Format("02 Jan"),
+	}
+
+	if err := wh.db.UpdateEnrollment(query); err != nil {
+		wh.logger.Errorf("Failed to update the last activity date for user %d: %v", userID, err)
+	}
+}
+
+// getBranchName returns the branch name from a push event ref.
+func getBranchName(ref string) string {
```
meling, line 252: `branchName`

web/hooks/pull_request.go
```diff
@@ -0,0 +1,196 @@
...
+	"gorm.io/gorm"
+)
+
+func (wh GitHubWebHook) handlePullRequestReview(payload *github.PullRequestReviewEvent) {
+	wh.logger.Debugf("Received pull request review event for pull request: %s, in repository: %s",
+		payload.GetPullRequest().GetTitle(), payload.GetRepo().GetName())
+
+	// Currently, QF only needs to do something if the PR is approved.
+	if payload.GetReview().GetState() != "approved" {
+		wh.logger.Debug("Ignoring pull request review event for non-approved review")
```
meling, line 20: This looks like it should return after logging and not continue...?

ag/ag.proto
oleespe: Is that not confusing? repositoryID can mean both the internal repository ID and the external GitHub repository ID.

meling: Not sure. Do we actually use the database internal repository ID as a foreign key? I agree that it might be good to distinguish these... I checked:

```proto
message Repository {
...
    uint64 repositoryID   = 3;  // ID of a github repository
```
```proto
message Issue {
    uint64 ID           = 1;
    uint64 repositoryID = 2;  // Represents the internal ID of a repository
```

Maybe we should adopt a convention for this stuff; e.g., to prefix database specific (foreign key) IDs with `dbRespositoryID`... But perhaps we should do that separately in a different PR that fixes all such issues. Alternatively, we could prefix other external IDs with e.g., `scmRepositoryID`. Not sure what makes the most sense. I'll create a separate issue for this.

PS: Noticed also that field number `2` seems to be missing.

oleespe: I will leave things as is for now then, and add a TODO(meling) for reference.
I also fixed the field number in the most recent commit.

web/hooks/pull_request.go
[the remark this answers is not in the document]
oleespe: Still not happy with this..

web/hooks/push.go
oleespe: It seems manually graded implies that an assignment has no tests.
But an assignment that uses pull requests requires tests to work.
An assignment would therefore have to be set as automatically approved, which kind of contradicts the pull request approval process.

oleespe: Apperantly, manually graded simply means that the reviewers field is greater than 0.
Any assignment using pull requests and so on can therefore not be manually graded, since they produce no test results.

JosteinLindhom: Indeed. 
Manual grading has teachers / TAs reviewing submitted code, and assigning a grade (passed/failed) to a number of criteria:
image
These criteria are defined in a `criteria.json` file. Look [here](https://github.com/quickfeed/quickfeed/blob/master/doc/teacher.md#reviewing-student-submissions) for more information if interested.

These submissions will not be tested. The submissions will be assigned a score based on how many of the criteria of a review have been approved.

The above method is only used if the `reviewers` field in `assignment.yml` is greater than 0.

For submissions that are tested, they will be graded (`Approved`, `Revision`, `Failed`) automatically if `autoapprove` is set to `true`. Otherwise, teachers / TAs have to manually assign a grade to the submission.

ag/ag.proto
```diff
@@ -199,6 +199,25 @@ message Issue {
...
+message PullRequest {
+    enum Stage {
+        NONE    = 0;
+        DRAFT   = 1;
+        REVIEW  = 2;
+        APPROVED = 3;
+    }
+    uint64 ID                   = 1;
+    // TODO(meling): When a naming convention has been decided, change this field's name.
+    uint64 externalRepositoryID = 2; // Represents the external repository ID
```
meling, line 211: We should update this to `scmRepositoryID` pending a final decision, but so far `scm` is my preference.

web/hooks/push.go
oleespe: Thank you for further info.

I was really scratching my head the other day, trying to understand the difference between graded and approved.
The specific example where this caused confusion is the following code:
```go
if assignment.GradedManually() {
wh.logger.Debugf("Assignment %s for course %s is manually reviewed", assignment.Name, course.Name)
if _, err := runData.RecordResults(wh.logger, wh.db, nil); err != nil {
	wh.logger.Error(err)
}
return nil
}
```
Intuitively, I imagined thinking that the GradedManually method checked whether an assignment is manually approved.
Would it maybe be better to have this method called ReviewedManually?
Even the log seems to imply this.

oleespe: Clarifying my confusion between manually graded and manually approved.

To me, a manually graded **Assignment** would mean an assignment that gets a grade from A-F.
Therfore, any grade above F would imply that it has passed, and is therby also approved.

ag/ag.proto
```diff
@@ -199,6 +199,25 @@ message Issue {
...
+    uint64 taskID               = 3; // foreign key
+    uint64 issueID              = 4; // foreign key
+    uint64 userID               = 5; // The user who owns this PR
+    uint64 commentID            = 6; // GitHub ID of the comment used for automatic feedback
```
meling, line 215: I guess this one should also be renamed to `scmCommentID` and the comment should probably not mention GitHub ; presumably we could implement a similar thing with GitLab in the future.

ag/pull_request_test.go
```diff
@@ -0,0 +1,48 @@
...
+	tests := map[string]struct {
+		pr   *pb.PullRequest
+		want bool
+	}{
+		"Valid":                        {pr: &pb.PullRequest{ExternalRepositoryID: 1, TaskID: 1, IssueID: 1, UserID: 1, SourceBranch: "A", Number: 1}, want: true},
+		"Invalid ExternalRepositoryID": {pr: &pb.PullRequest{ExternalRepositoryID: 0, TaskID: 1, IssueID: 1, UserID: 1, SourceBranch: "A", Number: 1}, want: false},
+		"Invalid TaskID":               {pr: &pb.PullRequest{ExternalRepositoryID: 1, TaskID: 0, IssueID: 1, UserID: 1, SourceBranch: "A", Number: 1}, want: false},
+		"Invalid IssueID":              {pr: &pb.PullRequest{ExternalRepositoryID: 1, TaskID: 1, IssueID: 0, UserID: 1, SourceBranch: "A", Number: 1}, want: false},
+		"Invalid UserID":               {pr: &pb.PullRequest{ExternalRepositoryID: 1, TaskID: 1, IssueID: 1, UserID: 0, SourceBranch: "A", Number: 1}, want: false},
+		"Invalid SourceBranchName":     {pr: &pb.PullRequest{ExternalRepositoryID: 1, TaskID: 1, IssueID: 1, UserID: 1, SourceBranch: "", Number: 1}, want: false},
```
meling, line 19: `SourceBranch`

ag/pull_request_test.go
```diff
@@ -0,0 +1,48 @@
...
+	pb "github.com/autograde/quickfeed/ag"
+)
+
+func TestIsValid(t *testing.T) {
+	tests := map[string]struct {
+		pr   *pb.PullRequest
+		want bool
+	}{
+		"Valid":                        {pr: &pb.PullRequest{ExternalRepositoryID: 1, TaskID: 1, IssueID: 1, UserID: 1, SourceBranch: "A", Number: 1}, want: true},
+		"Invalid ExternalRepositoryID": {pr: &pb.PullRequest{ExternalRepositoryID: 0, TaskID: 1, IssueID: 1, UserID: 1, SourceBranch: "A", Number: 1}, want: false},
```
meling, line 15: `ScmRepositoryID`

ag/task.go
```diff
@@ -22,3 +22,10 @@ func (t *Task) MarkDeleted() {
 func (t *Task) IsDeleted() bool {
 	return strings.HasPrefix(t.Title, deleted)
 }
+
+// LocalName returns the task name without the assignment part,
+// i.e. if a task has the name "assignment1/hello_world". it returns "hello_world".
+func (t *Task) LocalName() string {
+	s := strings.Split(t.GetName(), "/")
```
meling, line 29: Recommended approach with Go 1.18. You will need to also update `go.mod`
```go
	_, s, _ := strings.Cut(t.GetName(), "/")
	return s
```

assignments/pull_requests.go
```diff
@@ -0,0 +1,157 @@
...
+func CreateFeedbackComment(results *score.Results, task *pb.Task, assignment *pb.Assignment) string {
+	body := "## Test results from latest push\n"
+	body += "| Test Name | Score | Weight | % of Total |\n"
+	body += "| :-------- | :---- | :----- | ---------: |\n"
+
+	for _, score := range results.Scores {
+		if score.TaskName != task.LocalName() {
+			continue
+		}
+		percentageScore := (float64(score.Score) / float64(score.MaxScore)) * (float64(score.Weight) / results.TotalTaskWeight(task.LocalName()))
```
meling, line 48: We already calculate something similar in `kit/score/results.go`. This computation should also be done in this file.

assignments/pull_requests.go
```diff
@@ -0,0 +1,157 @@
...
+//  | Test 2	| 1/3   | 2	     |     24.46% |
+//  |   |		|  |    | |      |            |
+//  |   |		|  |    | |      |            |
+//  |   |		|  |    | |      |            |
+//  |   |		|  |    | |      |            |
+//  | Total		|		|		 |	   48.23% |
+// Once a total score of 80% is reached, reviewers are automatically assigned.
+//
+func CreateFeedbackComment(results *score.Results, task *pb.Task, assignment *pb.Assignment) string {
+	body := "## Test results from latest push\n"
```
meling, line 40: Should have two `\n` at end of this string.

assignments/pull_requests.go
```diff
@@ -0,0 +1,157 @@
...
+	teacherReviewCounter = make(map[uint64]map[uint64]int)
+	groupReviewCounter   = make(map[uint64]map[uint64]int)
+)
+
+// CreateFeedbackComment formats a feedback comment to be posted on pull requests.
+// It uses the test results from a student commit to create a table like the one shown below.
+// Only the test scores associated with the supplied task are used to generate this table.
+// Table formatting ref: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-tables
+//
+//  ## Test results from latest push
```
meling, line 27: Maybe use a full example instead of the many blank cells. Remove spaces before the two header lines. Add newline after the table.

assignments/pull_requests.go
meling: This could also be given as a multi-line string using the backtick symbol.

assignments/pull_requests.go
```diff
@@ -0,0 +1,157 @@
...
+//  | Test 1	| 2/6	| 1		 |	   12.23% |
+//  | Test 2	| 1/3   | 2	     |     24.46% |
+//  |   |		|  |    | |      |            |
+//  |   |		|  |    | |      |            |
+//  |   |		|  |    | |      |            |
+//  |   |		|  |    | |      |            |
+//  | Total		|		|		 |	   48.23% |
+// Once a total score of 80% is reached, reviewers are automatically assigned.
+//
+func CreateFeedbackComment(results *score.Results, task *pb.Task, assignment *pb.Assignment) string {
```
meling, line 39: This method does not need the `Task` struct; it only uses the `task.LocalName()`. And in `push.go` we similarly only need the local task name (see other comment to be added).

assignments/pull_requests.go
```diff
@@ -0,0 +1,157 @@
...
+	body += "| :-------- | :---- | :----- | ---------: |\n"
+
+	for _, score := range results.Scores {
+		if score.TaskName != task.LocalName() {
+			continue
+		}
+		percentageScore := (float64(score.Score) / float64(score.MaxScore)) * (float64(score.Weight) / results.TotalTaskWeight(task.LocalName()))
+		body += fmt.Sprintf("| %s | %d/%d | %d | %.2f%% |\n", score.TestName, score.Score, score.MaxScore, score.Weight, percentageScore*100)
+	}
+	// TODO(espeland): TaskSum retruns an int, while a float is used for individual tests
```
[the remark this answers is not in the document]
meling: Yes, I'm aware... Actually, both the new and old React frontends doesn't give a percentage for individual tests. We just give the ratio as a fraction `Score / MaxScore`. Perhaps we just do the same here.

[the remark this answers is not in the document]
meling: I guess that would mean we could eliminate the last column and give the total percentage in the same column as the score ratio, similar to the current frontends...

web/hooks/push.go
```diff
@@ -0,0 +1,286 @@
...
+
+// handlePullRequestPush attempts to find a pull requested associated with the non-default branch.
+// If successfull, it then finds the relevant task, and uses it to receive the relevant task score.
+// If a passing score is reached, it assigns reviewers to the pull request.
+// It also uses the test results and task to generate a feedback comment for the pull request.
+func (wh GitHubWebHook) handlePullRequestPush(payload *github.PushEvent, results *score.Results, assignment *pb.Assignment, course *pb.Course, repo *pb.Repository) {
+	wh.logger.Debugf("Attempting to find pull request for ref: %s, in repository: %s",
+		payload.GetRef(), payload.GetRepo().GetFullName())
+
+	pullRequest, err := wh.db.GetPullRequest(&pb.PullRequest{
```
meling, line 105: Suggest to move Lines 105-131 to a separate method that returns `pullRequest, taskName`. The rest of this method (after 131) only needs the `task.LocalName()`. See related comment in `pull_request.go:CreateFeedbackComment`.

assignments/pull_requests.go
```diff
@@ -0,0 +1,157 @@
...
+	return teacherReviewer, nil
+}
+
+// getNextStudentReviewer gets the student in a group with the least total reviews.
+func getNextStudentReviewer(db database.Database, groupID, ownerID uint64) (*pb.User, error) {
+	group, err := db.GetGroup(groupID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get group from database: %w", err)
+	}
+	if len(group.Users) == 0 {
```
meling, line 147: Since this should never happen; let's move this check into `db.GetGroup()` and return an error. It basically means the database has become inconsistent -- there should be checks that a group has members when adding/updating the group.

PS: I'm not sure if this will pass existing tests; didn't check. But if it doesn't it should be fixed.

assignments/pull_requests.go
```diff
@@ -0,0 +1,157 @@
...
+	return userWithLowestCount, nil
+}
+
+// getNextTeacherReviewer gets the teacher with the least total reviews.
+func getNextTeacherReviewer(db database.Database, course *pb.Course) (*pb.User, error) {
+	teachers, err := db.GetCourseTeachers(course)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get teachers from database: %w", err)
+	}
+	if len(teachers) == 0 {
```
meling, line 131: Same reasoning as below. A course must always have a teacher, and `db.GetCourseTeachers()` should return an error if it doesn't.

assignments/tasks_test.go
```diff
@@ -21,10 +21,10 @@ type foundIssue struct {
 
 // The test environment creates tasks based on the issues found on student repositories in the organization.
 // This is so that we can emulate the result of a previous push to the tests repository.
-// For these tasks to be created correctly the title of each task must use the following format:
+// For these tasks to be created correctly the title of each issue must use the following format:
```
meling, line 24: This requirement seems unnecessary to me. Can't we use the string `lab1/hello_world` or just `lab1` to lookup the assignment order from somewhere? (Didn't check)

assignments/pull_requests.go
```diff
@@ -0,0 +1,157 @@
...
+	pullRequest.SetReview()
+	return db.UpdatePullRequest(pullRequest)
+}
+
+// getNextReviewer gets the next reviewer from either teacherReviewCounter or studentReviewCounter,
+// based on whoever in total has been assigned to the least amount of pull requests.
+// It is simple, and does not account for how many current review requests any user has.
+//
+// Returns an error if the list of users is empty.
+func getNextReviewer(ID uint64, users []*pb.User, reviewCounter map[uint64]map[uint64]int) (*pb.User, error) {
```
meling, line 94: I think this would be much cleaner if you instead pass in a single `map[uint64]int` and do the `id` lookup and map initialization from the call site, e.g., maybe using something like:

```go
type countMap map[uint64]map[uint64]int

func (m countMap) newMap(id uint64) {
	m[id] = make(map[uint64]int)
}
```

Then you don't need the `id` variable (remember to use lower case for unexported variables).

assignments/pull_requests.go
```diff
@@ -0,0 +1,157 @@
...
+	if len(users) == 0 {
```
meling, line 95: This seems to add unnecessary complexity for the user having to check for an error that will never happen. Suggest to remove this check; since it is only used internally (unexported function).

assignments/tasks.go
```diff
@@ -75,6 +75,11 @@ func handleTasks(ctx context.Context, db database.Database, sc scm.SCM, course *
 		return err
 	}
 
+	// TODO(Espeland): This will create all tasks in the course creator's name.
```
meling, line 78: Not sure I see a problem with this? That said, the new GitHub app that @0xf8f8ff is developing in the context of the improved security architecture -- for that we may need to reconsider this.

assignments/tasks.go
```diff
@@ -75,6 +75,11 @@ func handleTasks(ctx context.Context, db database.Database, sc scm.SCM, course *
...
+	// A possible workaround is to create a new scm client for every repo, and create the issues in one of the students name.
+
+	// TODO(Espeland): Need to think about what happens if an assignment contains two assignments with the same name.
```
meling, line 81: Do you mean issues instead of assignments?

ci/parse_script.go
```diff
@@ -24,7 +25,18 @@ type AssignmentInfo struct {
 // {{ .AssignmentName }}, {{ .RandomSecret }}, etc.
 func (r RunData) parseScriptTemplate(secret string) (*Job, error) {
 	info := &AssignmentInfo{
-		AssignmentName:     r.Assignment.GetName(),
+		AssignmentName: r.Assignment.GetName(),
+		// TODO(Meling): I have added BranchName here, since we need to checkout the corresponding branch when necessary.
```
meling, line 29: Please remove this TODO from here and open an issue on the issue tracker to explain what needs to be done. That said, I have some unfinished code locally that should simplify the script, including avoiding this extra if and git checkout.

assignments/tasks.go
oleespe: Both this comment, and the previous are solved by QuickFeed being a GitHub app, and therefore both of these todo's should be removed.
I somewhat discuss this in the report.

internal/qtest/test_helper.go
```diff
@@ -168,8 +184,16 @@ func PopulateDatabaseWithInitialData(t *testing.T, db database.Database, sc scm.
 	if err != nil {
 		return err
 	}
-
-	admin := CreateFakeUser(t, db, uint64(1))
+	course.OrganizationID = org.GetID()
+	// TODO(Espeland): Remember to remove myself when done testing!
+	admin := &pb.User{Login: "oleespe2"}
```
meling, line 189: Perhaps a better solution is to add another function, e.g., `GetGitHubLogin` in `scm/env_helper.go` to read from `GITHUB_LOGIN` environment variable. I guess such a function could be given an boolean `skip` argument if the test should be skipped if the login is required for the test itself... if it is not required for the full test, then I guess it shouldn't be skip the test, but avoid using the specific user...

internal/qtest/test_helper.go
```diff
@@ -194,19 +217,42 @@ func PopulateDatabaseWithInitialData(t *testing.T, db database.Database, sc scm.
 		case pb.TestsRepo:
 			dbRepo.RepoType = pb.Repository_TESTS
 		default:
-			// TODO(Espeland): Could use CreateNamedUser() instead.
-			user = CreateFakeUser(t, db, nxtRemoteID)
+			login := strings.TrimSuffix(dbRepo.Name(), "-labs")
```
meling, line 220: This can be added as a helper in `ag/repo.go`:

```go
func (t *Repository) User() *User {
	repoName := t.Name()
	login := repoName[:len(repoName)-len(StudentRepoSuffix)]
	return &User{Login: login}
}
```

Alternatively, if it is more useful to return the `login` string, that's fine too.

scm/helper.go
```diff
@@ -116,6 +116,16 @@ func (opt *CreateIssueOptions) valid() bool {
...
 }
 
+func (opt RequestReviewersOptions) valid() bool {
+	return opt.Organization != "" && opt.Repository != "" &&
+		opt.Number > 0 && len(opt.Reviewers) != 0
+}
+
+func (opt IssueCommentOptions) valid() bool {
+	return opt.Organization != "" && opt.Repository != "" &&
+		opt.Body != ""
```
meling, line 126: Don't need extra line

scm/helper.go
```diff
@@ -116,6 +116,16 @@ func (opt *CreateIssueOptions) valid() bool {
 	return (opt.Organization != "" && opt.Repository != "" && opt.Title != "" && opt.Body != "")
```
meling, line 116: Don't need the parens here.

web/hooks/push_test.go
```diff
@@ -0,0 +1,31 @@
+package hooks
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestExtractChanges(t *testing.T) {
```
meling, line 9: Doesn't look like it is related to push functionality. Please rename to `extract_changes_test.go` and correspondingly for the relevant functions (from other files) should be renamed to `extract_changes.go`

assignments/pull_requests_test.go
```diff
@@ -0,0 +1,163 @@
...
+			{TestName: "Test2", TaskName: "1", Score: 3, MaxScore: 9, Weight: 3},
+			{TestName: "Test3", TaskName: "1", Score: 8, MaxScore: 8, Weight: 5},
+			{TestName: "Test4", TaskName: "1", Score: 2, MaxScore: 5, Weight: 1},
+			{TestName: "Test5", TaskName: "1", Score: 5, MaxScore: 7, Weight: 1},
+			{TestName: "Test6", TaskName: "2", Score: 5, MaxScore: 7, Weight: 1},
+			{TestName: "Test7", TaskName: "3", Score: 5, MaxScore: 7, Weight: 1},
+		},
+	}
+	body := CreateFeedbackComment(results, &pb.Task{Name: "lab1/1"}, &pb.Assignment{ScoreLimit: 80})
+	// TODO(espeland): Remember to reset these when done testing
```
meling, line 154: Probably better to use something like `GITHUB_LOGIN` env var

assignments/pull_requests_test.go
```diff
@@ -0,0 +1,163 @@
...
+	opt := &scm.IssueCommentOptions{
+		Organization: qfTestOrg,
+		Repository:   "oleespe-labs",
+		Body:         body,
+	}
+	if err := s.EditIssueComment(context.Background(), 1117670404, opt); err != nil {
```
meling, line 160: Couldn't we just use `scm.GetIssues()` command to get the issue number?

database/gormdb_course_test.go
```diff
@@ -357,3 +357,51 @@ func TestGormDBCourseUniqueContraint(t *testing.T) {
...
+
+func TestGetCourseTeachers(t *testing.T) {
+	tests := map[string]struct {
+		wantTeachers, students []*pb.User
+	}{
+		"Basic": {
+			wantTeachers: []*pb.User{{Login: "teacher1"}, {Login: "teacher2"}},
+			students:     []*pb.User{{Login: "student1"}},
+		},
+		"No teachers": {
```
meling, line 369: I think this should expect an error and not an empty `User` slice because the `CreateCourse` method requires a user id / course creator and said user is enrolled as teacher. Hence `wantTeachers` should be `nil`. There should be an additional field here to check for the expected error.

database/gormdb_tasks.go
```diff
@@ -98,3 +99,45 @@ func (db *GormDB) SynchronizeAssignmentTasks(course *pb.Course, taskMap map[uint
 
 	return createdTasks, updatedTasks, err
 }
+
+// CreatePullRequest creates a pull request
+func (db *GormDB) CreatePullRequest(pullRequest *pb.PullRequest) error {
+	if !pullRequest.Validate() {
+		return errors.New("pull request is not valid for creation")
+	}
+	pullRequest.SetDraft()
```
meling, line 108: Perhaps we should document in the comment and elsewhere that initially it is set to draft mode.

database/gormdb_tasks.go
```diff
@@ -98,3 +99,45 @@ func (db *GormDB) SynchronizeAssignmentTasks(course *pb.Course, taskMap map[uint
...
+	var pullRequest pb.PullRequest
+	if err := db.conn.Where(query).Last(&pullRequest).Error; err != nil {
+		return nil, err
+	}
+	return &pullRequest, nil
+}
+
+// HandleMergingPR handles merging a pull request
+func (db *GormDB) HandleMergingPR(pullRequest *pb.PullRequest) error {
+	// If a pull request has not been approved, it should not have been merged.
```
meling, line 123: Move comment to details to the doc comment, since this is not explaining specifics of the code, but the general idea.

scm/github.go
```diff
@@ -704,53 +705,83 @@ func (s *GithubSCM) CreateIssue(ctx context.Context, opt *CreateIssueOptions) (*
...
 	if err != nil {
 		return nil, ErrFailedSCM{
 			Method:   "CreateIssue",
-			Message:  fmt.Sprintf("failed to create Issue %s, make sure all the details are correct ", opt.Title),
+			Message:  fmt.Sprintf("failed to create issue %s, make sure all the details are correct ", opt.Title),
 			GitError: err,
 		}
 	}
-	s.logger.Debugf("CreateRepository: done creating issue %s", opt.Title)
+	s.logger.Debugf("Done creating issue %s", opt.Title)
```
meling, line 717: `"Created issue %q"`

scm/github.go
```diff
@@ -704,53 +705,83 @@ func (s *GithubSCM) CreateIssue(ctx context.Context, opt *CreateIssueOptions) (*
 		Assignees: opt.Assignees,
 	}
 
-	s.logger.Debugf("CreateRepository: creating issue %s on %s Repository", opt.Title, opt.Repository)
+	s.logger.Debugf("Creating issue %s on %s Repository", opt.Title, opt.Repository)
```
meling, line 708: `"Creating issue %q on %s"`

Hopefully the `opt.Repository` string makes it clear that it is a repository.

The `%q` should add quotes `"` around the issue title. Since it may contain spaces, it is helpful to place the title inside quotes.

scm/github.go
```diff
@@ -704,53 +705,83 @@ func (s *GithubSCM) CreateIssue(ctx context.Context, opt *CreateIssueOptions) (*
...
 	issue, _, err := s.client.Issues.Create(ctx, opt.Organization, opt.Repository, newIssue)
 	if err != nil {
 		return nil, ErrFailedSCM{
 			Method:   "CreateIssue",
-			Message:  fmt.Sprintf("failed to create Issue %s, make sure all the details are correct ", opt.Title),
+			Message:  fmt.Sprintf("failed to create issue %s, make sure all the details are correct ", opt.Title),
```
meling, line 713: Probably use `%q` also here. The last portion with `, make sure...` may not be so useful... If keep it, remove the last space after `correct`

scm/github.go
```diff
@@ -704,53 +705,83 @@ func (s *GithubSCM) CreateIssue(ctx context.Context, opt *CreateIssueOptions) (*
...
-	s.logger.Debugf("CreateRepository: done creating issue %s", opt.Title)
+	s.logger.Debugf("Done creating issue %s", opt.Title)
 
 	return toIssue(issue), nil
 }
 
-// GetRepoIssue implements the SCM interface
-func (s *GithubSCM) GetRepoIssue(ctx context.Context, issueNumber int, opt *RepositoryOptions) (*Issue, error) {
+// EditRepoIssue implements the SCM interface
+func (s *GithubSCM) EditIssue(ctx context.Context, issueNumber int, opt *CreateIssueOptions) (*Issue, error) {
```
meling, line 723: Perhaps this method should be called `UpdateIssue` instead of `EditIssue`? We don't have to follow GitHub's naming, if something else makes more sense for us.

scm/github.go
```diff
@@ -704,53 +705,83 @@ func (s *GithubSCM) CreateIssue(ctx context.Context, opt *CreateIssueOptions) (*
...
+	s.logger.Debugf("EditIssue: Editing issue %s on %s Repository", issueNumber, opt.Repository)
+	issue, _, err := s.client.Issues.Edit(ctx, opt.Organization, opt.Repository, issueNumber, issueReq)
+	if err != nil {
+		return nil, ErrFailedSCM{
+			Method:   "EditIssue",
+			Message:  fmt.Sprintf("failed to edit issue #%v, on repository: %s, at organization: %s", issueNumber, opt.Repository, opt.Organization),
+			GitError: err,
+		}
+	}
+	s.logger.Debugf("EditIssue: Done editing issue number  %s", issueNumber)
```
meling, line 747: `Updated issue number %d`

scm/github.go
```diff
@@ -704,53 +705,83 @@ func (s *GithubSCM) CreateIssue(ctx context.Context, opt *CreateIssueOptions) (*
...
+		Assignee:  opt.Assignee,
+		Assignees: opt.Assignees,
+	}
+
+	s.logger.Debugf("EditIssue: Editing issue %s on %s Repository", issueNumber, opt.Repository)
+	issue, _, err := s.client.Issues.Edit(ctx, opt.Organization, opt.Repository, issueNumber, issueReq)
+	if err != nil {
+		return nil, ErrFailedSCM{
+			Method:   "EditIssue",
+			Message:  fmt.Sprintf("failed to edit issue #%v, on repository: %s, at organization: %s", issueNumber, opt.Repository, opt.Organization),
```
meling, line 743: Does this make sense?

`"failed to update issue %d on %s/%s", issueNumber, opt.Organization, opt.Repository`

(note the reordering of org and repo)

scm/github.go
```diff
@@ -704,53 +705,83 @@ func (s *GithubSCM) CreateIssue(ctx context.Context, opt *CreateIssueOptions) (*
...
 		}
 	}
 
 	issue, _, err := s.client.Issues.Get(ctx, opt.Owner, opt.Path, issueNumber)
 	if err != nil {
 		return nil, ErrFailedSCM{
-			Method:   "GetRepoIssue",
-			Message:  fmt.Sprintf("Failed to get Issue %v, make sure all the details are correct ", issueNumber),
+			Method:   "GetIssue",
+			Message:  fmt.Sprintf("failed to get Issue %v, make sure all the details are correct ", issueNumber),
```
meling, line 765: `%d` and we do not need the `, make sure ...` part.

scm/github.go
```diff
@@ -704,53 +705,83 @@ func (s *GithubSCM) CreateIssue(ctx context.Context, opt *CreateIssueOptions) (*
...
 			Message: fmt.Sprintf("%+v", opt),
 		}
 	}
 	issueList, _, err := s.client.Issues.ListByRepo(ctx, opt.Owner, opt.Path, &github.IssueListByRepoOptions{})
 	if err != nil {
 		return nil, ErrFailedSCM{
-			Method:   "GetRepoIssues",
-			Message:  fmt.Sprintf("Failed to get Issues from repo %s make sure all the details are correct ", opt.Path),
+			Method:   "GetIssues",
+			Message:  fmt.Sprintf("failed to get Issues from repo %s make sure all the details are correct ", opt.Path),
```
meling, line 784: `failed to get issues for %s`

We do not need the `make sure ...` part.

scm/github.go
```diff
@@ -762,34 +793,64 @@ func (s *GithubSCM) GetRepoIssues(ctx context.Context, opt *RepositoryOptions) (
...
 		}
 	}
-	s.logger.Debugf("EditRepoIssue: done Editing issue number  %s", issueNumber)
 
-	return toIssue(issue), nil
+	return uint64(createdComment.GetID()), nil
+}
+
+// EditIssueComment implements the SCM interface
+func (s *GithubSCM) EditIssueComment(ctx context.Context, commentID int64, opt *IssueCommentOptions) error {
```
meling, line 838: Maybe `UpdateComment`...

assignments/pull_requests_test.go
oleespe: That is the commentID, not issue number.
I will think about a way to make this test more approachable.

scm/github_test.go
```diff
@@ -222,8 +223,74 @@ func TestEditRepoIssue(t *testing.T) {
...
+
+func TestCreateIssueComment(t *testing.T) {
+	qfTestOrg := scm.GetTestOrganization(t)
+	accessToken := scm.GetAccessToken(t)
+	s, err := scm.NewSCMClient(zap.NewNop().Sugar(), "github", accessToken)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// TODO(espeland): Remember to reset these when done testing
```
meling, line 262: Fix

scm/github_test.go
```diff
@@ -222,8 +223,74 @@ func TestEditRepoIssue(t *testing.T) {
...
+	s, err := scm.NewSCMClient(zap.NewNop().Sugar(), "github", accessToken)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	body := "| Test Name | Score | Weight | % of total |\n"
+	body += "| :-------- | :---- | :----- | ---------: |\n"
+	body += "| Test 1    |  6/7  |    1   | 20%        |\n"
+	body += "| **Total** |       |        | 20%        |\n"
+	// TODO(espeland): Remember to reset these when done testing
```
meling, line 287: Fix

scm/github_test.go
```diff
@@ -222,8 +223,74 @@ func TestEditRepoIssue(t *testing.T) {
...
+	opt := &scm.IssueCommentOptions{
+		Organization: qfTestOrg,
+		Repository:   "oleespe-labs",
+		Body:         body,
+	}
+	if err := s.EditIssueComment(context.Background(), 1117670404, opt); err != nil {
```
meling, line 293: Hardcoded issue number?

Can the test obtain the issue number from a previous create?

web/hooks/github.go
```diff
@@ -47,159 +43,19 @@ func (wh GitHubWebHook) Handle(w http.ResponseWriter, r *http.Request) {
 	case *github.PushEvent:
 		wh.logger.Debug(log.IndentJson(e))
```
meling, line 44: We log these events for all relevant events. Now that we limit the events that we receive, perhaps we can move this log outside the switch-case and log all events we receive, since they should be the same as those we currently switch on... I think??

web/hooks/github_test.go
```diff
@@ -76,26 +78,151 @@ func TestGitHubWebHook(t *testing.T) {
...
+		if err != nil {
+			return err
+		}
+
+		if len(existingScmIssues) == 0 {
+			continue
+		}
+		foundIssues[repo.ID] = make(map[string]*foundIssue)
+		for _, scmIssue := range existingScmIssues {
+			splitTitle := strings.Split(scmIssue.Title, ", ")
```
meling, line 185: I'm not a fan of this formatted string to get the assignmentOrder. Is this also used for the production version? For testing, it might be okay... although I would prefer a more robust approach.

web/hooks/github_test.go
```diff
@@ -76,26 +78,151 @@ func TestGitHubWebHook(t *testing.T) {
...
+	}
+	defer runner.Close()
+	webhook := NewGitHubWebHook(logger.Sugar(), db, runner, secret)
+
+	log.Println("starting webhook server")
+	http.HandleFunc("/webhook", webhook.Handle)
+	log.Fatal(http.ListenAndServe(":8080", nil))
+}
+
+// TODO(Espeland): This function is almost identical to the one in assignments/tasks_test.go.
```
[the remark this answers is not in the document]
meling: PS: If this helper is only meant to populate the database with tasks then maybe it doesn't need to fetch assignments from the tests repo? And if so can be simplified...

web/hooks/pull_request_test.go
```diff
@@ -0,0 +1,29 @@
...
+	}{
+		"Simple":         {body: "Fixes #30", expectError: false},
+		"Not a number":   {body: "Fixes #30nan", expectError: true},
+		"Invalid body":   {body: "Fixes #30nan #", expectError: true},
+		"Invalid body 2": {body: "Fixes", expectError: true},
+	}
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			gotIssueNumber, err := getLinkedIssue(tt.body)
+			if err != nil {
```
meling, line 19: This captures the case when we get an error and it is expected. We should also capture the case when we get an error and it is unexpected, e.g., something like:
```go
if err != nil {
  if tt.expectError {
    return
  }
  t.Errorf("getLinkedIssue() = %v, no error expected", err)
}
```
On the other hand, it might be easier to just make the map struct provide the expected error as a string:
```go
tests := map[string]struct {
	body   string
	errStr string
}{
	"Simple": {body: "Fixes #30", errStr: ""},
	"Not a number": {body: "Fixes #30nan", err: "failed to parse issue number..."},
}
```
You could also use `error` instead of string, but then it is tricker to create the correct error...

web/hooks/pull_request.go
```diff
@@ -0,0 +1,198 @@
...
+		return
+	}
+	// We make sure that the pull request is one that QF has a data record of
+	pullRequest, err := wh.db.GetPullRequest(&pb.PullRequest{
+		ExternalRepositoryID: uint64(payload.GetRepo().GetID()),
+		Number:               uint64(payload.GetPullRequest().GetNumber()),
+	})
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			wh.logger.Debugf("Ignoring pull request review event for non-managed pull request #%d, in %s",
```
meling, line 30: `"Ignoring review event for unknown pull request #%d in %s"`

web/hooks/pull_request.go
```diff
@@ -0,0 +1,198 @@
...
+				payload.GetPullRequest().GetNumber(), payload.GetRepo().GetFullName())
+		} else {
+			wh.logger.Errorf("Failed to get pull request from database %v", err)
```
meling, line 33: Maybe add PR number and repo as in the debug line above.

web/hooks/pull_request.go
```diff
@@ -0,0 +1,198 @@
...
+	"strconv"
+	"strings"
+
+	pb "github.com/autograde/quickfeed/ag"
+	"github.com/google/go-github/v35/github"
+	"gorm.io/gorm"
+)
+
+func (wh GitHubWebHook) handlePullRequestReview(payload *github.PullRequestReviewEvent) {
+	wh.logger.Debugf("Received pull request review event for pull request: %s, in repository: %s",
```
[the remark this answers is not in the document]
meling: Should it also include the PR number?

web/hooks/pull_request.go
meling: On second thought, the debug line at the top of the method should contain enough context.

web/hooks/pull_request.go
```diff
@@ -0,0 +1,198 @@
...
+		return
+	}
+
+	if err := wh.db.HandleMergingPR(pullRequest); err != nil {
+		wh.logger.Errorf("Failed to delete pull request from database %v", err)
+	}
+	wh.logger.Debugf("Pull request successfully closed for repository: %s", payload.GetRepo().GetFullName())
+}
+
+// TODO(Meling): I think it makes sense to have this function here, instead of in assignments/pull_request.go,
```
meling, line 116: It is fine for now at least. Not sure where is the best place for it atm. But I do see that its only side-effect is updating the database. It is also reading the database. The payload is GitHub specific, so moving it elsewhere would pollute other parts with the GitHub stuff.

meling: You can remove the TODO comment.

web/hooks/push.go
```diff
@@ -0,0 +1,286 @@
...
+	"github.com/google/go-github/v35/github"
+	"go.uber.org/zap"
+	"gorm.io/gorm"
+)
+
+func (wh GitHubWebHook) handlePush(payload *github.PushEvent) {
+	wh.logger.Debugf("Received push event for branch reference: %s (user's default branch: %s)",
+		payload.GetRef(), payload.GetRepo().GetDefaultBranch())
+
+	repos, err := wh.db.GetRepositories(&pb.Repository{RepositoryID: uint64(payload.GetRepo().GetID())})
```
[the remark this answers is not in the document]
meling: Another better suggestion is to do this:
```go
func (wh GitHubWebHook) getRepository(repoID int64) (*pb.Repository, error) {
	repos, err := wh.db.GetRepositories(&pb.Repository{RepositoryID: uint64(repoID)})
	if err != nil {
		return nil, fmt.Errorf("failed to get repository by remote ID %d: %w", repoID, err)
	}
	if len(repos) != 1 {
		return nil, fmt.Errorf("unknown repository: %d", repoID)
	}
	return repos[0], nil
}
```
And this:
```go
	repo, err := wh.getRepository(payload.GetRepo().GetID())
	if err != nil {
		wh.logger.Error(err)
		return
	}
	wh.logger.Debugf("Received push event for repository %v", repo)
```
And keep the course stuff separate as is now.

web/hooks/push.go
```diff
@@ -0,0 +1,286 @@
...
+	// TODO(meling): My idea is that when a teacher wants to assign a test to a specific task they will use the 'local' task name.
+	// For example, if a teacher has created the markdown file task-hello_world.md,
+	// they would do scores.AddWithTaskName(TestHelloWorld, "hello_world", max, weight).
+	// Do you concur with this approach?
+	// We could of course simply make teachers assign the global task name, but that seems somewhat counter-intuitive to me.
+
+	// TODO(espeland): Revise this when score task name format has been decided.
+	taskSum := results.TaskSum(task.LocalName())
+
+	// TODO(espeland): When the project is finished. Create a GitHub issue that states all places where
```
meling, line 141: Yes. You could mark it as a special TODO, e.g., `// TODO(Espeland) Update here for GitHub app.` and you could add the necessary context in a separate issue on the issue tracker.

web/hooks/github_test.go
oleespe: Again. I think this entire approach should be scrapped instead of fixed.

web/hooks/pull_request.go
```diff
@@ -0,0 +1,198 @@
...
+		wh.logger.Errorf("Failed to get user from database: %v", err)
+		return
+	}
+	reviewer, err := wh.db.GetEnrollmentByCourseAndUser(course.GetID(), user.GetID())
+	if err != nil {
+		wh.logger.Errorf("Failed to get reviewer from database: %v", err)
+		return
+	}
+
+	// Only if the review is from a course teacher, do we set the pull request to approved
```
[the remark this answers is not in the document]
oleespe: On line 19 we discard all events that are not approved reviews.

web/hooks/github_test.go
oleespe: Yes, it is only for testing

assignments/pull_requests.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/autograde/quickfeed/database"
+	"github.com/autograde/quickfeed/kit/score"
+	"github.com/autograde/quickfeed/scm"
+)
+
+var (
+	// These are used to track how many times someone has been assigned to review a pull request. They map as follows.
+	// teacherReviewCounter[courseID][userID] = count
+	// groupReviewCounter[groupID][userID] = count
+	teacherReviewCounter = make(map[uint64]map[uint64]int)
```
JosteinLindhom, line 18: Likely not critical, but this counter will reset if we have to restart the server for some reason.

Also, would it be an idea to decrement this counter for a user when a PR they are reviewing is done? This way we could assign PRs to TAs who have the least amount of outstanding PRs.
This would require us to store the ID of the reviewer in pull requests.

This could prove useful in scenarios where one TA is assigned PRs that are resolved quickly, while another TA is assigned some which requires more work relative to the other.
(However, these scenarios could arise not due to a high workload for the TA, but rather slow response time from students, or some other reason.)

We could also, if useful, use this to display an overview of PRs, and reviewers, in the frontend.

Implementing this will require database lookups.

oleespe: I guess it depends on how complex you want this feature to be, but I am definetly on board if so.

assignments/pull_requests.go
oleespe: Backticks meant I could no longer indent parts of the string, so I opted for a different solution.

JosteinLindhom:
@oleespe do you plan to resolve the remaining review comments, or shall we do it? If some of the comments have already been resolved please mark them as resolved.

assignments/pull_requests.go
```diff
@@ -91,35 +99,24 @@ func AssignReviewers(ctx context.Context, sc scm.SCM, db database.Database, cour
 // It is simple, and does not account for how many current review requests any user has.
 //
 // Returns an error if the list of users is empty.
```
meling, line 101: No longer returns an error.

web/hooks/github.go
oleespe: Moved it outside.
`Events: []string{"push", "pull_request", "pull_request_review"}` is used when creating hook, so what you say should be correct.

web/hooks/pull_request.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	// We make sure that the pull request is one that QF has a data record of
+	pullRequest, err := wh.db.GetPullRequest(&pb.PullRequest{
+		ExternalRepositoryID: uint64(payload.GetRepo().GetID()),
+		Number:               uint64(payload.GetPullRequest().GetNumber()),
+	})
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			wh.logger.Debugf("Ignoring pull request review event for non-managed pull request #%d, in %s",
+				payload.GetPullRequest().GetNumber(), payload.GetRepo().GetFullName())
```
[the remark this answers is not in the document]
oleespe: `GetFullName()` also includes the repository name, i.e., `qf101/oleespe-labs`, and I therefore feel it is more descriptive.
I changed the code at linke 16 to also use `GetFullName()`.

web/hooks/pull_request.go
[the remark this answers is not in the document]
oleespe: Added comment to further clarify.

assignments/pull_requests.go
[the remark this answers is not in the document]
oleespe: I will leave this as is, and you can change it the way you want.

web/hooks/push.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+}
+
+// runAssignmentTests runs the tests for the given assignment pushed to repo.
+func (wh GitHubWebHook) runAssignmentTests(assignment *pb.Assignment, repo *pb.Repository, course *pb.Course, payload *github.PushEvent) *score.Results {
+	runData := &ci.RunData{
+		Course:     course,
+		Assignment: assignment,
+		Repo:       repo,
+		// TODO(Espeland): If a student for some reason pushes to a remote branch with a different name than
```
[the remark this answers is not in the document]
oleespe: Yes, you are completely right. For some reason I thought that QF would copy the users local git repository, with its branch names, which in hindsight makes no sense.
Unforunate that I wrote about this in my report, but oh well.

oleespe:
Much of what remains relates to how the implementation is tested. As I have stated in other comments, I think the best option would be to remove all of it, and instead implement a better solution in a different pull request.

internal/qtest/test_helper.go
meling: Has been implemented via `ag/repo.go:UserName()`.

scm/github_test.go
```diff
@@ -222,8 +222,75 @@ func TestEditRepoIssue(t *testing.T) {
...
+func TestUpdateIssueComment(t *testing.T) {
+	qfTestOrg := scm.GetTestOrganization(t)
+	accessToken := scm.GetAccessToken(t)
+	s, err := scm.NewSCMClient(zap.NewNop().Sugar(), "github", accessToken)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// To use this test, the variables repository, body and commentID have to be set manually.
+	repository := "student-lab"
```
JosteinLindhom, line 285: I experimented with making these tests easier to use:

```go
// Helper function that creates a new issue, and returns a cleanup function that deletes the issue.
func createAndDeleteIssue(t *testing.T, s scm.SCM, org string, repo string) (*scm.Issue, func(issueNumber int, org string, repo string)) {
	t.Helper()
	issue, err := s.CreateIssue(context.Background(), &scm.CreateIssueOptions{
		Organization: org,
		Repository:   repo,
		Title:        "Test Issue",
		Body:         "Test Body",
	})
	if err != nil {
		t.Fatal(err)
	}

	return issue, func(issueNumber int, org string, repo string) {
		accessToken := scm.GetAccessToken(t)
		s := scm.NewGithubV4SCMClient(zap.NewNop().Sugar(), accessToken)
		if err := s.DeleteIssue(context.Background(), &scm.RepositoryOptions{Owner: org, Path: repo}, issueNumber); err != nil {
			t.Fatal(err)
		}
	}
}
```

Which can be used in most of the tests similar to this:

```go
func TestCreateIssueComment(t *testing.T) {
	qfTestOrg := scm.GetTestOrganization(t)
	accessToken := scm.GetAccessToken(t)
	qfTestUser := scm.GetTestUser(t)
	s, err := scm.NewSCMClient(zap.NewNop().Sugar(), "github", accessToken)
	if err != nil {
		t.Fatal(err)
	}

	body := "Test"
	opt := &scm.IssueCommentOptions{
		Organization: qfTestOrg,
		Repository:   ag.StudentRepoName(qfTestUser),
		Body:         body,
	}

	issue, delete := createAndDeleteIssue(t, s, opt.Organization, opt.Repository)
	defer delete(issue.IssueNumber, opt.Organization, opt.Repository)

	_, err = s.CreateIssueComment(context.Background(), issue.IssueNumber, opt)
	if err != nil {
		t.Fatal(err)
	}
}
```

This will create a new issue on the `QF_TEST_ORG` for the user `QF_TEST_USER`, and delete the issue when the test returns.

meling: Nice one. I think you can even simplify this further by having the returned function reuse the passed in `s` if you use the `scm.GithubV4SCM` type instead of `scm.SCM`:

```go
	return issue, func(issueNumber int, org string, repo string) {
		if err := s.DeleteIssue(context.Background(), &scm.RepositoryOptions{Owner: org, Path: repo}, issueNumber); err != nil {
			t.Fatal(err)
		}
	}
```

web/hooks/pull_request.go
```diff
@@ -0,0 +1,196 @@
...
+	repo := repos[0]
+	if !repo.IsGroupRepo() {
+		wh.logger.Debugf("Ignoring pull request opened event for non-group repository: %s", payload.GetRepo().GetFullName())
+		return
+	}
+	wh.createPullRequest(payload, repo)
+}
+
+func (wh GitHubWebHook) handlePullRequestClosed(payload *github.PullRequestEvent) {
+	wh.logger.Debugf("Received pull request closed event for repository: %s, in organization: %s",
```
meling, line 90: Could this be replaced with only this one:
```go
	wh.logger.Debugf("Received pull request closed event for repository: %s", payload.GetRepo().GetFullName())
```

web/hooks/pull_request.go
```diff
@@ -0,0 +1,196 @@
...
+	// If this method is to handle different types states, this check must be moved to here.
+	if reviewer.IsTeacher() {
+		pullRequest.SetApproved()
+		wh.db.UpdatePullRequest(pullRequest)
+		wh.logger.Debugf("Pull request successfully approved for repository: %s", payload.GetRepo().GetFullName())
+	}
+}
+
+func (wh GitHubWebHook) handlePullRequestOpened(payload *github.PullRequestEvent) {
+	wh.logger.Debugf("Received pull request opened event for repository: %s, in organization: %s",
```
meling, line 69: Could this be replaced with only this one:
```go
	wh.logger.Debugf("Received pull request opened event for repository: %s", payload.GetRepo().GetFullName())
```

scm/github_test.go
```diff
@@ -119,111 +122,205 @@ func TestCreateHook(t *testing.T) {
...
-	ctx := context.Background()
+	if gotIssue.Title != opt.Title || gotIssue.Body != opt.Body {
+		t.Fatalf("scm.TestUpdateIssue() want (title: %s, body: %s), got (title: %s, body: %s)", opt.Title, opt.Body, gotIssue.Title, gotIssue.Body)
+	}
+}
 
-	opt := &scm.RepositoryOptions{
-		Owner: qfTestOrg,
-		Path:  repo,
+func TestRequestReviewers(t *testing.T) {
```
meling, line 233: @JosteinLindhom Did you plan to update this test as well? Seem like it could also benefit from using the `ag.StudentRepoName(qfTestUser)` thing?? Not sure.

JosteinLindhom: This one is slightly more complicated. 
To run this test I think we need to:

- Have an open pull request on `QF_TEST_ORG/QF_TEST_USER-labs`, and its number.
- Implement `RemoveRequestedReviewers` to clean up for subsequent runs.

scm/githubv4.go
```diff
@@ -0,0 +1,83 @@
...
+			logger: logger,
+			client: client,
+			token:  token,
+		},
+		clientV4:     githubv4.NewClient(httpClient),
+		bypassClient: client,
+	}
+}
+
+func (s *GithubV4SCM) DeleteIssue(ctx context.Context, opt *RepositoryOptions, issueNumber int) error {
```
oleespe, line 36: With the inclusion of a new `DeleteIssue()` function, we can in theory delete issues, instead of marking them as deleted and closing them, whenever a teacher deletes a task.

oleespe:
Now also resolves #651

ag/task.go
oleespe: No longer valid

assignments/tasks_test.go
oleespe: No longer valid

assignments/pull_requests_test.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+		students = students[:len(students)-1]
+	}
+}
+
+// TestPublishFeedbackComment tests creating a feedback comment on a pull request, with the given result.
+func TestPublishFeedbackComment(t *testing.T) {
+	qfTestOrg := scm.GetTestOrganization(t)
+	accessToken := scm.GetAccessToken(t)
+	repository := scm.GetTestUser(t)
```
oleespe, line 73: Does this make sense? Here the `GetTestUser()` function is used to create the repository variable, meaning that it would be of the type `oleespe-labs`, while in other tests it is used together with the `ag.StudentRepoName()` function to generate a repository name.

internal/qtest/test_helper.go
oleespe: I suppose this has been implemented as `GetTestUser()`

assignments/pull_requests_test.go
oleespe: You created a todo assigned to yourself for this, so closing this comment.

assignments/pull_requests_test.go
meling: You are right. This is a bug. The `GetTestUser()` must be used together with `StudentRepoName()` to get the repository.

PR #646, opened 2022-05-18
Convert state to objects

### LINKED ISSUE #631: When accepting student to course the member list does not update with secondary problem
When accepting a student to a course the member list does not update. Even after several minutes while composing this issue.

- Note that, I'm in the USA, so it could be due to a timeout that normally doesn't trigger.
- Another root cause could of course be packets being dropped on either side of the connection (I'm on a hotel WiFi, but we should be resilient to such problems...)
- Reloading fixes the problem.

However without reloading, if clicking again on accept, you will get a prompt asking if you are sure you I want to promote the student to teacher. This should not be the intended behavior.

The same problem also appears when promoting student to teacher. The member list does not update. In fact, when clicking again on `promote` actually demotes the teacher back to student (because the list wasn't reloaded).

### LINKED ISSUE #643: Sort test listing by name, score or weight
It would be useful to be able to dynamically switch between sorting the tests by name, score or weight.

The default should be sort by name.

JosteinLindhom:
This PR changes the state structure of the frontend to use the less restrictive `AsObject` type variants of proto messages.

Fixes #631.
Fixes #643.

## Motivation

The motivation for this change originated due to how proto messages work.

When we receive an `Enrollment` message from the backend, this type has `getters` and `setters` for all its fields, as defined [here](https://github.com/quickfeed/quickfeed/blob/640e77a7cfea57f49924abc84816ff4a4ab08a45/ag/ag.proto#L109). However, if one of the fields of `Enrollment` are not set when sent from the backend, that field is not initialized when received in the frontend. Using a `getter` on an uninitialized field triggers a mutation of the message by setting the field to its zero value.

The state in the frontend is immutable. Any state mutation must be done through *actions*.

A very simplified example to illustrate the issue as follows:
```tsx
// Receive an enrollment from the backend
const enrollment = grpc.getEnrollment()
// Assign the enrollment to our state
// The returned enrollment does not have the `ID` field set
state.enrollment = enrollment

const Component = () => {
    const state = useState()
    return <h1>{state.enrollment.getId()}</h1> // Accessing `enrollment.getId()` triggers a `enrollment.setId(0)`
}
```
In the above snippet, the call to `enrollment.getId()`, where the `ID` field is not set, triggers a mutation to initialize the field to its zero value. Since this mutation occurs in our component, rather than within an `action`, we get an error that prevents the component from rendering. We are unable to determine whether or not a field is initialized without actually checking.

Some solutions to this Schroedinger's field situation include:
- Using a `getter` on every field before adding it to state, to check if the fields are initialized, triggering `setters` automatically if it is not.
- Wrapping every `getter` call in our components in a `try-catch` block.
- Converting messages to their object representation types, `AsObject`.

## Resolving the State Issue

Rather than storing received messages "as is", this PR modifies the state to store them as `<Message>.AsObject`.
```tsx
// State definition
type State = {
    self: User.AsObject // vs. self: User
}
// Assigning to state
const user = grpc.getUser()
// Convert user to AsObject and assign to state
state.self = user.toObject() // vs. state.self = user

// Use in components
const Component = () => {
    const state = useState()
    return <h1>{state.self.name}</h1> // vs. {state.self.getName()}
}
```

We access values from an `AsObject` object through named properties.
The named properties of `User.AsObject`:
```ts

export namespace User {
  export type AsObject = {
    id: number,
    isadmin: boolean,
    name: string,
    studentid: string,
    email: string,
    avatarurl: string,
    login: string,
    remoteidentitiesList: Array<RemoteIdentity.AsObject>,
    enrollmentsList: Array<Enrollment.AsObject>,
  }
}
```
By default, converting a `User` to `User.AsObject` initializes all fields in the object to either the respective `User` field, or it's zero value if none exists.
Any property accesses on `User.AsObject` are strictly read, unless we explicitly assign a new value to it.

```ts
const user = grpc.getUser() // name not set
state.user = user
state.userObject = user.toObject()
// in component
state.user.getName() // Error, attempting to mutate state, render fails
state.userObject.name // ""
state.userObject.name = "Name" // Error, attempting to mutate state
```

The example shows that the changes introduced in this PR now yield desired behaviour in our components. Reading from state now *only reads from state*, without any sneaky mutations occuring if the field was not set. Writing to state also produces an error, as desired.

### Readability

In some cases a message can contain a field which is another message, such as `Enrollment.getUser()`, which returns a `User`. The `AsObject` representation of `Enrollment` marks this field as optional, indicating that it may in fact be `undefined`.

```ts
// Current
data.push(enrollment.hasUser() ? (enrollment.getUser() as User).getName() : "")
// enrollment.hasUser() does not inform TypeScript that enrollment.getUser() is not undefined, hence the type assertion
// Along with potentially attempting to mutate the state, neither does enrollment.getUser()

// This PR
data.push(enrollment.user ? enrollment.user.name : "")
```

As shown in the example, we do no longer have to what should have been unnecessary type assertions.
Additionally, getting rid of `getters` on our state objects improves the readability of our code. We can now access object properties using dot notation with their named properties.

### Cloning state objects

Previously, cloning an object in state could be done using Overminds `json` function, or a proto message function `Message.clone()`. There are, however, issues with both.
The `json()` function only does a shallow clone, keeping references for any nested arrays. Modifying any nested arrays cause mutation errors.
Using the `Message.clone()` function works if, and only if, *all* fields for a message, and any child messages, are set. If not it will attempt to set these unset values, causing mutation errors.

The changes in this PR now allow us to do proper deep clones of objects.
```ts
// Helpers.ts ProtoConverter class
public static clone<T>(obj: T): T {
    return JSON.parse(JSON.stringify(obj)) as T
}
```
This function let us clone all `AsObject` types without mutating the state.

### Mocked gRPC Manager

This PR introduces a [mocked gRPC manager](https://github.com/quickfeed/quickfeed/blob/convert-state-to-objects/dev/src/MockGRPCManager.ts). Adding a mocked gRPC manager allow us to run the frontend without having the QuickFeed backend also running. Additionally, by using this manager, we can easier write tests for functionality that would require exchanging information with the backend.

The mocked gRPC manager comes pre-populated with relevant data.

`NOTE: In its current form it does not mimick the expected behaviour of the backend 1-1.`

### Development Mode

If the `mode` in `webpack.config.js` is set to either `development` or `test`, and the hostname is `localhost`, the mocked gRPC manager will be used:

```ts
export const grpcMan: GrpcManager | MockGrpcManager = (() => {
    let grpcMan: GrpcManager | MockGrpcManager
    if ((process.env.NODE_ENV === "development" || process.env.NODE_ENV === "test") && window.location.hostname === "localhost") {
        // If in development or test mode, and the hostname is localhost, use the MockGrpcManager.
        grpcMan = new MockGrpcManager()
    } else {
        // Otherwise, use the real gRPC manager.
        grpcMan = new GrpcManager()
    }
    return grpcMan
})()
```

When compiling our code we can add an argument: `webpack --mode=[production|development]` to toggle development mode off or on.

### Running the frontend as a standalone application

The new script [`run.sh`](https://github.com/quickfeed/quickfeed/blob/convert-state-to-objects/dev/run.sh) can be run to serve the QuickFeed frontend using [`webpack-dev-server`](https://webpack.js.org/configuration/dev-server/).

Serving the frontend through `webpack-dev-server` will utilize the mocked gRPC manager to mock server communication. Additionally, any changes to code will instantly be reflected in any browser used to view the frontend.


dev/src/DevelopmentMode.tsx
```diff
@@ -0,0 +1,43 @@
+import React from "react"
+import { MockGrpcManager } from "./MockGRPCManager"
+import { useActions, useGrpc } from "./overmind"
+
+
+// DevelopmentButtons contain functionality to save the current state
```
meling, line 6: Should this be `DevelopmentMode`?

dev/src/MockGRPCManager.ts
```diff
@@ -106,8 +116,27 @@ export class MockGrpcManager {
     public createCourse(course: Course): Promise<IGrpcResponse<Course>> {
         let data: Course | null = null
         const found = this.courses.getCoursesList().find(c => c.getId() === course.getId())
-        if (!found || !this.currentUser.getIsadmin()) {
+        const isAdmin = this.currentUser?.getIsadmin()
+        const user = this.currentUser
+        if (!found && user && isAdmin) {
+            // TODO: This is not a guaranteed unique ID
```
meling, line 122: What would be the solution to producing a unique ID here and other places?

I'm not sure I quite understand why the IDs must/should be unique, but could a solution be to use a map from some string key (e.g., `CourseID`) that points to another map of already used IDs that a `generateID("CourseID")` function should check before returning a new ID? (Or something along those lines???)

dev/src/components/CourseBanner.tsx
```diff
@@ -10,16 +10,18 @@ const CourseBanner = (): JSX.Element => {
     const actions = useActions()
 
     const enrollment = state.enrollmentsByCourseID[getCourseID()]
-    const style = isVisible(enrollment) ? "fa fa-star" : "fa fa-star-o"
+    const icon = isVisible(enrollment) ? "fa fa-star" : "fa fa-star-o"
     return (
         <div className="banner jumbotron">
```
meling, line 15: This was quite deep HTML nesting. Can these be collapsed?

dev/src/components/CourseCard.tsx
```diff
@@ -20,13 +20,13 @@ const CardColor = [
 const CourseCard = ({ course, enrollment }: CardProps): JSX.Element => {
     const actions = useActions()
     const history = useHistory()
-    const status = enrollment.getStatus()
+    const status = enrollment.status
 
     return (
         <div className="col-sm-4">
```
meling, line 26: This was quite deep HTML nesting. Can these be collapsed?

dev/src/components/CourseUtilityLinks.tsx
```diff
@@ -39,11 +39,11 @@ const CourseUtilityLinks = (): JSX.Element => {
                     Course Info
                 </a>
 
-                {enrollment.hasGroup() ?
+                {state.hasGroup(courseID) ?
                     <Link to={"/course/" + courseID + "/group"} className="list-group-item list-group-item-action">
                         View Group
                     </Link>
-                    : <Link to={"/course/" + courseID + "/group/create"} className="list-group-item list-group-item-action list-group-item-success">
+                    : <Link to={"/course/" + courseID + "/group"} className="list-group-item list-group-item-action list-group-item-success">
```
meling, line 46: No `/group/create` anymore?

dev/src/components/Groups.tsx
```diff
@@ -12,83 +12,76 @@ const Groups = (): JSX.Element => {
...
-                                        <a href={`https://github.com/${enrol.getUser()?.getLogin()}`} target="_blank" rel="noopener noreferrer">{enrol.getUser()?.getName()}</a>
-                                        {index >= group.getEnrollmentsList().length - 1 ? "" : ", "}
-                                    </span>
-                                )}
-                        </div>
-                    </td>
-                    <GroupButtons group={group} />
-                </tr>
-            </>
+            <tr hidden={groupSearch(group)}>
```
meling, line 53: Deep HTML nesting

dev/src/components/teacher/EditBenchmark.tsx
```diff
@@ -1,35 +1,40 @@
...
 
     const [editing, setEditing] = useState<boolean>(false)
     const [add, setAdd] = useState<boolean>(benchmark ? false : true)
-    const [newBenchmark, setNewBenchmark] = useState<GradingBenchmark>(new GradingBenchmark().setAssignmentid(assignment.getId()))
-    const bm = json(benchmark) ?? newBenchmark
-    const copy = json(benchmark)?.cloneMessage()
+
+    // Clone the criterion, or create a new one if none was passed in
+    const bm = benchmark
+        ? Converter.clone(benchmark)
```
meling, line 15: There were a few other uses of the ternary operator: `? :` where the `?` was on the line above, while the `:` was like this example. I prefer this style where the `?` and `:` line up on top of each other.

dev/src/MockGRPCManager.ts
JosteinLindhom: The current solution using `<Array>.length + 1` causes issues if an element in the array is deleted, and a new element is inserted, example:

`Courses -> [{courseID: 1}, {courseID: 2}, {courseID: 3}]`.

Delete course with ID == 1 -> `[{courseID: 2}, {courseID: 3}]`
Insert new course (ID is set to `courseList.length + 1`) -> `[{courseID: 2}, {courseID: 3}, {courseID: 3]`

Which will cause issues when mapping objects such as assignments to their related course by their course ID.

Your suggested solution sounds good.

dev/src/components/CourseUtilityLinks.tsx
JosteinLindhom: The `/group` route now renders, for the given course:
1. Teacher group view if the user is a teacher
2. Group creation view if a student is not in a group
3. Student group view if a student is in a group

dev/src/MockGRPCManager.ts
meling: Understand. Suggest to implement my solution to avoid problems down the line. I think it will be possible to do it with a separate function that works across different IDs.

meling: It might be a good thing to add a unit test for such a `generateID()` function.

dev/src/MockGRPCManager.ts
```diff
@@ -72,7 +92,9 @@ export class MockGrpcManager {
     private organizations: Organizations
     private submissions: Submissions
     private templateBenchmarks: GradingBenchmark[]
-
+    // idMap is a map of auto incrementing IDs
+    public idMap: Map<string, number> = new Map<string, number>()
+    public e: typeof Generate = Generate
```
meling, line 97: Not clear what this `e` is. Perhaps name it with a longer name or add a comment?

dev/src/MockGRPCManager.ts
```diff
@@ -120,7 +145,7 @@ export class MockGrpcManager {
         const user = this.currentUser
         if (!found && user && isAdmin) {
             // TODO: This is not a guaranteed unique ID
```
meling, line 147: Remove TODO?

dev/src/MockGRPCManager.ts
```diff
@@ -131,7 +156,7 @@ export class MockGrpcManager {
             enrollment.setUserid(user.getId())
             enrollment.setStatus(Enrollment.UserStatus.TEACHER)
             // TODO: This is not a guaranteed unique ID
```
meling, line 158: Remove TODO

dev/src/MockGRPCManager.ts
```diff
@@ -356,7 +382,7 @@ export class MockGrpcManager {
         request.setName(name)
         request.setCourseid(courseID)
         // TODO: Not a unique ID
```
meling, line 384: Remove TODOs

dev/src/__tests__/updateAdmin.test.tsx
```diff
@@ -1,37 +1,33 @@
...
 
-    it("If user does not confirm, do not make any changes", () => {
-        const user3 = new User().setId(3).setName("Test User3").setIsadmin(true).toObject()
-        const mockedOvermind2 = createOvermindMock(config, (state) => {
-            state.allUsers = [user3]
-        })
-        window.confirm = jest.fn(() => false)
-        updateAdmin(mockedOvermind2, user3)
-        expect(user3.isadmin).toBe(true)
+    const updateAdminTests: { desc: string, user: User.AsObject, confirm: boolean, want: boolean }[] = [
```
meling, line 7: Nice rewrite!!

dev/src/components/CourseCard.tsx
```diff
@@ -22,31 +23,38 @@ const CourseCard = ({ course, enrollment }: CardProps): JSX.Element => {
     const history = useHistory()
     const status = enrollment.status
 
+    const CourseEnrollmentButton = (): JSX.Element => {
```
meling, line 26: Nice rewrite!

dev/src/overmind/state.tsx
```diff
@@ -271,8 +275,8 @@ export const state: State = {
 
         const m = state.sortAscending ? -1 : 1
```
meling, line 276: @JosteinLindhom Would `sortOrder` be a good replacement for `m`?

dev/src/components/teacher/EditCriterion.tsx
```diff
@@ -1,36 +1,40 @@
...
 
     const [editing, setEditing] = useState<boolean>(false)
     const [add, setAdd] = useState<boolean>(criterion ? false : true)
-    const [newCriterion, setNewCriterion] = useState<GradingCriterion>(new GradingCriterion().setBenchmarkid(benchmarkID))
-    // Clone the criterion to use as backup in case of cancel
-    const copy = json(criterion)?.cloneMessage()
-    const c = json(criterion) ?? newCriterion
+
+    // Clone the criterion, or create a new one if none was passed in
+    const c = criterion
```
meling, line 14: Could this be named `newCriterion` instead of `c`? (Since `c.description` is not obvious when used later.)

An alternative could be to rename the input argument to `originalCriterion` and `c` to `criterion`.

dev/src/Helpers.ts
```diff
@@ -329,35 +328,35 @@ export enum SubmissionSort {
 }
 
 /** Sorting */
-const enrollmentCompare = (a: Enrollment, b: Enrollment, sortBy: EnrollmentSort, descending: boolean): number => {
+const enrollmentCompare = (a: Enrollment.AsObject, b: Enrollment.AsObject, sortBy: EnrollmentSort, descending: boolean): number => {
     const m = descending ? -1 : 1
```
meling, line 332: @JosteinLindhom Would `sortOrder` be a good replacement for `m`? (same issue as in `state.tsx`)

meling (approved):

PR #664, opened 2022-07-04
Fixed profile form bug

JosteinLindhom:
A bug was introduced in #646, which would not let users save their information in the profile form.

This PR fixes that bug.

### DIFF
--- a/dev/src/components/profile/ProfileForm.tsx
+++ b/dev/src/components/profile/ProfileForm.tsx
@@ -1,4 +1,4 @@
-import React, { Dispatch, SetStateAction, useMemo } from "react"
+import React, { Dispatch, SetStateAction, useMemo, useState } from "react"
 import { hasEnrollment } from "../../Helpers"
 import { useActions, useAppState } from "../../overmind"
 import FormInput from "../forms/FormInput"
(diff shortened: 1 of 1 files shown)


dev/src/components/profile/ProfileForm.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { Dispatch, SetStateAction, useMemo } from "react"
+import React, { Dispatch, SetStateAction, useMemo, useState } from "react"
```
meling, line 1: `useMemo` is no longer used. (DeepSource)... Other DeepSource (anti-pattern) issues are better resolved in a separate PR.

meling (changes requested):

dev/src/components/profile/ProfileForm.tsx
JosteinLindhom: Just realized that we can add this to `tsconfig.json` to be notified of such issues:

```
    "noUnusedLocals": true,
    "noUnusedParameters": true
```

meling: Yes, please do!

PR #667, opened 2022-07-05
Webpack optimization

### LINKED ISSUE #660: Performance optimizations of frontend
Webpack reports this:
```shell
% make ui
WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets:
  bundle.js (709 KiB)

WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  main (709 KiB)
      bundle.js


WARNING in webpack performance recommendations:
You can limit the size of your bundles by using import() or require.ensure to lazy load some parts of your application.
For more info visit https://webpack.js.org/guides/code-splitting/
```

JosteinLindhom:
Resolves #660.

This PR introduces changes to the configuration of Webpack.

Currently, our transpiled and stored in a single bundle file. This bundle includes all our code (components, proto definitions, dependencies, etc.). If we introduce a change, regardless of size, the user has to redownload the entire bundle.

In this PR some changes have been introduced to minimize the problem of redundant fetching, including:
- All bundle files now include a hash based on file contents. This forces the client to download any bundles that have changed (cache busting). Previously, users would have to force their browser to fetch the newest bundle, or wait until the cache expired.
- Multiple bundles are now generated. Notably, one bundle for our component code, one for proto code, and another for state code. 
- In addition to the above, one bundle is generated for each included dependency.

These changes do *not* necessarily reduce the load time of visiting the website for the first time. It will, however, reduce the total download size as changes to our code is introduced. Users being served outdated versions of our frontend should also no longer be an issue.

(Example: our current bundle is some ~700kb. Splitting it into several bundles, where our component specific bundle is ~120kb,  reduces the download size by ~80% when introducing changes to our component code)

A new file, `index.tmpl.html` is added. Webpack will use this file as a base, and inject all emitted bundles into this file. The file is saved as `index.html` in `/public/assets`.

Finally, additional config is added for the `webpack-dev-server` to work with the changes described above. Due to these changes the `run.sh` is no longer needed.


public/webpack.config.js
```diff
@@ -31,16 +68,46 @@ module.exports = {
...
+
     module: {
         rules: [
             // All files with a '.ts' or '.tsx' extension will be handled by 'ts-loader'.
             { test: /\.tsx?$/, loader: "ts-loader" },
 
             // All output '.js' files will have any sourcemaps re-processed by 'source-map-loader'.
             { enforce: "pre", test: /\.js$/, loader: "source-map-loader" },
-            { test: /\.css$/i,use: ["style-loader", "css-loader"],},
+            { test: /\.css$/i, use: ["style-loader", "css-loader"], },
```
meling, line 110: Should there be a `,` at the end before the `}`?

public/webpack.config.js
```diff
@@ -51,4 +118,4 @@ module.exports = {
         "react": "React",
         "react-dom": "ReactDOM",
     },
-};
\ No newline at end of file
+};
```
meling, line 121: Should there be `;` here? I think our typescript formatter would remove them. But maybe they are necessary for javascript?

PR #675, opened 2022-07-06
Replace Enzyme with React Testing Library

JosteinLindhom:
- Removed Enzyme in favor of React Testing Library
- Rewrote Enzyme tests using RTL instead


meling (approved):
Looks good to me!!

PR #679, opened 2022-07-07
Split protofile

### LINKED ISSUE #437: Reorganize proto file into separate files
Since the `ag.proto` file is becoming quite big as pointed out in #290, we should reorganize it into smaller proto-files and split the RPCs into multiple services. Further, we should consider if we can expose the remaining REST endpoints as RPC methods using the [buf](https://buf.build) tool.

_Originally posted by @meling in https://github.com/autograde/quickfeed/issues/290#issuecomment-749760747_

_Originally posted by @meling in https://github.com/autograde/quickfeed/issues/290#issuecomment-783735766_

0xf8f8ff:
This PR splits `qf.proto` into three separate files.
- `types/types.proto` contains all message types that are saved directly to the database.
- `types/requests.proto` contains message types only used in API calls.
- the original `qf.proto` only contains the service definition.

Resolves #437.


qf/types/requests.proto
```diff
@@ -0,0 +1,172 @@
+syntax = "proto3";
+package types;
+option go_package   = "github.com/quickfeed/quickfeed/qf/types";
+option swift_prefix = "";
+
+import "qf/types/types.proto";
+
+// Message types that are used as requests and responses in API calls.
+
+message SubmissionLink {
```
meling, line 10: This should be in the `types.proto` file (it is only used by the database ; not used in `qf.proto`).

qf/types/requests.proto
```diff
@@ -0,0 +1,172 @@
...
+    types.Assignment assignment = 1;
+    types.Submission submission = 2;
+}
+
+message EnrollmentLink {
```
meling, line 15: This should be in the `types.proto` file (it is only used by the database ; not used in `qf.proto`).

qf/types/requests.proto
```diff
@@ -0,0 +1,172 @@
...
+    uint64 groupID = 1;
+}
+
+message GroupRequest {
+    uint64 userID   = 1;
+    uint64 groupID  = 2;
+    uint64 courseID = 3;
+}
+
+message Provider {
```
meling, line 48: This actually don't seem to be used at all and can probably be removed. There is only a `IsValid()` method on this, but otherwise it doesn't seem like it is being used anymore.

qf/types/requests.proto
0xf8f8ff: Hmm, I thought the idea was to have messages saved directly to the database in a separate file (types.proto) and group the rest in the `requests.proto`. `SubmissionLink` is never saved to the database.

On the other hand, I also put `Courses`, `Enrollments' and others in `types.proto` while they also never go to the database. So yeah, `SubmissionLink` belongs in the `types.proto`.

0xf8f8ff: Moved.

qf/types/requests.proto
0xf8f8ff: Right. Removed it.

meling (approved):
Looks good to me!!

qf/types/requests.proto
meling: I was thinking about the separation more in terms of which messages are used as request messages in the RPCs. But let's review the decision again once we have some more experience with this separation. Maybe we learn something new. 

I know that some of the messages, such as `Courses` and `Enrollments` are used both for RPCs and stored in database... That's desirable. Perhaps we learn that it would make more sense to have another file with messages that are shared between both... For now I think we have at least some good starting point for the separation we wanted...

Issue #663, opened 2022-07-04
Add support for SCM environment variable


meling:
We currently provide a few constructors for creating `scm.SCM` instances. The main one requires that `"github"` or `"gitlab"` is passed as an argument. However, we should simply remove this argument and instead read from an environment variable `SCM` which to use. Each deployment of quickfeed should anyway only support one SCM.

If not defined, the default should obviously be `github`.

Any objections?

Issue #480, opened 2021-07-29
Fix webhook secret handling


meling:
We read from the `WEBHOOK_SECRET` environment variable but don't actually set it when starting the quickfeed server.

Perhaps the webhook secret should instead be generated and recorded in the database. There are two alternatives:
- generate when first initializing the quickfeed database (webhook secret per quickfeed instance)
- generate when creating a new course (webhook secret per course)

@JosteinLindhom Did you look into this before?? I seem to recall you mentioned something related to this before.
@0xf8f8ff Do you have any thoughts on this?

0xf8f8ff:
I like the per-course approach best, should be easy to implement as well by adding some `WHSecret` column to the `Course` table.

meling:
> I like the per-course approach best, should be easy to implement as well by adding some `WHSecret` column to the `Course` table.

Yes. I agree.

0xf8f8ff:
One issue I see with course-level secrets: we validate hook event payload against a secret before getting access to the payload, but we have to get access to the payload first to know which course the event is coming from and fetch the secret from the database.

PR #692, opened 2022-07-14
Removing envoy proxy.

0xf8f8ff:
- The envoy proxy has been replaced by the [Go server wrapper](https://github.com/improbable-eng/grpc-web/tree/master/go/grpcweb) that will multiplex HTTP and gRPC requests to the same port.
- `echo` package dependency has been replaced by the direct use of the `net/http` package.
- `goth/gothic` dependencies have been replaced by the `oauth2` package.
- HTTP endpoints have been simplified as follows:

```
/auth/<provider>
/auth/callback/<provider>
/logout
/hook/<provider>
```

The server can run with or without envoy. 

With envoy, the proxy will manage the TLS session.

Envoy is no longer needed to run the server locally. The gRPC server must maintain its own TLS session in this case. To run the server locally (without envoy proxy), pass the `-dev` command line argument on start.
 
Note: because of the changed callback URL, QuickFeed's OAuth app has to be updated to make it compatible with the changes introduced by this PR. The old callback URL `/auth/github/callback` should be changed to `/auth/callback/github`.


go.mod
```diff
@@ -12,8 +12,8 @@ require (
 	github.com/gorilla/sessions v1.2.1
 	github.com/gosimple/slug v1.12.0
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
+	github.com/improbable-eng/grpc-web v0.15.0
 	github.com/joho/godotenv v1.4.0
-	github.com/labstack/echo-contrib v0.12.0
 	github.com/labstack/echo/v4 v4.7.2
```
meling, line 17: Still here; seems to be a single method `login()` on the `testStore` still depending on echo in `web/auth/auth_test.go:359`. The `login()` method is not used by any code at the moment, but not sure tests are going to need it later... 

You can use:
```sh
go mod why github.com/labstack/echo/v4
```
To find where a dependency is being used.

main.go
```diff
@@ -98,27 +102,50 @@ func main() {
 	}
 
 	qfService := web.NewQuickFeedService(logger, db, scms, bh, runner)
-	go web.New(qfService, *public, *httpAddr)
-
-	lis, err := net.Listen("tcp", *grpcAddr)
+	// TODO(vera): harcoded or command line flags?
```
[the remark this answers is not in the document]
meling: Or maybe we should have a package `internal/env` that has methods for fetching such things, e.g.,:
```go
const (
  certFile = "QUICKFEED_CERT_FILE"
  certKey = "QUICKFEED_CERT_KEY"
)

func CertFile() (string, error) {
  cFile := os.Getenv(certFile)
  if cFile == "" {
    return "", fmt.Errorf("required environment variable %s is not set", certFile)
  }
  return cFile, nil
}
```

main.go
```diff
@@ -98,27 +102,50 @@ func main() {
...
 	go func() {
 		if err := httpServer.ListenAndServe(); err != nil {
-			log.Fatal("Unable to start a http server.")
+			log.Fatalf("failed to start a http server: %v\n", err)
 		}
 	}()
 
-	qf.RegisterQuickFeedServiceServer(grpcServer, qfService)
-	if err := grpcServer.Serve(lis); err != nil {
+	// qf.RegisterQuickFeedServiceServer(grpcServer, qfService)
```
meling, line 137: Remove commented lines (this one is also on line 116)

main.go
```diff
@@ -98,27 +102,50 @@ func main() {
...
+	muxServer := &http.Server{
+		Handler:      router,
+		Addr:         "127.0.0.1:8080",
+		WriteTimeout: 2 * time.Minute,
+		ReadTimeout:  2 * time.Minute,
+	}
+	if err := muxServer.ListenAndServeTLS(certFile, certKey); err != nil {
 		log.Fatalf("failed to start grpc server: %v\n", err)
 	}
+	// if err := grpcServer.Serve(lis); err != nil {
```
meling, line 148: Needed later?

main.go
```diff
@@ -98,27 +102,50 @@ func main() {
...
+
 	// Create a HTTP server for prometheus.
 	httpServer := &http.Server{
 		Handler: promhttp.HandlerFor(reg, promhttp.HandlerOpts{}),
 		Addr:    fmt.Sprintf("0.0.0.0:%d", 9097),
 	}
 	go func() {
 		if err := httpServer.ListenAndServe(); err != nil {
-			log.Fatal("Unable to start a http server.")
+			log.Fatalf("failed to start a http server: %v\n", err)
```
meling, line 133: Note that there is a difference between errors that are returned from a function (they should use lower-case initial letter), but when actually logging a message we should use Capitalization. See [here for details](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

public/src/GRPCManager.ts
```diff
@@ -382,4 +382,4 @@ export class GrpcManager {
                 + resp.status.getCode() + ": " + resp.status.getError())
         }
     }
-}
+}
```
meling, line 385: newline

main.go
```diff
@@ -57,11 +56,11 @@ var reg = prometheus.NewRegistry()
...
 	var (
-		baseURL  = flag.String("service.url", "", "base service DNS name")
-		dbFile   = flag.String("database.file", "qf.db", "database file")
-		public   = flag.String("http.public", "public", "path to content to serve")
-		httpAddr = flag.String("http.addr", ":8081", "HTTP listen address")
-		grpcAddr = flag.String("grpc.addr", ":9090", "gRPC listen address")
+		baseURL = flag.String("service.url", "", "base service DNS name")
+		dbFile  = flag.String("database.file", "qf.db", "database file")
+		public  = flag.String("http.public", "public", "path to content to serve")
+		// httpAddr = flag.String("http.addr", ":8081", "HTTP listen address")
```
meling, line 62: remove ; make env if needed.

web/auth/auth.go
```diff
@@ -41,12 +44,37 @@ const (
 	Redirect = "redirect"
 )
 
+// Temporary solution. Will be removed when sessions replaced by JWT.
```
meling, line 47: okay. could you be more precise how much will be replaced; otherwise it is a bit difficult to know how much of this I should "review" properly.

web/auth/auth.go
```diff
@@ -41,12 +44,37 @@ const (
...
+	sessionStore.Options.Secure = true
+}
+
 // UserSession holds user session information.
 type UserSession struct {
 	ID        uint64
 	Providers map[string]struct{}
 }
 
+func authenticationError(logger *zap.SugaredLogger, w http.ResponseWriter, err string) {
```
meling, line 73: Please pass in `err error` instead of `string`. Callers should use `fmt.Errorf()` instead of `fmt.Sprintf()` or `errors.New()` if the error is simple text.

web/auth/auth.go
```diff
@@ -177,145 +144,148 @@ func sessionData(session *sessions.Session) string {
...
-		logger.Debugf("qv: %v", qv)
-		redirect := extractRedirectURL(r, Redirect)
-		logger.Debugf("redirect: %v", redirect)
-		// TODO: Add a random string to protect against CSRF.
-		qv.Set(State, strconv.Itoa(teacher)+redirect)
-		logger.Debugf("State: %v", strconv.Itoa(teacher)+redirect)
-		r.URL.RawQuery = qv.Encode()
-		logger.Debugf("RawQuery: %v", r.URL.RawQuery)
+		// Get provider name.
+		provider := GetProviderName(r.URL.Path, 2)
```
meling, line 170: Can you please add a test for the `GetProviderName()` function with all the possible `URL.Path`s that it might receive.

I think it should be refactored...

I think it should return an error instead of an empty string. The error could then be passed on to `authError`. 

Note that there are two uses of the function with different `index` input, which is weird to me. What type of `URL.Path` input are expected here? And why do we need to pass the `index` input anyway?

web/auth/auth.go
```diff
@@ -177,145 +144,148 @@ func sessionData(session *sessions.Session) string {
...
+		if provider == "" {
+			authenticationError(logger, w, "incorrect request URL")
+			return
+		}
+		providerConfig, ok := authConfig.providers[provider]
```
meling, line 175: We should not be accessing a map inside a struct type. We should instead add a method on the `AuthConfig` struct:
```go
func (ac AuthConfig) GetConfig(provider string) (*oauth2.Config, error) { }
```

web/auth/auth.go
```diff
@@ -177,145 +144,148 @@ func sessionData(session *sessions.Session) string {
...
+		if !ok {
+			authenticationError(logger, w, "provider is not enabled")
+			return
+		}
 
-		url, err := gothic.GetAuthURL(w, r)
-		if err != nil {
-			logger.Error(err.Error())
-			return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+		// Check teacher suffix, update scopes if teacher.
```
meling, line 181: Similar for this code; seems reasonable that you could make things be a method:
```go
func (ac *AuthConfig) setScopes(urlPath string) {}
```
(but maybe not necessary to fix this if it will be deleted later anyway.

web/auth/auth.go
```diff
@@ -177,145 +144,148 @@ func sessionData(session *sessions.Session) string {
...
-		qv := r.URL.Query()
-		logger.Debugf("qv: %v", qv)
-		redirect := extractRedirectURL(r, Redirect)
-		logger.Debugf("redirect: %v", redirect)
-		// TODO: Add a random string to protect against CSRF.
-		qv.Set(State, strconv.Itoa(teacher)+redirect)
-		logger.Debugf("State: %v", strconv.Itoa(teacher)+redirect)
-		r.URL.RawQuery = qv.Encode()
-		logger.Debugf("RawQuery: %v", r.URL.RawQuery)
+		// Get provider name.
```
meling, line 169: It is not useful to add comments repeating the name of the function below it...

web/auth/auth.go
```diff
@@ -177,145 +144,148 @@ func sessionData(session *sessions.Session) string {
...
-// OAuth2Login tries to authenticate against an oauth2 provider.
-func OAuth2Login(logger *zap.SugaredLogger, db database.Database) echo.HandlerFunc {
-	return func(c echo.Context) error {
-		w := c.Response()
-		r := c.Request()
+// OAuth2Login redirects user to the provider's sign in page or, if user is already signed in with provider,
+// authenticates the user in the background.
+func OAuth2Login(logger *zap.SugaredLogger, authConfig *AuthConfig, secret string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		fmt.Println("AUTH2LOGIN started with: ", r.Method) // tmp
```
meling, line 151:
```go
logger.Debug()
```
Is a good option here, even if only for temporary logging.

web/auth/auth.go
```diff
@@ -177,145 +144,148 @@ func sessionData(session *sessions.Session) string {
...
+		if r.Method != "GET" {
+			authenticationError(logger, w, fmt.Sprintf("illegal request method: %s", r.Method))
+			return
+		}
 
-		provider, err := gothic.GetProviderName(r)
-		if err != nil {
-			logger.Error("failed to get gothic provider", zap.Error(err))
-			return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+		// Get provider.
```
meling, line 206: Remove comment.

web/auth/auth.go
```diff
@@ -177,145 +144,148 @@ func sessionData(session *sessions.Session) string {
...
+		provider := GetProviderName(r.URL.Path, 3)
```
meling, line 207: Why 3 here, when 2 was used in `OAuthLogin()`

web/auth/auth.go
meling: On second thought, why not hide this (provider lookup from `URL.Path`) in a method on `AuthConfig` that takes the `r` as input (similar to gothic).

web/auth/auth.go
meling: What's problem with?:
```go
provider, err := gothic.GetProviderName(r)
```

web/auth/auth.go
```diff
@@ -177,145 +144,148 @@ func sessionData(session *sessions.Session) string {
...
+		redirectURL := providerConfig.AuthCodeURL(secret)
+		logger.Debugf("redirecting to AuthURL: %v", redirectURL)
+		http.Redirect(w, r, redirectURL, http.StatusTemporaryRedirect)
 	}
 }
 
 // OAuth2Callback handles the callback from an oauth2 provider.
-func OAuth2Callback(logger *zap.SugaredLogger, db database.Database, scms *Scms) echo.HandlerFunc {
-	return func(c echo.Context) error {
+func OAuth2Callback(logger *zap.SugaredLogger, db database.Database, authConfig *AuthConfig, scms *Scms, secret string) http.HandlerFunc {
```
meling, line 197: This func is 169 lines long. There are many opportunities to extract code snippets into functions or methods. In particular, one or more methods could be added to the `AuthConfig` type.

web/auth/auth.go
```diff
@@ -177,145 +144,148 @@ func sessionData(session *sessions.Session) string {
...
+			authenticationError(logger, w, "got empty code on callback")
+			return
+		}
+
+		authToken, err := providerConfig.Exchange(context.Background(), code)
 		if err != nil {
-			logger.Error(err.Error())
-			return err
+			authenticationError(logger, w, fmt.Sprintf("failed to exchange access token: %s", err))
+			return
```
meling, line 241: Code from lines 207-241 could be made into a single method on `AuthConfig`, obviating the need to expose the `config.OAuth` (`providerConfig`).

PS: The method should return an `error` and not use the `authenticationError` helper function.

web/auth/auth.go
```diff
@@ -177,145 +144,148 @@ func sessionData(session *sessions.Session) string {
...
+		}
+
+		// Exchange code for access token.
+		code := r.FormValue("code")
+		if code == "" {
+			authenticationError(logger, w, "got empty code on callback")
+			return
+		}
+
+		authToken, err := providerConfig.Exchange(context.Background(), code)
```
meling, line 238: Should the `context` be passed in from the top-level??

web/auth/auth.go
meling: The proposed method should return a string-based `accessToken` instead of `oauth2.Token` since we don't use it.

web/auth/auth.go
```diff
@@ -177,145 +144,148 @@ func sessionData(session *sessions.Session) string {
...
-				logger.Debugf("no session cookie found in w: %v", w)
-			}
-			logger.Debugf("Redirecting: %s", redirect)
-			return c.Redirect(http.StatusFound, redirect)
+		// Complete user authentication.
+		externalUser := &externalUser{}
+		if err := json.NewDecoder(bytes.NewReader(responseBody)).Decode(&externalUser); err != nil {
+			authenticationError(logger, w, fmt.Sprintf("failed to decode user information: %v", err))
+			return
 		}
```
meling, line 274: Code from lines 244-274 could be made into a function `fetchExternalUser() (*externalUser, error)` or `fetchRemoteID()` if you include lines 287. But then you maybe don't want to log the JSON output... (since I wouldn't suggest to pass in the logger to this method...)

web/auth/auth.go
```diff
@@ -336,41 +306,49 @@ func OAuth2Callback(logger *zap.SugaredLogger, db database.Database, scms *Scms)
...
 
 		// in case this is a new user we need a user object with full information,
 		// otherwise frontend will get user object where only name, email and url are set.
 		user, err = db.GetUserByRemoteIdentity(remote)
 		if err != nil {
-			logger.Error(err.Error())
-			return err
+			authenticationError(logger, w, fmt.Sprintf("failed to fetch user %s for remote identity: %s", externalUser.Login, err))
+			return
 		}
```
meling, line 329: Lines 289-329 could be a method: `fetchUser(remote) (*qf.User, error)`

web/auth/auth_config.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/github"
+)
+
+type AuthConfig struct {
```
meling, line 11: Does this need to be exported? Suggest it be renamed to just `Config` to avoid stutter: `auth.AuthConfig` vs `auth.Config`.

web/auth/auth_config.go
```diff
@@ -0,0 +1,37 @@
...
+	providers map[string]*oauth2.Config
+}
+
+// AuthConfig creates a new OAuth 2.0 config for every enabled OAuth 2.0 provider.
+// Currently only works for GitHub.
+func NewAuthConfig(baseURL string) (*AuthConfig, error) {
+	providers := make(map[string]*oauth2.Config)
+
+	// Enable GitHub.
+	clientID := os.Getenv("GITHUB_KEY")
```
meling, line 21: Not sure this is the place to read the env vars. I think it is best to read the env in main.go and pass in these things.

web/auth/util.go
```diff
@@ -1,16 +1,37 @@
...
 // GetEventsURL returns the event URL for a given base URL and a provider.
 func GetEventsURL(baseURL, provider string) string {
-	return GetProviderURL(baseURL, "hook", provider, "events")
+	return GetProviderURL(baseURL, "hook", provider)
 }
 
 // GetProviderURL returns a URL endpoint given a base URL and a provider.
-func GetProviderURL(baseURL, route, provider, endpoint string) string {
-	return "https://" + baseURL + "/" + route + "/" + provider + "/" + endpoint
+func GetProviderURL(baseURL, route, provider string) string {
```
meling, line 18: This method should be unexported; it is only used by the two methods above. Moreover, please add a TODO here and remind us to replace this with the new [`URL.JoinPath()`](https://pkg.go.dev/net/url@master#URL.JoinPath) coming in Go 1.19.

web/auth/auth_config.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/github"
+)
+
+type AuthConfig struct {
+	providers map[string]*oauth2.Config
+}
+
+// AuthConfig creates a new OAuth 2.0 config for every enabled OAuth 2.0 provider.
```
meling, line 15: NewAuthConfig

web/auth/auth_config.go
```diff
@@ -0,0 +1,37 @@
...
+// Currently only works for GitHub.
```
meling, line 16: Should be generalized

web/auth/auth_config.go
```diff
@@ -0,0 +1,37 @@
...
+		ClientSecret: clientSecret,
+		Endpoint:     github.Endpoint,
+		RedirectURL:  GetCallbackURL(baseURL, "github"),
+		Scopes:       []string{"repo:invite"},
+	}
+	providers["github"] = githubConfig
+	return &AuthConfig{
+		providers: providers,
+	}, nil
+}
```
meling, line 37: Referring to #663 and new issue #694.

Suggest to remove the map here and simply add a method here like this:
```go
func NewGitHubConfig(baseURL, clientID, clientSecret string) *oauth2.Config {
	return &oauth2.Config{
		ClientID:     clientID,
		ClientSecret: clientSecret,
		Endpoint:     github.Endpoint,
		RedirectURL:  GetCallbackURL(baseURL, "github"),
		Scopes:       []string{"repo:invite"},
	}
}
```
The above func could be called from `main.go` and the `Config` object can be passed in to the `OAuth2` methods and used without having to examine the `r.URL.Path` to get the provider name and looking up in the map.

web/auth/auth_config.go
meling: See [comment below](https://github.com/quickfeed/quickfeed/pull/692/files#r921802570). I don't think we need this anymore since we don't want to support multiple providers with one quickfeed server.

web/auth/auth_config.go
meling: Maybe this new method should be placed in a new file `github.go` and we can add a corresponding `gitlab.go` version later.

web/auth/util.go
```diff
@@ -1,16 +1,37 @@
...
+	return "https://" + baseURL + "/" + route + "/" + provider // + "/" + endpoint
+}
+
+func GetProviderName(url string, index int) string {
```
meling, line 22: Remove; since no longer needed.

web/webserver.go
```diff
@@ -1,182 +1,74 @@
...
-		return
-	}
-	l.Fatal("failed to start server", zap.Error(srvErr))
-	// TODO(meling) pretty sure the following lines are unreachable; check
-
-	quit := make(chan os.Signal, 1)
-	signal.Notify(quit, os.Interrupt)
-	<-quit
+	// Register hooks.
+	ghHook := hooks.NewGitHubWebHook(s.logger, s.db, s.runner, s.bh.Secret)
```
meling, line 70: Here we use `s.bh.Secret` which is read from `WEBHOOK_SECRET` env var. However, on Line 64 `callbackSecret` is generated. Should these be the same? Recall also #480.

We use this struct below; perhaps it would make sense to add the `oauth2.Config` to it... and use that to pass as input to the `OAuth2` funcs. That said, if we do that, then is should be given a better name.

```go
// BaseHookOptions store the application's callback url and secret
type BaseHookOptions struct {
	BaseURL string
	Secret  string
	Config *oauth2.Config
}
```

web/auth/auth_test.go
```diff
@@ -34,343 +28,316 @@ func init() {
 	})
 }
 
-func TestOAuth2Logout(t *testing.T) {
```
meling, line 37: This is security critical logic and we must have proper tests for this.

web/auth/auth.go
meling: See https://github.com/quickfeed/quickfeed/pull/692/files#r921802570

meling (changes requested):
Missing important tests... Another suggestion to help reviewers from getting bogged down having to inspect legacy code; make the new code be in a new file (named differently from existing files) and then if necessary we can rename later... The old files should be deleted. For example, `auth/auth.go` (old version) should be deleted and the new version could be named `auth/authv2.go` or something like that... Then once the PR has been merged we can create a rename PR to name files the way we want it.

go.mod
0xf8f8ff: Yup, tests need a big rewrite. Will tidy the module files after that.

web/webserver.go
0xf8f8ff: The difference here is that `callbackSecret` is used in the user authentication (in `OAuth2Login` we pass the secret to GitHub, and in `OAuth2Callback` we check the secret in the response from GitHub to make sure the response has not been forged. So it is better to randomly generate `callbackSecret` instead of reading it from env for security reasons.

Webhook secret, on the other hand, is constant and set when the webhook is created (currently the same secret for each new course). This means the `s.bh.secret` will never change, so it makes sense to read it from the environment.

web/auth/auth.go
0xf8f8ff: Endpoint for `OAuth2Login` is `/auth/<provider>`. Endpoint for `OAuth2Callback` is `/auth/callback/<provider>`.

All `gothic` interactions rely heavily on custom provider data saved in the user session. Should I try to fine-tune sessions here if I am planning on replacing them with JWTs in the following PR?

web/auth/auth.go
0xf8f8ff: I wonder if we'd rather have different endpoints for different providers and refactor the common methods? I feel that reading anything from URLs is just wrong... 

This way we could have a constant `/auth/github` endpoint just for GitHub interactions, and add more if needed, like `/auth/gitlab` and so on.

main.go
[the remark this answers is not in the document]
0xf8f8ff: Should I also move `GITHUB_KEY` and `GITHUB_SECRET` env vars there?

main.go
0xf8f8ff: Right, but the rest of the `log.Fatalf()` messages are not capitalized, this is why I changed this one. Should I fix all the occurrences in the `main.go`?

web/auth/auth.go
0xf8f8ff: Everything session-related will be removed in PR that replaces sessions with JWT. Then the teacher-suffix part will disappear with the introduction of a GitHub App.

web/auth/auth.go
0xf8f8ff: Yes, we won't need to adjust scopes with a GitHub App.

web/auth/auth.go
0xf8f8ff: Should we change `/auth/callback/<provider>` to `/callback/provider`?

main.go
```diff
@@ -98,27 +102,50 @@ func main() {
...
+		MuxServer: grpcWebServer,
+	}
+
+	// Register HTTP endpoints and webhooks
+	router := qfService.RegisterRouter(authConfig, scms, multiplexer, *public)
+
 	// Create a HTTP server for prometheus.
 	httpServer := &http.Server{
 		Handler: promhttp.HandlerFor(reg, promhttp.HandlerOpts{}),
 		Addr:    fmt.Sprintf("0.0.0.0:%d", 9097),
```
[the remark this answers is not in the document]
0xf8f8ff: Hmm, it looks to be a dedicated server for Prometheus, is it really a good idea to mix it with the authentication logic? Changed address to 127.0.0.1.

web/auth/auth.go
0xf8f8ff: Yes, it is better to take the context from the request. Fixed.

main.go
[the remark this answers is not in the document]
meling: I've merged #695 now. These should be added to `internal/env` as you suggest. The functions should be named `ClientKey()` and `ClientSecret()`. Note that these are required vars, so should not provide defaults, and return an error if not provided.

For the env var names perhaps they should be prefixed with `QUICKFEED_` as the others.

The question is whether or not they should be:
1. `QUICKFEED_CLIENT_KEY`
2. `QUICKFEED_GITHUB_CLIENT_KEY`
3. `QUICKFEED_GITLAB_CLIENT_KEY`

Let's use option 1 for now.

main.go
meling: Yes.

main.go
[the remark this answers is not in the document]
meling: Should be fine to keep this separate.

web/auth/auth.go
meling: I agree that examining the URLs is the wrong approach. I believe [my comment](https://github.com/quickfeed/quickfeed/pull/692/files#r921802570) mentioned above, resolves this problem, obviating the need for the `GetProviderName()` method altogether.

web/auth/auth.go
meling: I think we don't need the `GetProviderName()` approach at all, if we just use the approach outlined in [my comment](https://github.com/quickfeed/quickfeed/pull/692/files#r921802570).

meling: The point is that we don't want to support more than one provider per running instance of quickfeed. If you want to run with gitlab, then you need to deploy another instance of quickfeed (with a separate database etc.)

The old approach of supporting multiple providers was a design decision that we should have dropped long ago.

web/auth/auth_config.go
meling: I have decided that we no longer want to support multiple providers per running quickfeed instance. Hence, we should only need a single auth endpoint. That said, if we need to have a `auth.Config` type, that's fine. But with the current code, I don't see the need. We can just return an `oauth2.Config` object.

Unless you expect a future need in the context of JWTs etc that would require this `auth.Config` type, then I propose to leave it out.

web/auth/auth.go
```diff
@@ -177,200 +143,91 @@ func sessionData(session *sessions.Session) string {
...
+		// Start or refresh the session.
+		// Issue: server generates a new random key to encode sessions on each restart.
+		// A session from before the restart will be detected, but cannot be decoded with
+		// the new key, resulting in an error. Instead of returning, we attempt to refresh the session first.
+		s, err := sessionStore.New(r, SessionKey)
 		if err != nil {
-			logger.Error(err.Error())
-			return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+			if err := sessionStore.Save(r, w, s); err != nil {
+				authenticationError(logger, w, fmt.Errorf("failed to create new session (%s): %w", SessionKey, err))
```
meling, line 162: Using `SessionKey` here is not useful since it will just print `session`.

web/auth/auth.go
```diff
@@ -177,200 +143,91 @@ func sessionData(session *sessions.Session) string {
...
+			authenticationError(logger, w, fmt.Errorf("failed to fetch user %s for remote identity: %w", externalUser.Login, err))
+			return
 		}
 		logger.Debugf("Fetching full user info for %v, user: %v", remote, user)
 
 		// Register user session.
+		// Temporary. Will be removed when sessions are replaced with JWT.
+		s, err := sessionStore.Get(r, SessionKey)
+		if err != nil {
+			authenticationError(logger, w, fmt.Errorf("failed to get user session (%s): %w", SessionKey, err))
```
meling, line 217: Using `SessionKey` here is not useful since it will just print `session`.

web/auth/auth_test.go
```diff
@@ -2,425 +2,289 @@ package auth_test
...
-
-	fakeSessionKey  = "fake"
-	fakeSessionName = fakeSessionKey + gothic.SessionName
+	testSecret = "top-secret"
 )
 
 func init() {
 	goth.UseProviders(&auth.FakeProvider{
-		Callback: "/auth/fake/callback",
+		Callback: "/auth/callback/fake",
```
meling, line 22: Would it make sense to add something like this to the `web` package, e.g., in `webserver.go`:
```go
const (
	Auth     = "/auth/"
	Teacher  = "/auth/teacher/"
	Callback = "/auth/callback/"
	Logout   = "/logout"
	Hook     = "/hook/"
)
```
And use them here, e.g.:
```go
Callback: web.Callback + "fake",
```
And similar for others below??

(Just in case we want to change the endpoints later...)

web/auth/auth_test.go
```diff
@@ -2,425 +2,289 @@ package auth_test
...
-	ns = len(store.store[r].Values)
-	// Sessions should be cleared.
-	if ns != 0 {
-		t.Errorf("have %d sessions want %d", ns, 0)
-	}
+func TestOAuth2Login(t *testing.T) {
+	logger := qlog.Logger(t)
+	authConfig := auth.NewGitHubConfig("", "TEST", "TEST")
+	// Incorrect request method.
+	apitest.New().Debug().
```
meling, line 30: I like this new testing framework. :-) 

However, is there a way to turn off debug mode? E.g., so that we can run `go test -v` to enable debug mode (or more verbose output)...?

meling (changes requested):
Looks good now. Only a few minor things that needs your input.

PR #699, opened 2022-07-23
Moved self-signed certification generation to internal/cert

meling:
This moves the generation of self-signed certs to the internal
package cert to allow it to be used from other main.go or other
packages to generate self-signed certs on-demand.


0xf8f8ff (approved):

.env-template
```diff
@@ -1,12 +1,15 @@
...
 # WARNING: Do not push your file with secret values to GitHub or elsewhere.
 
-# GitHub OAUTH App keys
-GITHUB_KEY="KEY"
-GITHUB_SECRET="SECRET"
+# GitHub OAuth Application keys for localhost
+QUICKFEED_CLIENT_ID=""
+QUICKFEED_CLIENT_SECRET=""
+# Certificate chain and private key file
+QUICKFEED_KEY_FILE =$QUICKFEED/internal/config/certs/privkey.pem
```
JosteinLindhom, line 8: Remove whitespace in `FILE =$QUICKFEED`.

internal/env/load.go
```diff
@@ -2,17 +2,37 @@ package env
...
 
-var dotEnvPath = ".env"
+const dotEnvPath = ".env"
+
+var quickfeedRoot string
+
+func init() {
+	quickfeedRoot = os.Getenv("QUICKFEED")
+	if quickfeedRoot == "" {
+		out, err := sh.Output("go list -m -f {{.Dir}}")
```
meling, line 20: @r0qs Pointed out that this may not work if compiled into a binary, unless the go command is actually installed. This should not be a problem for now, as long as we run it from a system with go installed. However, we might consider in the future to replace `go list` with a library function to scan for the `go.mod` file. Not sure if there is a library func for this in e.g., `x/tools` or some other place.

PR #700, opened 2022-07-25
Replacing session-based authentication with JWTs.

0xf8f8ff:
This PR replaces HTTP sessions with JSON Web Tokens for user authentication.

Note: this PR implements some basic functionality for refreshing JWTs, but the main feature will be added in the following PR dedicated to the new (interceptor-based) access control process.


web/auth/interceptors/request_validator.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// RequestValidator returns a new unary server interceptor that validates requests
+// that implements the validator interface.
+// Invalid requests are rejected without logging and before it reaches any
+// user-level code and returns an illegal argument to the client.
+// In addition, the interceptor also implements a cancel mechanism.
+func RequestValidator(logger *zap.Logger) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		methodName := info.FullMethod[strings.LastIndex(info.FullMethod, "/")+1:]
```
meling, line 31: Not sure we discussed it before, but the prometheus metrics stuff seems like a good target for a separate interceptor in `metrics.go`. Also, the `qf.Ag*` types should probably be moved to the interceptor package instead of the `qf` package? 

Similarly, I wonder if the following:
```go
	reg.MustRegister(
		promgrpc.NewServerMetrics(),
		qf.AgFailedMethodsMetric,
		qf.AgMethodSuccessRateMetric,
		qf.AgResponseTimeByMethodsMetric,
	)
```
could be move to the interceptor package's `init()` func. You would need to add a func like this:
```go
var reg = prometheus.NewRegistry()

func NewMetricsRegistry() *prometheus.Registry {
  return reg
}
```
So that you can get the `reg` to be passed to the http server. Or you could move this too:
```go
// Create an HTTP server for prometheus.
	httpServer := &http.Server{
		Handler: promhttp.HandlerFor(reg, promhttp.HandlerOpts{}),
		Addr:    fmt.Sprintf("127.0.0.1:%d", 9097),
	}
	go func() {
		if err := httpServer.ListenAndServe(); err != nil {
			log.Fatalf("Failed to start a http server: %v", err)
		}
	}()
```
And provide some other way to start it from main... but be aware of the `log.Fatalf` not being in `main.go` is an issue...

JosteinLindhom: I split the interceptors in this branch [grpc-interceptors](https://github.com/quickfeed/quickfeed/tree/grpc-interceptors).

Metrics: [metrics.go](https://github.com/quickfeed/quickfeed/blob/grpc-interceptors/qf/metrics.go)
Validation: [validation.go](https://github.com/quickfeed/quickfeed/blob/grpc-interceptors/qf/validation.go)
User Verifier: [auth.go](https://github.com/quickfeed/quickfeed/blob/83054cfbacd03b341db6abc85098f9703db507b4/web/auth/auth.go#L510)

Usage in `main.go`: [link](https://github.com/quickfeed/quickfeed/blob/83054cfbacd03b341db6abc85098f9703db507b4/main.go#L107)

`auth.go` contains both unary and streaming variants of the interceptor.

0xf8f8ff: Ok, should I then wait until it is ready and merged into master (and work on something else meanwhile)?

web/auth/auth_test.go
```diff
@@ -31,18 +33,18 @@ func TestOAuth2Login(t *testing.T) {
...
 		Expect(t).
 		Status(http.StatusTemporaryRedirect).
 		End()
-	// Outdated session with expected name should not break API.
+	// Outdated auth cookie with expected name should not break API.
 	apitest.New().Debug().
 		HandlerFunc(auth.OAuth2Login(logger, authConfig, "")).
 		Get(web.Auth).
-		Cookie("session", "empty").
+		Cookie("auth", "empty").
```
meling, line 47: We should ideally use a global/exported `const` instead of `"auth"`. There are two:
- `auth.CookieName`
- `tokens.authCookieName`

Suggest to rename the latter and use it as `tokens.CookieName` or `tokens.AuthCookieName`. This is because `auth` cannot be imported into `tokens` package.

web/auth/tokens/jwt.go
```diff
@@ -0,0 +1,156 @@
+package tokens
```
meling, line 1: We should avoid packages with plural names. So maybe just `token` works? Or maybe it could be `jwt` (I prefer `token`). Another options is to move the content of the `tokens` package into the `auth` package.

web/auth/tokens/jwt_test.go
```diff
@@ -0,0 +1,174 @@
...
+	"testing"
+	"time"
+
+	"github.com/golang-jwt/jwt"
+	"github.com/quickfeed/quickfeed/internal/qtest"
+	"github.com/quickfeed/quickfeed/qf"
+	"github.com/quickfeed/quickfeed/web/auth/tokens"
+)
+
+func TestNewManager(t *testing.T) {
```
meling, line 13: General comment about tests. I didn't look at these tests carefully in terms of the following. But I think it is possible that there are some benefits to be had from rephrasing the tests as table-driven tests. Things like:

```go
tests := []struct{
  inputA string
  inputB string
  want string
}{ ...
```

web/auth/tokens/update.go
```diff
@@ -0,0 +1,83 @@
...
+func (tm *TokenManager) UpdateRequired(claims *Claims) bool {
+	for _, token := range tm.updateTokens {
+		if claims.UserID == token {
+			return true
+		}
+	}
+	return claims.ExpiresAt <= time.Now().Unix()
+}
+
+// Update removes user ID from the manager and updates user record in the database
```
meling, line 19: Remember to add `.` at the end of doc comments.

web/interceptor/helpers.go
```diff
@@ -0,0 +1,58 @@
...
+		logger.Errorf("Failed to extract claims from JWT: %v", err)
+		return nil, ErrInvalidAuthCookie
+	}
+	if tm.UpdateRequired(claims) {
+		logger.Debug("Updating token for user ", claims.UserID)
+		updatedToken, err := tm.NewAuthCookie(claims.UserID)
+		if err != nil {
+			logger.Errorf("Failed to update cookie: %v", err)
+		}
+		if err := grpc.SendHeader(ctx, metadata.Pairs("Set-Cookie", updatedToken.String())); err != nil {
```
meling, line 39: Should the `Set-Cookie` use a const?

web/interceptor/helpers.go
```diff
@@ -0,0 +1,58 @@
...
+func getAuthenticatedContext(ctx context.Context, logger *zap.SugaredLogger, tm *tokens.TokenManager) (context.Context, error) {
+	meta, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		logger.Error("Failed to extract metadata")
+		return nil, ErrContextMetadata
+	}
+	token, err := extractToken(meta)
+	if err != nil {
+		logger.Errorf("Failed to extract token: %v", err)
+		return nil, ErrInvalidAuthCookie
```
meling, line 26: The returned error from `extractToken` is already `ErrInvalidAuthCookie` and can be returned directly.

web/interceptor/helpers.go
```diff
@@ -0,0 +1,58 @@
...
+	claims, err := tm.GetClaims(token)
+	if err != nil {
+		logger.Errorf("Failed to extract claims from JWT: %v", err)
+		return nil, ErrInvalidAuthCookie
+	}
+	if tm.UpdateRequired(claims) {
+		logger.Debug("Updating token for user ", claims.UserID)
+		updatedToken, err := tm.NewAuthCookie(claims.UserID)
+		if err != nil {
+			logger.Errorf("Failed to update cookie: %v", err)
```
meling, line 37: Should we continue after this fail? or should the error be returned?

web/interceptor/auth_test.go
```diff
@@ -0,0 +1,114 @@
...
+				t.Errorf("GetUser().Code(): %v, want: %v", s.Code(), user.code)
+			}
+		}
+		if user.wantUser != nil {
+			// ignore comparing remote identity
+			user.wantUser.RemoteIdentities = nil
+		}
+		wantUser := user.wantUser
+		if diff := cmp.Diff(wantUser, gotUser, protocmp.Transform()); diff != "" {
+			t.Errorf("GetSelf() mismatch (-wantUser +gotUser):\n%s", diff)
```
meling, line 111: `s/GetSelf/GetUser/`

meling (changes requested):
Mostly minor... Main issue is package rename or reorganizing.

web/auth/tokens/jwt.go
0xf8f8ff: Moved into the `auth` package.

web/auth/tokens/jwt_test.go
0xf8f8ff: Does it really apply to the tests in this file? I agree in general but can't see how it could be beneficial in this case.

web/interceptor/helpers.go
0xf8f8ff: It absolutely should. Fixed.

web/interceptor/helpers.go
0xf8f8ff: I think it is better to return an error. Fixed.

PR #706, opened 2022-07-29
Replacing OAuth App with GitHub App

0xf8f8ff:


internal/env/scm.go
```diff
@@ -46,6 +50,24 @@ func ClientSecret() (string, error) {
 	return clientSecret, nil
 }
 
+// AppID returns the appliction ID for the current SCM provider.
```
meling, line 53: Does it make sense to provide a default path and AppID? Similar to `CertFile` and `KeyFile`.

Makefile
```diff
@@ -155,5 +155,8 @@ run:
 runlocal:
```
meling, line 155: Do we need both `runlocal` and `rundev` targets?

scm/scm_manager.go
```diff
@@ -0,0 +1,115 @@
...
+type SCMManager struct {
+	Scms         *Scms
+	githubConfig *GithubConfig
+}
+
+// newGitHubConfig creates a new configuration for GitHub app
+func newGitHubConfig(appID, appKey, appSecret, appKeyFile string) *GithubConfig {
+	return &GithubConfig{
+		appID:    appID,
+		clientID: appKey,
```
meling, line 39: Confusing that `appKey` is assigned to the `clientID` field. Mistake? If not, suggest to rename one of them.

scm/scm_manager.go
```diff
@@ -0,0 +1,115 @@
...
+	"go.uber.org/zap"
+)
+
+const installationAPI = "https://api.github.com/app/installations"
+
+// GithubConfig keeps parameters of the GitHub app.
+type GithubConfig struct {
+	appID     string
+	clientID  string
+	secret    string
```
meling, line 23: Is this `clientSecret` or `appSecret`? Should use full name to avoid confusion.

scm/scm_manager.go
```diff
@@ -0,0 +1,115 @@
...
+	return conf.appID != "" && conf.keyPath != "" &&
+		conf.clientID != "" && conf.secret != ""
+}
+
+// NewSCMManager creates base client for the QuickFeed GitHub Application. This client
+// cannto access organization-related GitHub API and is used to derive installation
+// API clients for each course organization.
+func NewSCMManager(appID, appKey, appSecret, appKeyFile string) (*SCMManager, error) {
+	config := newGitHubConfig(appID, appKey, appSecret, appKeyFile)
+	if !config.valid() {
```
meling, line 56: I prefer to have the validity check in the `newGitHubConfig()` function, and return an error from it that can simply be forwarded from here like this:
```go
config, err := newGitHubConfig(...)
if err != nil {
  return err
}
```
That is, I don't see the need for the `valid()` method, unless it will be call by others.

scm/scm_manager.go
```diff
@@ -0,0 +1,115 @@
...
+		return nil, fmt.Errorf("error configuring GitHub App: %+v", config)
+	}
+	createAppKey, err := key.FromFile(appKeyFile)
+	if err != nil {
+		wd, _ := os.Getwd()
```
meling, line 61: Do you need this? I mean, I would expect the `err` to include the full path to the `appKeyFile` that it isn't able to read.

scm/scm_manager.go
```diff
@@ -0,0 +1,115 @@
...
+		return nil, fmt.Errorf("wd %s, error reading key from file: %s", wd, err)
```
meling, line 62: Please use the `%w` verb (not `%s`) when wrapping `err`.

scm/scm_manager.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	appClientConfig, err := app.NewConfig(config.appID, createAppKey)
+	if err != nil {
+		return nil, fmt.Errorf("error creating GitHub application client: %s", err)
```
meling, line 66: Please use the `%w` verb (not `%s`) when wrapping `err`.

scm/scm_manager.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/beatlabs/github-auth/app"
+	"github.com/beatlabs/github-auth/key"
+	"github.com/google/go-github/v45/github"
+	"go.uber.org/zap"
+)
+
+const installationAPI = "https://api.github.com/app/installations"
+
+// GithubConfig keeps parameters of the GitHub app.
+type GithubConfig struct {
```
meling, line 20: After more careful checking, it seems this struct only needs `appID` and `appConfig`; all the other fields are not used.

go.mod
```diff
@@ -34,6 +34,7 @@ require (
 	github.com/Microsoft/go-winio v0.5.2 // indirect
 	github.com/ProtonMail/go-crypto v0.0.0-20210428141323-04723f9f07d7 // indirect
 	github.com/acomagu/bufpipe v1.0.3 // indirect
+	github.com/beatlabs/github-auth v0.0.0-20220721134423-2b8d98e205d1 // indirect
```
meling, line 37: Should run `go mod tidy` since this is a `direct` dependency.

scm/scm_manager.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// valid ensures that all configuration fields are not empty
+func (conf *GithubConfig) valid() bool {
+	return conf.appID != "" && conf.keyPath != "" &&
+		conf.clientID != "" && conf.secret != ""
+}
+
+// NewSCMManager creates base client for the QuickFeed GitHub Application. This client
+// cannto access organization-related GitHub API and is used to derive installation
```
meling, line 52: Typo: cannot. But suggest to rephrase:
`This client can be used to install API clients for each course organization.`

scm/scm_manager.go
```diff
@@ -0,0 +1,115 @@
...
+		wd, _ := os.Getwd()
+		return nil, fmt.Errorf("wd %s, error reading key from file: %s", wd, err)
+	}
+	appClientConfig, err := app.NewConfig(config.appID, createAppKey)
+	if err != nil {
+		return nil, fmt.Errorf("error creating GitHub application client: %s", err)
+	}
+	config.appConfig = appClientConfig
+	return &SCMManager{
+		githubConfig: config,
```
meling, line 70: Since we only need the `appConfig` and `appID` I suggest to remove the `GithubConfig` struct and instead add those two fields directly in `SCMManager` to avoid things like this: `sm.githubConfig.appConfig.`

meling (changes requested):
This looks very nice. I've requested a few changes mainly in `scm_manager.go`

Makefile
0xf8f8ff: I have no idea about `runlocal` and where and when it is used, so I didn't dare to remove it.

meling: Looks like you [added it 3 years ago](https://github.com/quickfeed/quickfeed/commit/588e9111f7f55a5216c4866cdaac9066c5d8f1be) 😜

internal/env/scm.go
0xf8f8ff: I can add a default file path, but then I must know what the default file name will be (the .pem file must be generated on the app's web page and then saved, it must be done manually). 

`AppID` is an app-specific variable just like `clientID` and `clientSecret` (OAuth apps had only two (ID, secret), GitHub apps have three (app ID, client ID, client secret). Providing a default `appID` would expose the ID of the GitHub App created for UIS.

Makefile
0xf8f8ff: The one I added was `@aguis -service.url 127.0.0.1 -http.addr :9091 -http.public ./public` and was used to run on the `<username>.itest.run` with port forwarding. 

The one I didn't dare to remove is `@quickfeed -service.url 127.0.0.1` and it couldn't be used without a registered hostname (which I never had). So I still have no idea about its purpose.

scm/scm_manager.go
0xf8f8ff: Agree, it could use some renaming.

scm/scm_manager.go
0xf8f8ff: Just debugging artifacts, will remove.

internal/env/scm.go
meling: Seems like providing a default file path makes sense. The user would be expected to place the `.pem` file at the default location, unless another path is provided via the env var.

Another thing I've been thinking about in relation to this. It seems reasonable that these clientID, clientSecret, appID and so on could be bundled into a struct that can be passed around instead of individual arguments...

0xf8f8ff: Where should the default location be then? In the project's root folder or some dedicated place?

Should I return the `githubConfig` struct that stored all these variables back then? Probably should rename it to something like `scmConfig` too.

meling: Yes. `scmConfig` is nice. And the folder could be `internal/config/github/quickfeed.pem` or something like that?

Makefile
```diff
@@ -157,7 +157,7 @@ run:
 	@quickfeed -service.url $(DOMAIN) -database.file ./tmp.db
 
 runlocal:
-	@quickfeed -service.url 127.0.0.1
+	@quickfeed -dev -service.url 127.0.0.1:8081
```
meling, line 160: Should we use the `-dev` only shorthand? Or do we want the `-service.url` also?

assignments/assignments.go
```diff
@@ -22,19 +22,16 @@ import (
 const MaxWait = 5 * time.Minute
 
 // UpdateFromTestsRepo updates the database record for the course assignments.
-func UpdateFromTestsRepo(logger *zap.SugaredLogger, db database.Database, course *qf.Course) {
+func UpdateFromTestsRepo(logger *zap.SugaredLogger, db database.Database, s *scm.SCMManager, course *qf.Course) {
```
meling, line 25: `s/SCMManager/Manager/` (Avoids stutter...)

scm/scm_manager.go
```diff
@@ -0,0 +1,173 @@
...
+
+// SCMClient gets an existing SCM client by organization name or creates a new client for course organization.
+func (s *SCMManager) GetOrCreateSCM(ctx context.Context, logger *zap.SugaredLogger, organization string) (SCM, error) {
+	client, ok := s.Scms.scms[organization]
+	if !ok {
+		cli, err := s.newInstallationClient(ctx, organization)
+		if err != nil {
+			return nil, err
+		}
+		client = &GithubSCM{
```
meling, line 93: Being part of the `Manager`, this should avoid directly constructing for GitHub only. We should use the `NewSCMClient()` method.

scm/scm_manager.go
```diff
@@ -0,0 +1,173 @@
...
+		Scms:      NewScms(),
+	}, nil
+}
+
+func (s *SCMManager) SCMWithToken(ctx context.Context, logger *zap.SugaredLogger, organization string) (SCM, error) {
+	scmClient, err := s.GetOrCreateSCM(ctx, logger, organization)
+	if err != nil {
+		return nil, err
+	}
+	token, err := s.getInstallationToken(ctx, organization)
```
meling, line 77: Does not depend on `scmClient`, so can be called first. Thus, the `token` can be passed to `internalGetOrCreateSCM`... (see below.)

Makefile
0xf8f8ff: We need it because the default `service.url` is empty, but it is used to construct the callback and webhook URLs so we need it to be correct. 

One option is to add the logic that will set `127.0.0.1` in combination with a correct port as `baseURL` on start in case `-dev` is used.

meling: The reason I mentioned this was this code:
```go
	if *dev {
		*baseURL = "127.0.0.1" + *httpAddr
	}
```
Which seems to make the `-service.url 127.0.0.1:8081` unnecessary when you use `-dev`

scm/scm_manager.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return nil, err
+	}
+	scmClient.SetToken(token)
+	return scmClient, err
+}
+
+// SCMClient gets an existing SCM client by organization name or creates a new client for course organization.
+func (s *SCMManager) GetOrCreateSCM(ctx context.Context, logger *zap.SugaredLogger, organization string) (SCM, error) {
```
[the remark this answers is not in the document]
meling: Okay. That makes sense. Perhaps we should name the function `RefreshToken` and add a doc comment to explain this.

Makefile
JosteinLindhom: Would it be a possibility to remove `-service.url` and use `*baseURL = "127.0.0.1" + *httpAddr` (or something like  `127.0.0.1 + env.Port()`) if `-dev`, and `env.Domain()` otherwise?

scm/scm_manager.go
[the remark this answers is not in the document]
meling: On the other hand, `SetToken` is only called from this location, so it doesn't need to be exported. Or do you foresee that it will be called from other places as well?

Makefile
0xf8f8ff: > Would it be a possibility to remove `-service.url` and use `*baseURL = "127.0.0.1" + *httpAddr` (or something like `127.0.0.1 + env.Port()`) if `-dev`, and `env.Domain()` otherwise?

I like this idea.

scm/scm_manager.go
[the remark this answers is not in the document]
0xf8f8ff: No, I think it is only useful for cloning.

meling (approved):

PR #708, opened 2022-08-09
Use autocert to manage certificates

JosteinLindhom:


.env-template
```diff
@@ -12,6 +12,11 @@ ENVOY_CONFIG=$QUICKFEED/internal/config/envoy/envoy-localhost.yaml
 
 # Quickfeed server domain or ip
 DOMAIN=localhost
+
+# Comma-separated list of domains to allow certificates for
```
meling, line 16:
```
# ... certificates for, e.g., "www.example.com,example.com".
# IP addresses and "localhost" are *not* valid.
# The whitelist ...
```

internal/env/cert.go
```diff
@@ -22,6 +25,25 @@ func Domain() string {
 	return domain
 }
 
+// WhiteList returns a list of domains that the server will create certificates for.
+func WhiteList() ([]string, error) {
```
meling, line 29: Should be just `Whitelist` (since its an English word)

internal/env/cert.go
```diff
@@ -22,6 +25,25 @@ func Domain() string {
...
+	domains := os.Getenv("QUICKFEED_WHITELIST")
+	if domains == "" {
+		return []string{}, errors.New("no domains in whitelist. Please set the WHITELIST environment variable")
```
meling, line 32: Maybe avoid mentioning the env variable to avoid forgetting to update both ;-) 

What about:
```
required whitelist is undefined.
```
Also, you should `return nil, errors...` instead of constructing an empty slice. Since if a user of `Whitelist()` forgets to check for `nil` error, the empty slice returned might cause a problem... (probably not, but better to be safe...)

internal/env/cert.go
```diff
@@ -22,6 +25,25 @@ func Domain() string {
...
+	}
+	// Check if domains contain any IP addresses or localhost.
```
meling, line 34: Comment seems unnecessary.

main.go
```diff
@@ -142,7 +144,24 @@ func main() {
...
 		}
 	}
-	if err := muxServer.ListenAndServe(); err != nil {
+	whitelist, err := env.WhiteList()
+	if err != nil {
+		log.Fatalf("Failed to get whitelist: %v", err)
+	}
+	certManager := autocert.Manager{
+		Prompt: autocert.AcceptTOS,
+		Cache:  autocert.DirCache("certs"),
```
meling, line 153: Here we should use the `env.CertPath()` method that gets the path from `QUICKFEED_CERT_PATH`.

However, `internal/env/certPath` is currently unexported. But we should exported it.

main.go
```diff
@@ -142,7 +144,24 @@ func main() {
...
+		HostPolicy: autocert.HostWhitelist(
+			whitelist...,
+		),
+	}
+	muxServer.TLSConfig = &tls.Config{
```
meling, line 158: Deepsource complains about two issues related to security here.

meling (changes requested):
Some issues to fix

main.go
JosteinLindhom: Agreed. I updated the default path to `/internal/config/certs` as the previous default, `/etc/letsencrypt/live` requires elevated privileges to write to.

meling (approved):
Looks good to me!

PR #711, opened 2022-08-10
Convert gRPC service to connect-go

JosteinLindhom:
- Experiment with buf generate
- Update server with connect-go interceptors and handlers
- Update QuickFeed service to use connect-go
- Update tests to be compatible with connect-go
- Add connect-go generated service


public/package.json
```diff
@@ -3,6 +3,8 @@
   "version": "0.2.0",
   "private": true,
   "dependencies": {
+    "@bufbuild/connect-web": "^0.1.0",
+    "@bufbuild/protobuf": "^0.0.10",
     "@types/react": "^16.14.2",
     "@types/react-dom": "^16.9.10",
     "google-protobuf": "^3.15.0-rc.1",
```
meling, line 10: Do we need `grpc-web` below? And do we need `google-protobuf`?

public/src/App.tsx
```diff
@@ -17,6 +18,8 @@ const App = (): JSX.Element => {
     useEffect(() => {
         async function setup() {
             await actions.fetchUserData()
+            // TODO: Remove this once finished testing
```
meling, line 21: Let's remove it.

meling (approved):

Issue #701, opened 2022-07-25
Separate frontend proto types from other types to reduce bundle size


meling:
Since the frontend does not need to access all message types, we should not generate JavaScript / TypeScript code for message types we don't need. Hence, we should split out remaining types into a separate file or reorganize the current files so as to avoid generating code that won't be used in the frontend. The key idea here is that only message types used by the RPC methods should be imported by the `quickfeed.proto` file, not both message-only imports:
```proto
import "qf/types.proto";
import "qf/requests.proto";
```
The `requests.proto` is obviously needed, but not all of `types.proto`, and so we should figure out which types are needed on the frontend, and import those from a separate `proto` file... maybe `exported_types.proto` or something like that?

Note that `requests.proto` imports `types.proto`. If this idea is to work, then `requests.proto` can only import `exported_types.proto`.

This should save load time for the frontend, since the bundle size should be more compact.

Clearly, if there is little to be gained from this, e.g., only a few message types is removed, then this might not be worth doing. @JosteinLindhom Could you check if this is worth doing?

We would need to update the `proto` target in the `Makefile` to generate Go code and JavaScript / TypeScript in separate passes, since now we invoke `protoc` with all plugins in one pass.

meling:
As the following shows, the bundle can be a bit big:

```log
assets by status 110 KiB [cached] 23 assets
assets by status 607 KiB [emitted]
  asset npm.google-protobuf.d82c031c96e6cb3ae777.bundle.js 232 KiB [emitted] [immutable] [minimized] (name: npm.google-protobuf) (id hint: vendor) 1 related asset
  asset proto.b17ab13761208046fc32.bundle.js 143 KiB [emitted] [immutable] [minimized] (name: proto) 1 related asset
  asset index.f632af8b1c89b17462eb.bundle.js 121 KiB [emitted] [immutable] [minimized] (name: index) 1 related asset
  asset overmind.89d1c93a5cf375418788.bundle.js 108 KiB [emitted] [immutable] [minimized] (name: overmind) 1 related asset
  asset ../assets/index.html 3.44 KiB [emitted] [compared for emit]
Entrypoint index 124 KiB (375 KiB) = npm.react-router-dom.0dcb98e90404f91d618a.bundle.js 3.46 KiB index.f632af8b1c89b17462eb.bundle.js 121 KiB 2 auxiliary assets
Entrypoint overmind 213 KiB (755 KiB) = 21 assets 21 auxiliary assets
[...]

PR #712, opened 2022-08-10
Interceptor-based access control

0xf8f8ff:


meling:
I've only briefly looked at the code. 

Could it be an idea to add a test matrix that tries to invoke (as a gRPC client) all the `QuickFeedService` methods with different user permissions, different enrollment in courses, etc and check that the:
- expected accesses are granted
- unexpected accesses are denied
- any thing else??

Another test to add should be:
- Check that all `QuickFeedService` interface methods is listed in the access control matrix and that none are forgotten and that they are all being tested.
- This can be done using reflection on the `QuickFeedService` interface, and check that each method's name is in the access control map.

0xf8f8ff:
> * Check that all `QuickFeedService` interface methods is listed in the access control matrix and that none are forgotten and that they are all being tested.
> * This can be done using reflection on the `QuickFeedService` interface, and check that each method's name is in the access control map.

There are a few methods that any user can call, they are not in the matrix. Should I add them with empty roles instead?

meling:
> > * Check that all `QuickFeedService` interface methods is listed in the access control matrix and that none are forgotten and that they are all being tested.
> > * This can be done using reflection on the `QuickFeedService` interface, and check that each method's name is in the access control map.
> 
> There are a few methods that any user can call, they are not in the matrix. Should I add them with empty roles instead?

Yes, because we should still test them to see that they aren't being restricted.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/quickfeed/quickfeed/web/auth"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+)
+
+type (
+	role      int
+	roles     []role
+	requestID interface {
+		FetchID(string) uint64
```
meling, line 17: Please refactor this as follows:
```go
	requestID interface {
		IDFor(role) uint64
	}
```
See below for additional suggestions.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+)
+
+const (
+	// user role implies that user attempts to access information about himself.
+	user role = iota
```
meling, line 23: I've read somewhere (can't remember where), that we should have a `none` value as the first entry for enum/iota types. This makes sense also because the database is not great when it comes to updating things to the zero value etc...

That is make sense here?

web/interceptor/access_control.go
```diff
@@ -0,0 +1,150 @@
...
+		if ok {
+			logger.Debug("Got roles: ", roles) // tmp
+			claims, err := tm.GetClaims(ctx)
+			if err != nil {
+				logger.Error("Access control: failed to get claims from request context: %v", err)
+				return handler(ctx, req)
+			}
+			logger.Debug("Got user claims: ", claims) // tmp
+			for _, role := range roles {
+				switch role {
```
meling, line 95: Here you can define:
```go
hasID := func(id uint64) bool {
	if m, ok := req.(reqID); ok {
		return m.IDFor(role) == id
	}
	return false
}
hasStatus := func(status qf.Enrollment_UserStatus) bool {
	if m, ok := req.(reqID); ok {
		claimStatus, found := claims.Courses[m.IDFor(role)]
		return found && claimStatus == status
	}
	return false
}
```

web/interceptor/access_control.go
```diff
@@ -0,0 +1,150 @@
...
+				case user:
+					if m, ok := req.(requestID); ok {
```
meling, line 97: Here you can then do:
```go
if hasID(claims.UserID) {
	return handler(ctx, req)
}
```
and similar for the others that check only equality with an ID.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,150 @@
...
+						groupID := m.FetchID("group")
+						for _, group := range claims.Groups {
+							if group == groupID {
+								return handler(ctx, req)
+							}
+						}
+					}
+				case student:
+					if m, ok := req.(requestID); ok {
```
meling, line 112: Here you can do:
```go
if hasStatus(qf.Enrollment_STUDENT) {
	return handler(ctx, req)
}
```
And similar for the others.

web/interceptor/access_control.go
meling: Maybe it could serve as the `everyone` else role.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,150 @@
...
+	// teacher: user enrolled in the course with teacher status.
+	teacher
+	// courseAdmin: an admin user who is also enrolled into the course.
+	courseAdmin
+	// admin is the user with admin privileges.
+	admin
+)
+
+// If there are several roles that can call a method, a role with the least privilege must come first.
+// If method is not in the map, there is no restrictions to call it.
```
meling, line 37: I think we should add all methods to the map.
And we should panic if the `len(qfServiceInfo.Methods) != len(access)`... and if they are equal length, we should check that both maps contains exactly the same method names... and panic if not. Clearly, this should be done early in the start up process.

Maybe the method could be called: `AccessControlMethodCheck()` or something like that.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,150 @@
...
+var access = map[string]roles{
+	"GetUser":                {},
```
meling, line 39: What does it mean when the `roles` are empty? Should we require to specify `everyone` or something, instead of leaving it empty?

web/interceptor/access_control.go
0xf8f8ff: Wouldn't I need to move enum with roles into the `qf` package and make it exported for this to work? That would be problematic because `User`, `Group`, and other names are already defined there.

web/interceptor/access_control.go
0xf8f8ff: Made it a `none` value, or should it have some other name?

meling: Sounds good to me!

web/interceptor/access_control.go
meling: You are right. However, I found a trick, that I think will work for us...

1. Move / copy these to `qf/roles.go` (I've removed the comments from here; feel free to keep them):

```go
type (
	Role  int
	Roles []Role
)

const (
	none Role = iota
	user
	group
	student
	teacher
	courseAdmin
	admin
)
```

2. Change the interface like this in `access_control.go`:

```go
type requestID interface {
	IDFor(qf.Role) uint64
}
```

3. Update this in `access_control.go`:

```go
const (
	none qf.Role = iota
(the rest of the consts doesn't need to change)
...
var access = map[string]qf.Roles{
...
(and possibly a few other places)
```

This should do the trick. The only thing to keep in mind is that the consts are now in two files that must be synced. It is probably a good idea to make a cross-reference comment pointing to both files.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,157 @@
...
+	"GetOrganization":         {admin},
+	"CreateCourse":            {admin},
+}
+
+func AccessControl(logger *zap.SugaredLogger, tm *auth.TokenManager) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		method := info.FullMethod[strings.LastIndex(info.FullMethod, "/")+1:]
+		logger.Debugf("ACCESS CONTROL for method %s", method) // tmp
+		roles, ok := access[method]
+		if ok {
```
meling, line 88: This should be changed to:
```go
if !ok {
  // error logging and return early...
}
// Then here comes the ok part (avoiding the indentation).
```
This because it is difficult to read the code, when you have to scroll to the bottom to check that the erroneous case is handled appropriately. Is it?

web/interceptor/access_control.go
```diff
@@ -0,0 +1,157 @@
...
+	"GetSubmissionsByCourse":  {teacher, courseAdmin},
+	"GetUserByCourse":         {teacher, admin},
+	"GetUsers":                {admin},
+	"GetOrganization":         {admin},
+	"CreateCourse":            {admin},
+}
+
+func AccessControl(logger *zap.SugaredLogger, tm *auth.TokenManager) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		method := info.FullMethod[strings.LastIndex(info.FullMethod, "/")+1:]
```
meling, line 85: Just to check; since we already have the `req` object here, would it be possible to do:
```go
request, ok := req.(requestID)
if !ok {
	// return early if the request does not implement the interface.
}
```
Or are there `req` types that don't implement the interface? If all expected `req` types were to implement the interface, we could always rely on the type assertion to pass. Then we could use the `request` in the remaining code, allowing us to call `request.IDFor(role)` without further type assertions. This would largely obviate the need for the `hasID()` function.

PS: I would perhaps change the `req interface{}` in the func signature to `request interface{}` and leaving `req` as is after the type assertion:
```go
req, ok := request.(requestID)
```

web/interceptor/access_control.go
```diff
@@ -0,0 +1,157 @@
...
+						courseID := m.IDFor("course")
+						status, ok := claims.Courses[courseID]
+						if ok && status == qf.Enrollment_STUDENT {
+							return handler(ctx, req)
+						}
+					}
+				case teacher:
+					// TODO(vera): needs different handling for "GetUserByCourse"
+					if m, ok := req.(requestID); ok {
+						courseID := m.IDFor("course")
```
meling, line 129: I just noticed that `"course"` is not a role. Hence, it seems like the wrong idea to make the interface be:
```go
type requestID interface {
	IDFor(qf.Role) uint64
}
```
As I proposed above... (need to reconsider my recommendation here...)

web/interceptor/access_control.go
meling: I've given this some more thought, given my discovery late last night that the `IDFor(string)` approach isn't really expecting a `role` as input (there is no `course` role...) Thus, it seems inappropriate to use my proposed approach with `IDFor(qf.Role)`.

What had me confused were these implementations that seem to indicate that these were the same roles as those in `access_control.go`. But they are not.
```go
func (r *GroupRequest) IDFor(role string) uint64 { }

func (r *SubmissionRequest) IDFor(role string) uint64 { }
```

I suggest to leave it as `IDFor(string)` for now. We should probably reconsider this in the context of [this suggestion to merge some proto messages](https://github.com/quickfeed/quickfeed/issues/701#issuecomment-1202409979).

web/interceptor/access_control.go
0xf8f8ff: There are requests like `GetUser`, `GetCourses` that send `Void` message which doesn't have any IDs. We can of course add a dummy `IDFor()` method for this message to avoid making exceptions.

meling: Yes. I think this makes good sense.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,206 @@
...
+				}
+			}
+		}
+		logger.Errorf("%f failed for user %d: insufficient user privileges. Required roles: %v", method, claims.UserID, roles)
+		return nil, ErrAccessDenied
+	}
+}
+
+// hasCourseStatus checks if user is enrolled in a course with a specific status.
+func hasCourseStatus(claims *auth.Claims, courseID uint64, status qf.Enrollment_UserStatus) bool {
```
meling, line 184: This func should be made into a method on `auth.Claims` and the fields of `Claims` should be unexported unless they need to be exposed. 

The following code:
```go
if ok && currentStatus == status {
	return true
}
return false
```
Is equivalent to:
```go
return ok && currentStatus == status
```
But that said, maybe it would be enough to just do:
```go
func (claims Claims) HasCourseStatus(courseID uint64, status qf.Enrollment_UserStatus) bool {
	return claims.courses[courseID] == status
```
You might want to add a few tests, just to be sure.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,206 @@
...
+			return handler(ctx, request)
+		}
+		logger.Debugf("Got claims for user %d\n  courses (%v)\n groups (%v)\n: ", claims.UserID, claims.Courses, claims.Groups) // tmp
+		for _, role := range roles {
+			logger.Debugf("Checking role: %v", role) // tmp
+			switch role {
+			case none:
+				return handler(ctx, request)
+			case user:
+				if req.IDFor("user") == claims.UserID {
```
meling, line 113: Would it be helpful to add something like this to `jwt.go`:
```go
type requestID interface {
	IDFor(string) uint64
}

func (claims Claims) HasID(req requestID) bool {
	return claims.userID == req.IDFor("user")
}
```

web/interceptor/access_control.go
meling: Perhaps this would be a better version:
```go
func (claims Claims) HasCourseStatus(req requestID, status qf.Enrollment_UserStatus) bool {
	courseID := req.IDFor("course")
	return claims.Courses[courseID] == status
}
```

web/users_test.go
```diff
@@ -281,21 +276,6 @@ func TestUpdateUserFailures(t *testing.T) {
...
-	nameChangeRequest := &qf.User{
-		ID:        wantAdminUser.ID,
-		IsAdmin:   false,
-		Name:      "Scrooge McDuck",
-		StudentID: "99",
-		Email:     "test@test.com",
-		AvatarURL: "www.hello.com",
-	}
-	// current user u (non-admin) is in the ctx and tries to change adminUser
-	_, err := ags.UpdateUser(ctx, nameChangeRequest)
```
meling, line 294: Just checking my understanding. This is no longer the right place to check this, since an RPC call would never reach this point because the interceptor would have stopped it.

0xf8f8ff: Correct, it would never reach this point with interceptors. Moreover, in any test in the `web` package any user can call any method because there are no access control checks inside the methods.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,181 @@
...
+						return handler(ctx, request)
+					}
+				}
+			case admin:
+				if claims.Admin {
+					return handler(ctx, request)
+				}
+			}
+		}
+		logger.Errorf("Access denied (%s), required roles %v, user claims %v", method, access[method], claims)
```
JosteinLindhom, line 178: The output from this error is rather hard to parse:

```
Access denied (GetUsers), required roles [6], user claims {{ 1660727622  1660726722 QuickFeed 0 } 2 false map[] []}
```

We could implement the stringer interface on `Claims` and pass claims as `%s` with a sensible format.

0xf8f8ff: Fair point, added the `String()` method for user claims.

assignments/walk_tests_repo.go
```diff
@@ -83,6 +83,12 @@ func readTestsRepositoryContent(dir string, courseID uint64) ([]*qf.Assignment,
 			if err := json.Unmarshal(contents, &benchmarks); err != nil {
 				return nil, "", fmt.Errorf("failed to unmarshal %q: %s", criteriaFile, err)
 			}
+			for _, bm := range benchmarks {
```
meling, line 86: Could you add a comment to explain why this is necessary... Because it may not be obvious even to us in 3-6 months.

main.go
```diff
@@ -120,13 +122,16 @@ func main() {
...
 		grpcServer = web.GRPCServer(unaryOptions, streamOptions)
 	}
-
 	qfService := web.NewQuickFeedService(logger, db, scmManager, bh, runner)
 	if err := qfService.InitSCMs(context.Background()); err != nil {
 		log.Fatalf("Failed to initialize SCM clients: %v", err)
 	}
 
 	qf.RegisterQuickFeedServiceServer(grpcServer, qfService)
+	if err = web.VerifyAccessControlMethods(grpcServer); err != nil {
```
meling, line 131: Minor: would it be possible to call it earlier or must we first call `RegisterQuickFeedServiceServer`?

assignments/walk_tests_repo.go
0xf8f8ff: Done.

main.go
0xf8f8ff: I think we need to register first, but I'll check.

0xf8f8ff: Just like I expected, getting `gRPC server missing qf.QuickFeedService service` without registering (it is after all just an empty gRPC server at this point).

web/interceptor/access_control.go
```diff
@@ -0,0 +1,181 @@
...
+	return func(ctx context.Context, request interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		method := info.FullMethod[strings.LastIndex(info.FullMethod, "/")+1:]
+		req, ok := request.(requestID)
+		// The GetUserByCourse method sends a CourseUserRequest which has no IDs and needs a database query.
+		if !ok {
+			logger.Errorf("%s failed: message type '%s' does not implement IDFor interface",
+				method, reflect.TypeOf(request).String())
+			return nil, ErrAccessDenied
+		}
+		roles, ok := access[method]
```
meling, line 92: Can it actually happen that `ok == false`? Given that we do a pre-check that all methods are in the `access` map. Moreover, the returned `roles` from a false lookup should give an empty slice, which would not enter the `for range roles` loop below, ending at 178 and returning the same `ErrAccessDenied`.

meling: That is, I suggest to remove the `!ok` check and simply do:
```go
roles := access[method]
for _, role := range roles {
...
```
You could even do:
```go
for _, role := range access[method] {
...
```
But maybe that is less clear... Unless the `access` map is renamed to `accessRolesFor[method]` or something.

0xf8f8ff: So something like `for _, role := range access[method]` is pretty safe?

0xf8f8ff: Done.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+		for _, role := range roles {
+			switch role {
+			case none:
+				return handler(ctx, request)
+			case user:
+				if req.IDFor("user") == claims.UserID {
+					// Make sure the user is not updating own admin status.
+					if method == "UpdateUser" {
+						if req.(*qf.User).GetIsAdmin() != claims.Admin {
```
meling, line 105: Tried to replace the `!=` with `==` and the tests still seems to pass. I would have expected it to fail.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,181 @@
...
+							logger.Errorf("Access control: user %d attempted to change admin status from %v to %v",
+								claims.UserID, claims.Admin, req.(*qf.User).GetIsAdmin())
```
meling, line 107: Shouldn't there be a `return nil, ErrAccessDenied` here?

web/interceptor/access_control.go
0xf8f8ff: Interesting. Checking.

web/interceptor/access_control.go
0xf8f8ff: Absolutely. Fixing, adding a test case for this specific situation.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,181 @@
...
+						}
+					}
+					return handler(ctx, request)
+				}
+			case student:
+				// GetSubmissions is used to fetch individual and group submissions.
+				// For individual submissions needs an extra check for user ID in request.
+				if method == "GetSubmissions" && req.IDFor("group") == 0 {
+					if claims.UserID != req.IDFor("user") {
```
meling, line 116: [Suggested before](https://github.com/quickfeed/quickfeed/pull/712/files#r946516332) also, but maybe forgotten; add this to `jwt.go`:
```go
// SameUser returns true if the request corresponds to the user's claims.
func (c *Claims) SameUser(req requestID) bool {
	return req.IDFor("user") == c.UserID
}
```
And do this here:
```go
if !claims.SameUser(req) { ...
```

0xf8f8ff: Done.

web/interceptor/access_control.go
0xf8f8ff: Fixed.

web/interceptor/access_control.go
meling: Can be reused above of course.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,176 @@
...
+		req, ok := request.(requestID)
+		// The GetUserByCourse method sends a CourseUserRequest which has no IDs and needs a database query.
+		if !ok {
+			logger.Errorf("%s failed: message type %T does not implement IDFor interface", method, request)
+			return nil, ErrAccessDenied
+		}
+		claims, err := tm.GetClaims(ctx)
+		if err != nil {
+			logger.Errorf("AccessControl(%s): failed to get claims from request context: %v", method, err)
+			return handler(ctx, request)
```
meling, line 93: Should this really call the `handler` and not return `ErrAccessDenied`? All tests still pass with this code:
```go
	return nil, ErrAccessDenied
```
Seems like we are missing a test for this code path.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,176 @@
...
+
+type (
+	role      int
+	roles     []role
+	requestID interface {
+		IDFor(string) uint64
+	}
+)
+
+const (
```
meling, line 21: Not sure if my rephrasing is accurate, please check. I prefer to line them up with comments on the rhs. You should read it left to right, as in "_none can be anyone_", "_user can access ..._ ", "_group user is enrolled_ ..." and so on.
```go
const (
	none        role = iota // can be anyone.
	user                    // can access its own information.
	group                   // user is enrolled in the course and is member of the given group.
	student                 // user is enrolled in the course with any role.
	teacher                 // user is enrolled in the course with teacher status.
	courseAdmin             // user is enrolled in the course and is also admin.
	admin                   // user has admin privileges.
)
```
Does it make sense?

I'm wondering, do we need the `courseAdmin` role? Is it the same as the `CourseCreatorID`? How do you get this role?

web/interceptor/access_control.go
```diff
@@ -0,0 +1,176 @@
...
+						return nil, ErrAccessDenied
+					}
+					return handler(ctx, request)
+				}
+				if claims.HasCourseStatus(req, qf.Enrollment_TEACHER) {
+					return handler(ctx, request)
+				}
+			case courseAdmin:
+				if claims.Admin {
+					if method == "GetUserByCourse" {
```
meling, line 156: The `GetUserByCourse` method does not list the `courseAdmin` role.

Also, there doesn't seem to be any tests for the `courseAdmin` role, since this doesn't cause a test failure:
```go
case courseAdmin:
	return nil, ErrAccessDenied
```

web/interceptor/helpers.go
```diff
@@ -14,45 +14,65 @@ import (
...
+		return fmt.Errorf("missing required method(s) in access control table: %v", missingMethods)
+	}
+	if len(superfluousMethods) > 0 {
+		return fmt.Errorf("superfluous method(s) in access control table: %v", superfluousMethods)
+	}
+	return nil
+}
+
+// AuthTokenString returns a string with JWT with correct format ("auth=JWT").
+func AuthTokenString(token string) string {
```
meling, line 76: Seems like a more appropriate place for this func would be in `auth` package. So that calls like this:
```go
interceptor.AuthTokenString()
```
become:
```go
auth.TokenString()
```
Also pass the `cookie` directly instead of passing its `.Value`:
```go
func TokenString(cookie *http.Cookie) string {
...
```

internal/qtest/test_helper.go
```diff
@@ -192,6 +193,13 @@ func WithUserContext(ctx context.Context, user *qf.User) context.Context {
 	return metadata.NewIncomingContext(ctx, meta)
 }
 
+// WithAuthCookie returns context containing an authentication cookie with JWT.
+func WithAuthCookie(ctx context.Context, token string) context.Context {
```
meling, line 197: Looks like all uses of this helper uses the `interceptor.AuthTokenString()`. Hence, this func should instead provide the following signature:
```go
func WithAuthCookie(ctx context.Context, cookie *http.Cookie) context.Context {
```

See also my other related comment elsewhere regarding moving `interceptor.AuthTokenString()` to `auth` package.

web/interceptor/helpers.go
```diff
@@ -14,45 +14,65 @@ import (
...
-		return nil, err
-	}
-	claims, err := tm.GetClaims(token)
-	if err != nil {
-		logger.Errorf("Failed to extract claims from JWT: %v", err)
-		return nil, ErrInvalidAuthCookie
-	}
 	if tm.UpdateRequired(claims) {
 		logger.Debug("Updating token for user ", claims.UserID)
 		updatedToken, err := tm.NewAuthCookie(claims.UserID)
```
meling, line 29: Maybe this should be called `updatedCookie`? (and other references to `token` in the strings/comments if any)

web/interceptor/tokens.go
```diff
@@ -0,0 +1,75 @@
...
+	"DeleteGroup": // Group members removed from the group.
+	func(ctx context.Context, tm *auth.TokenManager, msg userIDs) error {
+		if grp, ok := msg.(isGroup); ok {
+			group, err := tm.Database().GetGroup(grp.GetGroupID())
+			if err != nil {
+				return err
+			}
+			return defaultTokenUpdater(ctx, tm, group)
+		}
+		return nil // TODO(meling) This should probably be an error??
```
meling, line 53: Should this return `nil` or an error? With the current tests it is unreachable. That may be fine... but it is not clear to me what should happen if the `isGroup` interface is not implemented.... Seems like a programmer error that we should return an error for...

web/interceptor/tokens.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// TokenRefresher updates list of users who need a new JWT next time they send a request to the server.
+// This method only logs errors to avoid overwriting the gRPC error messages returned by the server.
+func TokenRefresher(logger *zap.SugaredLogger, tm *auth.TokenManager) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		method := info.FullMethod[strings.LastIndex(info.FullMethod, "/")+1:]
+		if tokenUpdateFn, ok := tokenUpdateMethods[method]; ok {
+			if msg, ok := req.(userIDs); ok {
+				log.Println("REFRESHER: ", method) // tmp
```
meling, line 64: Seems reasonable to log this to `logger` instead as:
```go
logger.Debugf("TokenRefresher(%s): %d", method, msg.UserIDs())
```
Or is this useless info that only generates lots of noise in the logs??

web/interceptor/tokens.go
```diff
@@ -0,0 +1,75 @@
...
+
+				if err := tokenUpdateFn(ctx, tm, msg); err != nil {
+					logger.Error(err)
```
meling, line 67: Should this return an error or fall through to 73 and `handler`? Same question for Line 70.

web/interceptor/tokens_test.go
```diff
@@ -0,0 +1,159 @@
...
+	ctx := context.Background()
+	conn, err := grpc.DialContext(ctx, "bufnet", grpc.WithContextDialer(bufDialer), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("Failed to dial bufnet: %v", err)
+	}
+	defer conn.Close()
+
+	client := qf.NewQuickFeedServiceClient(conn)
+
+	adminCookie, err := tm.NewAuthCookie(admin.ID)
```
meling, line 64: Consider if these can be replaced with a helper along the lines to what I did in `access_control_test.go`:
```go
	f := func(t *testing.T, id uint64) context.Context {
		token, err := tm.NewAuthCookie(id)
		if err != nil {
			t.Fatal(err)
		}
		return qtest.WithAuthCookie(ctx, interceptor.AuthTokenString(token.Value))
	}
```
That is, it might be better to make a `qtest` func that does the same; it would need `ctx, t, tm, id` as arguments... so maybe it is easier to just do like above...

web/interceptor/tokens_test.go
```diff
@@ -0,0 +1,159 @@
...
+			ExpiresAt: time.Now().Add(1 * time.Minute).Unix(),
+		},
+	}
+	if tm.UpdateRequired(adminClaims) || tm.UpdateRequired(userClaims) {
+		t.Error("No users should be in the token update list at the start")
+	}
+	if _, err := client.GetUsers(adminCtx, &qf.Void{}); err != nil {
+		t.Fatal(err)
+	}
+	if tm.UpdateRequired(userClaims) {
```
meling, line 93: Shouldn't this be `adminClaims`? Or is this just testing that the `userClaims` wasn't impacted by the above `GetUsers(adminCtx ...` thing?

web/interceptor/tokens_test.go
```diff
@@ -0,0 +1,159 @@
...
+	qf.RegisterQuickFeedServiceServer(s, ags)
+
+	go func() {
+		if err := s.Serve(lis); err != nil {
+			t.Errorf("Server exited with error: %v", err)
+			return
+		}
+	}()
+
+	admin := qtest.CreateFakeUser(t, db, 1)
```
meling, line 52: These two should be moved below `client` on line 62 since they are not used until that time... Ref also to #732 (to be done in a separate PR later).

web/interceptor/tokens_test.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	if !tm.UpdateRequired(adminClaims) {
+		t.Error("Admin must be in the token update list after creating a new course")
+	}
+	qtest.EnrollStudent(t, db, user, course)
+	if _, err := client.CreateGroup(adminCtx, group); err != nil {
+		t.Fatal(err)
+	}
+	if tm.UpdateRequired(userClaims) {
+		t.Error("User should not be in the token update list")
```
meling, line 133: I guess it would be helpful to add context similar to 126 above, e.g., `... after enrolling in course`. Should the check be done right after EnrollStudent on line 128?

web/interceptor/tokens_test.go
```diff
@@ -0,0 +1,159 @@
...
+	if tm.UpdateRequired(userClaims) {
+		t.Error("No users should be in the token update list")
+	}
+	if _, err := client.UpdateUser(adminCtx, user); err != nil {
+		t.Fatal(err)
+	}
+	if !tm.UpdateRequired(userClaims) {
+		t.Error("User must be in the token update list")
+	}
+	if _, err := client.GetUser(userCtx, &qf.Void{}); err != nil {
```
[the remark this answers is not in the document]
meling: Better yet, do like below and add the context in the error messages... like `... after XYZ...`

web/interceptor/access_control_test.go
```diff
@@ -0,0 +1,417 @@
...
+	adminContext := f(t, admin.ID)
+
+	freeAccessTest := accessTests{
+		{"admin", courseAdminContext, 0, course.ID, 0, true},
+		{"student", studentContext, 0, course.ID, 0, true},
+		{"student", groupStudentContext, 0, course.ID, 0, true},
+		{"user", userContext, 0, course.ID, 0, true},
+		{"non-teacher admin", adminContext, 0, course.ID, 0, true},
+	}
+	testUnrestrictedAccess(client, t, freeAccessTest)
```
meling, line 153: Please use subtests instead of testXYZ functions:
```go
	for _, tt := range freeAccessTest {
		t.Run("UnrestrictedAccess/"+tt.name, func(t *testing.T) {
			_, err := client.GetUser(tt.ctx, &qf.Void{})
			verifyAccess(t, err, tt.access, "GetUser", tt.name)
			_, err = client.GetCourse(tt.ctx, &qf.CourseRequest{CourseID: tt.courseID})
			verifyAccess(t, err, tt.access, "GetCourse", tt.name)
			_, err = client.GetCourses(tt.ctx, &qf.Void{})
			verifyAccess(t, err, tt.access, "GetCourses", tt.name)
		})
	}
```
This makes it much easier to identify which tests were run and also allows to run only a selected subtest when debugging, e.g.,:
```sh
% go test -v -run TestAccessControl/Unrestricted
```

PS: Since `tt.name` is now part of the subtest name, it is no longer needed by the `verifyAccess` function.

web/interceptor/access_control_test.go
```diff
@@ -0,0 +1,417 @@
...
+	BufSize = 1024 * 1024
+)
+
+type accessTests []struct {
+	name     string
+	ctx      context.Context
+	userID   uint64
+	courseID uint64
+	groupID  uint64
+	access   bool
```
meling, line 32: Ref my comment regarding the `verifyAccess` func at the bottom of the file...
Would it not be better to replace this `access` field with a:
```go
wantErr error
```
You can simply define alias var something like this:
```go
var deniedErr = interceptor.ErrAccessDenied
```
that you can use in the struct tables to avoid long vars.

meling (changes requested):
Some improvements in the tests are needed, but should be relatively quick to fix. There are also some questions remaining... I still need to carefully check `web/quickfeed_service.go`.

meling:
@JosteinLindhom Please could you also check carefully the `web/quickfeed_service.go`... there probably aren't many issues, but there might be some simplifications in other (unrelated) parts that we should write up as an issue...

web/quickfeed_service.go
```diff
@@ -271,14 +209,13 @@ func (s *QuickFeedService) UpdateEnrollments(ctx context.Context, in *qf.Enrollm
 			if ok, parsedErr := parseSCMError(err); ok {
 				return nil, parsedErr
 			}
-			return nil, status.Error(codes.InvalidArgument, "failed to update enrollment")
+			return nil, err // status.Error(codes.InvalidArgument, "failed to update enrollments")
```
meling, line 212: Is this temporary?

web/interceptor/access_control.go
0xf8f8ff: It is this role from the original access control:
```
// Access policy:
// Admin enrolled in CourseID,
```

web/interceptor/tokens.go
0xf8f8ff: I'm a bit conflicted about returning errors from this interceptor, but probably yes, it is better to return an error in both cases.

web/interceptor/tokens.go
0xf8f8ff: It would be too noisy, removing.

web/interceptor/tokens.go
0xf8f8ff: Agree, returning an error now.

web/interceptor/tokens_test.go
0xf8f8ff: Looks like it is reasonable to check both users, updated.

web/interceptor/tokens_test.go
0xf8f8ff: `EnrollStudent` does not affect the token update list as it is performed directly on the database, but `CreateGroup` shouldn't and must be checked.

web/interceptor/access_control.go
0xf8f8ff: Returning an error now, added test with an empty context.

web/interceptor/helpers.go
0xf8f8ff: Renamed.

web/interceptor/access_control.go
```diff
@@ -0,0 +1,167 @@
...
+	"GetOrganization":         {admin},
+	"CreateCourse":            {admin},
+}
+
+// AccessControl checks user information stored in the JWT claims against the list of roles required to call the method.
+func AccessControl(logger *zap.SugaredLogger, tm *auth.TokenManager) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, request interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		method := info.FullMethod[strings.LastIndex(info.FullMethod, "/")+1:]
+		req, ok := request.(requestID)
+		// The GetUserByCourse method sends a CourseUserRequest which has no IDs and needs a database query.
```
meling, line 85: Is this comment still relevant here? It feels out of place.

web/interceptor/access_control.go
meling: Yeah, I just don't remember why it is needed.
```go
// ensure that current user is teacher or enrolled admin to process the submission request
if !s.hasCourseAccess(usr.GetID(), in.GetCourseID(), func(e *qf.Enrollment) bool {
	switch e.Status {
	case qf.Enrollment_TEACHER:
		return true
	case qf.Enrollment_STUDENT:
		return usr.IsAdmin
	}
	return false
}) {
```
So basically, it is a user enrolled as a student in the course. That student also has `admin` privileges. When is this scenario relevant?

The `courseAdmin` users can only use:
- GetSubmissions
- GetSubmissionsByCourse

I'm wondering if this is some legacy issue that is no longer relevant.

@JosteinLindhom Is this something that the frontend needs?

web/interceptor/helpers.go
```diff
@@ -14,45 +14,60 @@ import (
...
-	}
-	claims, err := tm.GetClaims(token)
-	if err != nil {
-		logger.Errorf("Failed to extract claims from JWT: %v", err)
-		return nil, ErrInvalidAuthCookie
-	}
 	if tm.UpdateRequired(claims) {
-		logger.Debug("Updating token for user ", claims.UserID)
-		updatedToken, err := tm.NewAuthCookie(claims.UserID)
+		logger.Debug("Updating cookie for user ", claims.UserID)
```
meling, line 28: Suggest that this if-clause adjusted like this:
```go
logger.Debug("Updating cookie for user ", claims.UserID)
meta, err = tm.UpdateCookie(ctx, claims)
if err != nil {
	logger.Errorf("Failed to update cookie: %v", err)
	return nil, ErrInvalidAuthCookie
}
```
And the corresponding method in `jwt_update.go` could look something like this:
```go
func (tm *TokenManager) UpdateCookie(ctx context.Context, claims *Claims) (metadata.MD, error) {
	updatedCookie, err := tm.NewAuthCookie(claims.UserID)
	if err != nil {
		return nil, fmt.Errorf("failed to update cookie for user %d: %w", claims.UserID, err)
	}
	if err := tm.Remove(claims.UserID); err != nil {
		return nil, err
	}
	if err := grpc.SendHeader(ctx, metadata.Pairs(SetCookie, updatedCookie.String())); err != nil {
		return nil, fmt.Errorf("failed to set grpc header: %w", err)
	}
	return metadata.New(map[string]string{
		Cookie: TokenString(updatedCookie),
	}), nil
}
```
The only thing I'm not sure about is whether or not to pass `ctx` required for `grpc.SendHeader` or to just return the `updatedCookie`, and leave the metadata stuff in the `getAuthenticatedContext` func, like this:
```go
func (tm *TokenManager) UpdateCookie(ctx context.Context, claims *Claims) (*http.Cookie, error) {
	updatedCookie, err := tm.NewAuthCookie(claims.UserID)
	if err != nil {
		return nil, fmt.Errorf("failed to update cookie for user %d: %w", claims.UserID, err)
	}
	if err := tm.Remove(claims.UserID); err != nil {
		return nil, err
	}
	return updatedCookie, nil
}
```
(also should the `Remove` be done before the `NewAuthCookie`?)

web/interceptor/access_control.go
0xf8f8ff: I don't remember why it existed in the first place, but I think the `courseAdmin` role is not really needed for `GetSubmissions`, at least I can't see what use case it could break. 

But `GetSubmissionsByCourse` is only used by the `approvelist` tool, what kind of permissions are needed there? Should a non-teacher be able to use it?

web/interceptor/access_control_test.go
0xf8f8ff: Well, `wantErr` will only have two possible values: `ErrAccessDenied` and any other error (including nil). Is it really better than a Boolean?

web/interceptor/helpers.go
0xf8f8ff: But what if there is an error while updating a cookie? Then `Remove` would remove user from the token update list, and the user's JWT would never get updated before it expires.

web/interceptor/access_control.go
0xf8f8ff: You are right. Removing.

web/interceptor/access_control.go
meling: `GetSubmissionsByCourse` is also used by the frontend (actions.tsx):
```typescript
    const result = await effects.grpcMan.getSubmissionsByCourse(courseID, SubmissionsForCourseRequest.Type.ALL, true)
    const groups = await effects.grpcMan.getSubmissionsByCourse(courseID, SubmissionsForCourseRequest.Type.GROUP, true)
```
But `approvelist` should only be needed by a teacher.

web/interceptor/helpers.go
meling: Okay. I guess it is just me that didn't recall the details of the Remove func.

web/interceptor/access_control.go
meling: Looks like this has been fixed. Marking as resolved.

web/interceptor/access_control_test.go
meling: Not sure. Let's leave it as is for now.

meling (approved):

PR #721, opened 2022-08-12
Fix to allow docker to run as non-root

### LINKED ISSUE #720: QuickFeed execution user vs docker root user
QuickFeed now clones repositories as the local user running the quickfeed server. Then quickfeed binds the cloned folders into the docker container. However, since the docker container runs as root and in the case when `run.sh` script copies files between the cloned folders (or in general creates files), then the files will be owned by root.

This prevents quickfeed from cleaning up after itself, since it is not running as root.

It is likely that this will leave behind a substantial amount of files in the `/tmp` folder. We need to find a solution for this problem.

meling:
Fixes #720

### DIFF
--- a/.vscode/quickfeed-words.txt
+++ b/.vscode/quickfeed-words.txt
@@ -41,6 +41,7 @@ fullchain
 Getwd
 githubv
 GOBIN
+GOCACHE
 Godkjent
 gofumports
 gofumpt
--- a/ci/docker.go
+++ b/ci/docker.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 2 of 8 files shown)


meling:
This fixes #720 by running the docker container as the quickfeed user, which resulted in a secondary problem in that the official Go docker image runs as root and expects to have its compilation cache stored in `/root/.cache/go-build`. For context, [see this issue on the golang issue tracker.](https://github.com/golang/go/issues/29267)

The second problem occurred because to run as the quickfeed user via setting `User: "uid:gid"`, there isn't really a docker internal user in the container's `passwd`. Hence, there is no HOME folder for the quickfeed user.

The fix was simple enough, once we figured this out. However, while we have a test that can reproduce the second problem, the former problem is difficult to reproduce; and has only been tested manually in a WSL deployment. It is unclear if the problem in #720 would surface on pure Linux deployment. However, it should now be fixed according to manual testing.

PR #730, opened 2022-08-17
Create App from Manifest

JosteinLindhom:
- Implemented GitHub application flow
- Rewrite save, handler logic


web/manifest/manifest.go
```diff
@@ -0,0 +1,200 @@
...
+// TODO(meling) Should reuse those in env package.
+const (
+	defaultPath  = "internal/config/github/quickfeed.pem"
+	appID        = "QUICKFEED_APP_ID"
+	appKey       = "QUICKFEED_APP_KEY"
+	clientID     = "QUICKFEED_CLIENT_ID"
+	clientSecret = "QUICKFEED_CLIENT_SECRET"
+)
+
+type Manifest struct {
```
meling, line 26: Do we need to export this struct type and several of the methods below? Looks like it is only `StartFlow` that is used from the outside?

web/manifest/manifest.go
```diff
@@ -0,0 +1,200 @@
...
+	router.Handle("/manifest/callback", m.Conversion())
+	router.Handle("/manifest", m.CreateApp())
+	m.Server = http.Server{
+		Addr:    addr,
+		Handler: router,
+	}
+	return m
+}
+
+func StartFlow(addr string) error {
```
meling, line 45: Could we use a less generic name? Something like
- `StartAppCreationFlow`
- `StartAppCreation`
- `StartAppFlow`

web/manifest/manifest.go
```diff
@@ -0,0 +1,200 @@
...
+			m.done <- true
+		}
+	}()
+	defer m.Close()
+
+	fmt.Printf("Go to https://%s/manifest to create an app.\n", env.Domain())
+
+	<-m.done
+	// Refresh environment variables
+	return env.Load("")
```
meling, line 61: There is also a refresh Load on Line 114; do we need both?

meling (approved):

PR #736, opened 2022-08-18
Exchange Refresh Token for Access Token

JosteinLindhom:
- Fixed accepting repository invites
- Store refresh token, and exchange for access token

### DIFF
--- a/scm/scm_invite.go
+++ b/scm/scm_invite.go
@@ -13,13 +13,13 @@ type SCMInvite interface {
 
 // RepositoryInvitationOptions contains information on which organization and user to accept invitations for.
 type RepositoryInvitationOptions struct {
-	Login string // GitHub username.
-	Owner string // Name of the organization.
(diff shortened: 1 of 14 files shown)


scm/scm_invite.go
```diff
@@ -0,0 +1,28 @@
...
+type SCMInvite interface {
+	// Accepts repository invites.
+	AcceptRepositoryInvites(context.Context, *RepositoryInvitationOptions) error
+}
+
+// RepositoryInvitationOptions contains information on which organization and user to accept invitations for.
+type RepositoryInvitationOptions struct {
+	Login string // GitHub username.
+	Owner string // Name of the organization.
+	Scm   SCM
```
meling, line 18: Note to self: `Scm` should be renamed to `ScmApp` to signal that it must be an SCM app implementation.

meling (approved):

PR #757, opened 2022-08-28
Minor tweaks and fixes to the frontend

JosteinLindhom:


public/src/components/Members.tsx
```diff
@@ -90,7 +90,7 @@ const Members = (): JSX.Element => {
                     <Search />
                 </div>
                 <div className="ml-auto">
-                    <div className={edit ? "btn btn-sm btn-danger" : "btn btn-sm btn-primary"} onClick={() => setEditing(!edit)}>
+                    <div className={edit ? "btn btn-danger" : "btn btn-primary"} onClick={() => setEditing(!edit)}>
                         {edit ? "Cancel" : "Edit"}
```
meling, line 94: I think the `Cancel` button should be renamed `Done` because if you've accepted or rejected a student, there is no cancelation option.

meling (approved):
Looks good to me! There is a minor button label that could be fixed. And of course, you mentioned there was an error during webpack ... I tried webpack locally, and it compiled just fine.

PR #772, opened 2022-09-06
Added functionality to fetch a single submission

JosteinLindhom:


public/src/overmind/actions.tsx
```diff
@@ -449,8 +449,8 @@ export const getAllCourseSubmissions = async ({ state, actions, effects }: Conte
     state.isLoading = true
 
     // None of these should fail independently.
-    const result = await effects.grpcMan.getSubmissionsByCourse(courseID, SubmissionsForCourseRequest.Type.ALL, true)
-    const groups = await effects.grpcMan.getSubmissionsByCourse(courseID, SubmissionsForCourseRequest.Type.GROUP, true)
+    const result = await effects.grpcMan.getSubmissionsByCourse(courseID, SubmissionsForCourseRequest.Type.ALL, false)
```
meling, line 452: Reminder: We can remove the bool argument, since it is no longer used.

public/src/overmind/actions.tsx
```diff
@@ -667,7 +690,17 @@ export const deleteBenchmark = async ({ actions, effects }: Context, { benchmark
...
+export const refreshSubmission = async ({ effects }: Context, { link }: { link: SubmissionLink.AsObject }): Promise<SubmissionLink.AsObject> => {
+    if (link.submission && link.assignment) {
+        const response = await effects.grpcMan.getSubmission(link.assignment.courseid, link.submission.id)
+        if (success(response) && response.data) {
+            link.submission = response.data.toObject()
+        }
+    }
+    return link
+}
+export const setActiveSubmissionLink = async ({ state, actions }: Context, link: SubmissionLink.AsObject): Promise<void> => {
```
meling, line 702: Add a blank line above here.

qf/quickfeed.proto
```diff
@@ -48,6 +48,7 @@ service QuickFeedService {
 
     // Get latest submissions for all course assignments for a user or a group.
     rpc GetSubmissions(SubmissionRequest) returns (Submissions) {}
+    rpc GetSubmission(SubmissionReviewersRequest) returns (Submission) {}
```
meling, line 51: We should revisit the naming of the request types (in a future PR).

web/interceptor/access_control.go
```diff
@@ -43,6 +43,7 @@ var accessRolesFor = map[string]roles{
 	"UpdateUser":              {user, admin},
 	"GetEnrollmentsByUser":    {user, admin},
 	"GetSubmissions":          {student, group, teacher},
+	"GetSubmission":           {teacher},
```
meling, line 46: Should students be allowed to invoke `GetSubmission`? Or do they only need `GetSubmissions`?

web/quickfeed_service.go
```diff
@@ -361,6 +361,15 @@ func (s *QuickFeedService) DeleteGroup(ctx context.Context, in *connect.Request[
 	return &connect.Response[qf.Void]{}, nil
 }
 
+func (s *QuickFeedService) GetSubmission(ctx context.Context, in *connect.Request[qf.SubmissionReviewersRequest]) (*connect.Response[qf.Submission], error) {
+	submission, err := s.db.GetSubmission(&qf.Submission{ID: in.Msg.GetSubmissionID()})
```
meling, line 365: So server-side does not need courseID?

JosteinLindhom: We need to pass the course ID to pass the `teacher` check in `access_control.go`:

```go
case teacher:
    if method == "GetUserByCourse" {
        if err := claims.IsCourseTeacher(tm.Database(), request.Any().(*qf.CourseUserRequest)); err != nil {
            return nil, connect.NewError(connect.CodePermissionDenied, fmt.Errorf("AccessControl(%s): %w", method, err))
        }
	return next(ctx, request)
    }
    if claims.HasCourseStatus(req, qf.Enrollment_TEACHER) {
	return next(ctx, request)
    }
```
However, a teacher for course X, with claims for course X, will pass this check. I have no checks implemented yet that also check if the supplied submission ID actually belongs to course X. Currently requesting a submission related to course Y, with claims only for X, will still return the requested submission.

This could likely be avoided by looking up and checking the `CourseID` of the assignment related to the requested submission.

web/interceptor/access_control.go
JosteinLindhom: We could allow students to invoke `GetSubmission`. This will require checking that the user ID of the submission matches the user ID in the incoming context. If the submission is for a group assignment, we need to match the user ID with the members of the group.

If we do allow them to invoke the method, we could add a `Refresh` button in the frontend (or some similar functionality) to refresh the submission they are currently viewing.

web/quickfeed_service.go
```diff
@@ -361,6 +361,15 @@ func (s *QuickFeedService) DeleteGroup(ctx context.Context, in *connect.Request[
 	return &connect.Response[qf.Void]{}, nil
 }
 
+func (s *QuickFeedService) GetSubmission(ctx context.Context, in *connect.Request[qf.SubmissionReviewersRequest]) (*connect.Response[qf.Submission], error) {
```
meling, line 364: Should replace `ctx` with `_` to satisfy DeepSource.

public/src/MockGRPCManager.ts
```diff
@@ -456,6 +456,14 @@ export class MockGrpcManager {
         return this.grpcSend<Submissions>(submissions)
     }
 
+    public getSubmission(courseID: number, submissionID: number): Promise<IGrpcResponse<Submission>> {
+        if (!this.enrollments.getEnrollmentsList().find(e => e.getCourseid() === courseID && e.getUserid() === this.currentUser?.getId())) {
```
meling, line 460: I think I would prefer to split this into two lines for clarity. One to find the enrollment `e`, and another for the check that it exists or not.

public/src/MockGRPCManager.ts
```diff
@@ -456,6 +456,14 @@ export class MockGrpcManager {
...
+            return this.grpcSend<Submission>(null, new Status().setCode(StatusCode.NOT_FOUND))
```
meling, line 461: This looks a bit odd to do a send with `null`. Could we instead make a local function that simply returns the expected type necessary for the caller to handle a not found reply.

meling (changes requested):
There are two trivial DeepSource issues and a conflict with master.

web/quickfeed_service.go
```diff
@@ -348,6 +348,15 @@ func (s *QuickFeedService) DeleteGroup(ctx context.Context, in *connect.Request[
 	return &connect.Response[qf.Void]{}, nil
 }
 
+func (s *QuickFeedService) GetSubmission(_ context.Context, in *connect.Request[qf.SubmissionReviewersRequest]) (*connect.Response[qf.Submission], error) {
```
meling, line 351: Please add a doc comment that explain what this returns. We should also say where it is used.

I'm not entusiastic about adding `GetLastSubmission` to the db interface. Right now I'm unable to find a better alternative that keeps the database stuff in the database package. But let's fix the database stuff in another PR.

database/gormdb_submission.go
```diff
@@ -127,6 +127,20 @@ func (db *GormDB) GetSubmission(query *qf.Submission) (*qf.Submission, error) {
 	return &submission, nil
 }
 
+func (db *GormDB) GetLastSubmission(courseID uint64, query *qf.Submission) (*qf.Submission, error) {
```
meling, line 130: Add doc comment.

meling (changes requested):
Minor doc

meling (approved):

PR #773, opened 2022-09-06
Fixed an issue where remaining slipdays would not reach the frontend

### LINKED ISSUE #751: SetSlipDays are called from multiple places and on different types
Ideally, there should be fewer specific places to do this updating. Maybe it should only be done in the `database` package. Needs investigation.

JosteinLindhom:
Fixes #751.

### DIFF
--- a/database/gormdb_enrollment.go
+++ b/database/gormdb_enrollment.go
@@ -105,7 +105,7 @@ func (db *GormDB) UpdateSlipDays(usedSlipDays []*qf.UsedSlipDays) error {
 	return nil
 }
 
-// internalUpdateSlipdays updates or creates UsedSlipDays record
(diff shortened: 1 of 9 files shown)


database/gormdb_enrollment.go
```diff
@@ -88,6 +88,12 @@ func (db *GormDB) getEnrollments(model interface{}, statuses ...qf.Enrollment_Us
 		Find(&enrollments); err != nil {
 		return nil, err
 	}
+	for _, enrollment := range enrollments {
+		if enrollment.GetCourse() == nil {
```
meling, line 92: I don't think this check is necessary. If it is necessary, we should move the check inside the `SetSlipDays` method, and simply return if the provided course is `nil`.

PR #774, opened 2022-09-07
Pass Claims in Context

JosteinLindhom:
- Added user verifier interceptor & `wantCode` field to test structs
- Added helper to set/get claims as/from context value
- Added claims to context in interceptors

### DIFF
--- a/web/auth/jwt.go
+++ b/web/auth/jwt.go
@@ -177,6 +177,16 @@ func (c Claims) Context(ctx context.Context) context.Context {
 	return context.WithValue(ctx, ContextKeyUserID, c.UserID)
 }
 
+func (c *Claims) ClaimsContext(ctx context.Context) context.Context {
+	ctx = c.Context(ctx) // Can be removed if tests are performed "externally" (i.e. with claims cookie in request header).
+	return context.WithValue(ctx, ContextKeyClaims, c)
+}
+
+func ClaimsFromContext(ctx context.Context) (*Claims, bool) {
+	claims, ok := ctx.Value(ContextKeyClaims).(*Claims)
+	return claims, ok
+}
+
 // HasCourseStatus returns true if user has enrollment with given status in the course.
 func (c *Claims) HasCourseStatus(req requestID, status qf.Enrollment_UserStatus) bool {
 	courseID := req.IDFor("course")
--- a/web/auth/jwt.go
+++ b/web/auth/jwt.go
@@ -178,7 +178,7 @@ func (c Claims) Context(ctx context.Context) context.Context {
 }
 
 func (c *Claims) ClaimsContext(ctx context.Context) context.Context {
-	ctx = c.Context(ctx) // Can be removed if tests are performed "externally" (i.e. with claims cookie in request header).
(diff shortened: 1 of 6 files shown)


web/auth/jwt.go
```diff
@@ -177,6 +177,16 @@ func (c Claims) Context(ctx context.Context) context.Context {
 	return context.WithValue(ctx, ContextKeyUserID, c.UserID)
 }
 
+func (c *Claims) ClaimsContext(ctx context.Context) context.Context {
+	ctx = c.Context(ctx) // Can be removed if tests are performed "externally" (i.e. with claims cookie in request header).
```
meling, line 181: Suggest to revise comment to:
```go
// Can be removed if tests are rewritten to be performed as gRPC calls with claims cookie in the request header
```

web/interceptor/access_control_test.go
```diff
@@ -27,6 +27,7 @@ type accessTests []struct {
 	courseID uint64
 	groupID  uint64
 	access   bool
```
meling, line 29: While we are at it, suggest to rename `access` to `wantAccess`.

web/interceptor/access_control_test.go
```diff
@@ -124,27 +126,27 @@ func TestAccessControl(t *testing.T) {
 	adminCookie := f(t, admin.ID)
 
 	freeAccessTest := accessTests{
-		{"admin", courseAdminCookie, 0, course.ID, 0, true},
-		{"student", studentCookie, 0, course.ID, 0, true},
-		{"student", groupStudentCookie, 0, course.ID, 0, true},
-		{"user", userCookie, 0, course.ID, 0, true},
-		{"non-teacher admin", adminCookie, 0, course.ID, 0, true},
-		{"empty context", "", 0, course.ID, 0, false},
+		{"admin", courseAdminCookie, 0, course.ID, 0, true, connect.CodePermissionDenied},
```
meling, line 129: While we are at it, I also suggest that these test tables should use named fields:
```go
	{name: "admin", cookie: courseAdminCookie, userID: 0, courseID: course.ID, groupID: 0, wantAccess: true, wantCode: connect.CodePermissionDenied},
```
For tables for which the `groupID`, `userID` etc isn't relevant, they can be dropped. As in this example.

Moreover, we should also make the tests into a map instead as well. See [here for an example](https://github.com/golang/go/wiki/TableDrivenTests#using-a-map-to-store-test-cases). This means that the `name` field wouldn't be necessary in the struct, but instead be the map key.

web/interceptor/access_control_test.go
```diff
@@ -398,11 +402,10 @@ func requestWithCookie[T any](message *T, cookie string) *connect.Request[T] {
 	return request
 }
 
-func checkAccess(t *testing.T, err error, wantAccess bool, method string) {
+func checkAccess(t *testing.T, err error, wantAccess bool, method string, wantCode connect.Code) {
```
meling, line 405: Suggest reordering the arguments: `t, method, err, wantCode, wantAccess`

meling (approved):

PR #775, opened 2022-09-08
Hardcoded URLs now passed from .env file

JosteinLindhom:
- Added `dotenv` to Webpack to allow injecting values
- Hardcoded URLs are now fetched from process.env

### DIFF
--- a/public/.env-template
+++ b/public/.env-template
@@ -1,0 +1,10 @@
+##################################################
+# Make a copy of this file and rename it to .env #
+################################################ #
+
+# If set, an admin will be taken to this URL rather
+# than the `create organization` page on GitHub.
+QUICKFEED_ENTERPRISE_URL=""
+# The URL an admin will be sent to for installing
+# a GitHub application on a course organization
+QUICKFEED_APP_URL=""
--- a/public/.env-template
+++ b/public/.env-template
@@ -3,4 +3,5 @@
 # A course creator will be sent to this URL to create a GitHub organization for a course.
 QUICKFEED_ORGANIZATION_URL="https://github.com/organizations/plan"
 # The URL for installing the QuickFeed GitHub app on a course organization.
-QUICKFEED_APP_URL="https://github.com/apps/quickfeed-uis"
+# QUICKFEED_APP_URL="https://github.com/apps/quickfeed-uis"
+QUICKFEED_APP_URL=""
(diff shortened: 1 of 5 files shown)


public/src/components/admin/CourseCreationInfo.tsx
```diff
@@ -11,13 +11,13 @@ const CourseCreationInfo = (): JSX.Element => {
                 This is to keep the student roster for the different runs of the course separate.
             </p>
             <p>
-                <a className="badge-pill badge-success" href="https://github.com/enterprises/university-of-stavanger" target="_blank" rel="noopener noreferrer">
+                <a className="badge-pill badge-success" href={process.env.QUICKFEED_ENTERPRISE_URL ?? "https://github.com/organizations/plan"} target="_blank" rel="noopener noreferrer">
```
meling, line 14: See my revised `.env-template` file. Using a default value in the `.env` file avoids the hardcoded link here and the `??` thing.

Note: I renamed the var to `QUICKFEED_ORGANIZATION_URL` instead of enterprise.

public/.env-template
```diff
@@ -0,0 +1,6 @@
+# Make a copy of this file to .env and update the variables as needed.
+
+# A course creator will be sent to this URL to create a GitHub organization for a course.
+QUICKFEED_ORGANIZATION_URL="https://github.com/organizations/plan"
```
meling, line 4: I think it makes sense to include this link as the default.

public/.env-template
```diff
@@ -0,0 +1,6 @@
...
+# The URL for installing the QuickFeed GitHub app on a course organization.
+QUICKFEED_APP_URL="https://github.com/apps/quickfeed-uis"
```
meling, line 6: Not sure it makes sense to have this as the default, so perhaps we should remove this default line. But I guess other deployments could still us ours, or am I just not thinking straight... (may well be the case as this hour)

JosteinLindhom: I'm not sure if a separate deployment of QuickFeed could use our app. The app is configured with our endpoints, meaning all OAuth logins would go through us. Additionally, their local instance would not have access to use the SCM package without the keys for the app.

meling: Ok; then perhaps we should just remove the URL. Or we could leave it in, and just add a comment line that it must be updated for another deployment. Unless you think this might cause problems, if someone forgets to change it...

PR #780, opened 2022-09-19
Added support for adding certificates to keychain

### LINKED ISSUE #779: Remove unused command line arguments in main.go
We really don't use most of the command line arguments, since we started using `.env` for real. Remove the unused ones and make sure that `$DOMAIN` and `httpAddr` is set correctly.

meling:
- Removed domain from NewTokenManager
- Added support for adding cert to keychain on macOS
- Fixed CommonName cert problem
- Removed service.url argument
- Removed service.url from docs and Makefile
- Removed unused env config in .env-template

Fixes #779.


main.go
```diff
@@ -63,18 +62,14 @@ func main() {
...
 
 	var srvFn web.ServerType
 	if *dev {
 		srvFn = web.NewDevelopmentServer
-		log.Printf("Starting QuickFeed in development mode on %s", *httpAddr)
 	} else {
 		srvFn = web.NewProductionServer
-		log.Printf("Starting QuickFeed in production mode on %s", *baseURL)
 	}
+	log.Printf("Starting QuickFeed on %s", env.Domain())
```
0xf8f8ff, line 72: Could we also log the port, at least if it differs from the default one (:443)?

PR #786, opened 2022-09-21
Implemented a foundation for gRPC streaming

JosteinLindhom:


web/interceptor/user_auth.go
```diff
@@ -11,36 +11,76 @@ import (
...
+func NewUserInterceptor(logger *zap.SugaredLogger, tm *auth.TokenManager) *UserInterceptor {
+	return &UserInterceptor{
+		tm:     tm,
+		logger: logger,
+	}
+}
+
+func (u *UserInterceptor) WrapStreamingHandler(next connect.StreamingHandlerFunc) connect.StreamingHandlerFunc {
+	return connect.StreamingHandlerFunc(func(ctx context.Context, conn connect.StreamingHandlerConn) error {
+		cookie := conn.RequestHeader().Get(auth.Cookie)
```
meling, line 28: Most of this code is similar to `WrapUnary`. Can we make a shared function that can be called from both?

web/interceptor/user_auth.go
```diff
@@ -11,36 +11,76 @@ import (
...
+			conn.ResponseHeader().Set(auth.SetCookie, updatedCookie.String())
+		}
+
+		return next(claims.ClaimsContext(ctx), conn)
+	})
+}
+
+func (*UserInterceptor) WrapStreamingClient(next connect.StreamingClientFunc) connect.StreamingClientFunc {
+	return connect.StreamingClientFunc(func(ctx context.Context, spec connect.Spec) connect.StreamingClientConn {
+		return next(ctx, spec)
```
meling, line 52: I guess we don't support streaming from clients, but should we instead return an error? I'm not sure how, though, since the signature doesn't have an error return type.

Moreover, I guess the same applies to all interceptors for streaming clients.

web/interceptor/tokens.go
```diff
@@ -49,27 +49,45 @@ var tokenUpdateMethods = map[string]func(context.Context, *auth.TokenManager, us
...
+	})
+}
+
+func (*TokenInterceptor) WrapStreamingClient(next connect.StreamingClientFunc) connect.StreamingClientFunc {
+	return connect.StreamingClientFunc(func(ctx context.Context, spec connect.Spec) connect.StreamingClientConn {
+		return next(ctx, spec)
+	})
+}
+
 // TokenRefresher updates list of users who need a new JWT next time they send a request to the server.
```
meling, line 76: Method no longer called `TokenRefresher`. Should be updated.

web/interceptor/access_control.go
```diff
@@ -76,89 +76,107 @@ var accessRolesFor = map[string]roles{
...
+	})
+}
+
+func (*AccessControlInterceptor) WrapStreamingClient(next connect.StreamingClientFunc) connect.StreamingClientFunc {
+	return connect.StreamingClientFunc(func(ctx context.Context, spec connect.Spec) connect.StreamingClientConn {
+		return next(ctx, spec)
+	})
+}
+
 // AccessControl checks user information stored in the JWT claims against the list of roles required to call the method.
```
meling, line 99: Method name mismatch with doc.

web/stream/pool.go
```diff
@@ -0,0 +1,86 @@
...
+	// Create a new channel.
+	p.channels[id] = make(chan *T)
+	return p.channels[id]
+}
+
+// Remove removes a channel from the pool.
+func (p *Pool[T]) Remove(id uint64) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	if ch, ok := p.channels[id]; ok {
```
meling, line 39: Since this is called from `Add`, before taking the lock in `Add`, it will take the lock here in `Remove` and release it again. Then in `Add` it will take the lock again. Not sure this is what you want, leaving to another goroutine to grab the lock in-between. Also, it adds pressure on the lock.

Suggested alternative; implement unexported `remove` without locking. Then require that callers of `remove` must already hold the lock before calling.

```go
func (p *Pool[T]) remove(id uint64) {
	if ch, ok := p.channels[id]; ok {
		close(ch)
 	}
 	delete(p.channels, id)
```

meling: That is, `Remove` (exported) and `Add` must hold the lock before calling `remove` (unexported).

web/stream/pool.go
```diff
@@ -0,0 +1,86 @@
+package stream
+
+import (
+	"errors"
+	"sync"
+)
+
+// Pool is a pool of channels.
+// The channels are used to send data to connected clients.
+type Pool[T any] struct {
```
meling, line 10: The `Pool` type should have a `pool_test.go` that tests the API, including adding and removing users, sending stuff on its channels and more.

Also, make sure that all methods in the API are used (didn't check yet). If we don't need, e.g., `CloseBy`, we shouldn't add it until we need it.

meling: Do we need to export this type and the methods? Looks like they are only being used in the `stream` package.

web/stream/pool.go
meling: If we don't want to export `Remove`, we should call it `remove`, and the other one should be called `internalRemove` to signal that we should be used with care (and documentation should explain that it requires holding the lock.)

Another possible name is `noLockRemove`.

web/stream/service.go
```diff
@@ -0,0 +1,113 @@
...
+// CloseBy closes a single stream by ID.
+func (s *Service[T]) CloseBy(id uint64) error {
+	if stream, ok := s.streams[id]; ok {
+		stream.Close()
+	}
+	return nil
+}
+
+// GetStream returns a stream by ID.
+func (s *Service[T]) GetStream(id uint64) (*Stream[T], error) {
```
meling, line 100: Needed?

web/stream/service.go
```diff
@@ -0,0 +1,113 @@
...
+		errs = append(errs, err)
+	}
+	return multierr.Join(errs...)
+}
+
+// Add adds a new stream to the service.
+// It returns the stream which must be run by the caller.
+func (s *Service[T]) Add(userID uint64, st *connect.ServerStream[T]) *Stream[T] {
+	// Delete the stream if it already exists.
+	s.Remove(userID)
```
meling, line 64: Same issue as for the `Pool.Add()` method.

web/stream/pool.go
JosteinLindhom: Also, `CloseBy` does not make sense to keep after giving it a second thought. I cannot see any cases where we would want to only close a channel (signalling to a stream that it should close), but still keep it in our map of channels.

web/stream/stream_test.go
```diff
@@ -0,0 +1,108 @@
...
+	if counter != 90 {
+		t.Errorf("expected 90, got %d", counter)
+	}
+
+	for _, s := range streams {
+		msgMsp := make(map[string]int)
+		for _, data := range s.Messages {
+			msgMsp[data.Msg]++
+		}
+		t.Log(msgMsp)
```
meling, line 72: Once we feel comfortable that the test is working as expected, maybe we should remove the printing.

web/stream/stream_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	wg := sync.WaitGroup{}
+	for i := 1; i < 10; i++ {
+		st := service.AddStream(uint64(1), newMockStream[Data](ctx, uint64(1), &counter))
+		streams = append(streams, st.(*mockStream[Data]))
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			err := st.Run()
+			t.Log(err)
```
meling, line 47: Once we feel comfortable that the test is working as expected, maybe we should remove the printing.

meling (changes requested):

web/stream/service.go
```diff
@@ -0,0 +1,110 @@
...
+	for _, userID := range userIDs {
+		stream, ok := s.streams[userID]
+		if !ok {
+			continue
+		}
+		stream.Send(data)
+	}
+}
+
+// Add adds a new stream to the service.
```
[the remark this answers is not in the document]
JosteinLindhom: `AddStream` has been removed. `Add` now accepts a `StreamInterface[T]`, and no longer returns a stream.
The caller of `Add` is now responsible for creating and passing a `StreamInterface`, and subsequently running it.

meling (approved):

PR #787, opened 2022-09-21
Minor refactoring of test helper methods.

0xf8f8ff:
- Refactor StartGrpcAuthServer to be used by several tests that use different interceptor combinations
- Combine two similar methods that created a test version of an SCMManager into one
- Rename the TestSCMmanager test that had the same name as a test helper method


ci/run_tests_test.go
```diff
@@ -86,7 +86,7 @@ func TestRunTests(t *testing.T) {
 	ctx, cancel := runData.Assignment.WithTimeout(2 * time.Minute)
 	defer cancel()
 
-	scm, _ := qtest.FakeProviderMap(t)
+	scm, _ := qtest.TestSCMManager(t)
```
meling, line 89: Since this is in the `qtest` package, I think it would make sense to just name it `SCMManager`. On the other hand, I see that we often use only the `scm` value and not the manager itself. Perhaps we should have separate test helpers for these two ... but didn't check what is needed in current tests.

0xf8f8ff: I thought about splitting, but we need only the scm twice across all tests, both the scm and the manager twice, and only the manager four times, so I'm still not sure if it makes sense.

scm/test_helper.go
```diff
@@ -2,24 +2,30 @@ package scm
...
 import (
 	"context"
+	"testing"
 
 	"github.com/beatlabs/github-auth/app"
+	"github.com/quickfeed/quickfeed/internal/env"
 )
 
-func TestSCMManager() *Manager {
+// TestSCMManager sets the current provider to "fake", creates a "test" organization
```
meling, line 11: "fake" seems to be "test" now. So the comment is wrong. But should we use `test-name` and `test-path` or something simpler to signal which fields the variable is referring to? With `"test"` in both values it may be easy to mix things up.

scm/test_helper.go
```diff
@@ -2,24 +2,30 @@ package scm
...
+func TestSCMManager(t *testing.T) (SCM, *Manager) {
+	t.Helper()
+	env.SetFakeProvider(t)
 	conf := &Config{
 		"test",
 		"test",
 		&app.Config{},
 	}
 	sc := NewFakeSCMClient()
 	_, _ = sc.CreateOrganization(context.Background(), &OrganizationOptions{
```
meling, line 22: Since `CreateOrganization` is only used by the fake (mock) provider, maybe we could simply remove it and instead just return a "mocked" organization via the `GetOrganization` that may be initialized via NewFakeSCM if necessary.

If you agree, then we can fix this as a separate PR, but would be good to create an issue for it first that can be closed after together with the PR.

scm/test_helper.go
0xf8f8ff: But it calls `env.SetFakeProvider` which sets `provider = 'fake'`, or am I missing something?

I am currently a bit confused about the `Path` field in organization-related things. Looks like it can be safely renamed to `Name` in the `Organization` type. And in the `OrganizationOptions` it is only used by GitLab's `CreateOrganization` method and we can probably just ignore this field in test helpers and fake scm.

web/grpc_auth_test.go
```diff
@@ -48,19 +41,12 @@ func TestGrpcAuth(t *testing.T) {
...
-	if err != nil {
-		t.Fatalf("failed to connect to grpc server: %v", err)
-	}
-	defer conn.Close()
-
 	client := qtest.QuickFeedClient("")
 
+	ctx := context.Background()
 	// create request context with the helpbot's secret token
 	reqCtx := metadata.NewOutgoingContext(ctx,
```
meling, line 51: Still needed?

scm/test_helper.go
0xf8f8ff: Agree. I am working on the mock scm client right now, and this adjustment looks like a good match for the same PR. Will create a related issue.

scm/test_helper.go
```diff
@@ -2,24 +2,30 @@ package scm
...
+	"testing"
 
 	"github.com/beatlabs/github-auth/app"
+	"github.com/quickfeed/quickfeed/internal/env"
 )
 
-func TestSCMManager() *Manager {
+// TestSCMManager sets the current provider to "fake", creates a "test" organization
+// and a fake scm client for this organization.
+func TestSCMManager(t *testing.T) (SCM, *Manager) {
```
meling, line 13: I think the test helper should be called something different; the `Test` prefix should be reserved for actual tests, especially for things like this one, since it takes a `testing.T` as an argument.

scm/test_helper.go
0xf8f8ff: GitLab scm client also uses the same method, is it safe to remove?

scm/test_helper.go
0xf8f8ff: MockSCMManager?

scm/test_helper.go
meling: Yes. I think we should remove it. If we add support for GitLab in the future, perhaps their APIs have changed so much that we have to do other things anyway, e.g., maybe they also have support for Apps now?

scm/test_helper.go
meling: Yes, that's fine.

web/grpc_auth_test.go
0xf8f8ff: Not really. Updated.

meling (approved):

PR #788, opened 2022-09-21
Add qcm tool for local testing of run.sh scripts and updates for cloning assignments repo

### LINKED ISSUE #750: Student code that doesn't compile results in missing tests in the results view
Currently, if a student submits code that doesn't compile, the Tests and corresponding scores may not show up in the results view.

One idea to fix this is to build the tests with a custom version with the list of tests; they could be in the tests repo, protected by a special build tag. These tests should ideally give 0 score on all tests, as previously supported with the init stuff.

There isn't actually something to do in the kit/score package, but we could add tests of the functionality there, I think. That is, the functionality must be supported for each lab/course.

Not sure if the same problem exists for other courses that use python.

PS: A build failure can easily be caused by adding a build tag (not activated by the run.sh script) to student submitted code.

### LINKED ISSUE #761: [Tool] Testing submitted code locally
We should have a tool that can:
- Fetch the tests repository for a given course
- Fetch the user repository for a given user
- Combine these into a single folder locally

We currently have functionality that clone both repositories mentioned above in `run_tests.go: cloneRepositories`.
Joining of these folders is currently done via a script file found in the tests folder.

This could be beneficial in the process of debugging submitted code and provided tests.

meling:
This adds support for a new tool, `qcm`, for local testing of `run.sh` scripts. Additionally, it adds support for cloning the assignments folder, which will enable `run.sh` scripts to initialize the `MaxScore` for all tests before running tests on student code, thus preventing students from disabling tests by making them not compile.

Fixes #761
Fixes #750

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ public/public/index.html
 public/public/styles.css
 public/assets/index.html
 TODO
+x
--- a/.vscode/quickfeed-words.txt
+++ b/.vscode/quickfeed-words.txt
@@ -1,5 +1,6 @@
 addstore
 agpath
+alecthomas
 allstudents
 allteachers
 apitest
--- a/Makefile
+++ b/Makefile
@@ -143,6 +143,9 @@ webpack-dev-server:
 selenium:
 	@cd public && npm run test:selenium
 
+qcm:
+	@cd cmd/qcm; go install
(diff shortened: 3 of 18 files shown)


meling:
Superseded by #804

PR #797, opened 2022-09-27
Updating mock SCM client.

0xf8f8ff:


scm/github.go
```diff
@@ -335,16 +327,7 @@ func (s *GithubSCM) CreateHook(ctx context.Context, opt *CreateHookOptions) erro
...
-	// prioritize creating an organization hook
-	if opt.Organization != "" {
-		_, _, err = s.client.Organizations.CreateHook(ctx, opt.Organization, hook)
-		if err != nil {
-			return fmt.Errorf("CreateOrgHook: failed to create GitHub hook for org %s: %w", opt.Organization, err)
-		}
-	} else {
-		_, _, err = s.client.Repositories.CreateHook(ctx, opt.Repository.Owner, opt.Repository.Path, hook)
-	}
+	_, _, err := s.client.Organizations.CreateHook(ctx, opt.Organization, hook)
```
meling, line 330: Should have a comment about this being a course organization hook. Maybe even at the top level (interface) comment.

scm/test_helper.go
```diff
@@ -18,11 +17,7 @@ func MockSCMManager(t *testing.T) (SCM, *Manager) {
 		"test",
 		&app.Config{},
 	}
-	sc := NewFakeSCMClient()
```
meling, line 21: We should avoid using non-descriptive names such as `test` for arguments/fields because it can difficult to know which field cause a problem.

web/courses_test.go
```diff
@@ -26,7 +26,7 @@ var allCourses = []*qf.Course{
 		Tag:              "Spring",
 		Provider:         "fake",
 		OrganizationID:   1,
-		OrganizationPath: "test",
+		OrganizationName: "test",
```
meling, line 29: Nitpicking: I would prefer also here to use more descriptive names than `test` and `test-1` and so on. Suggest to use dat or qf prefixes…

0xf8f8ff: Changed the first organization that is used in all tests to `qfTestOrg`, the rest are only used in combination with the predefined set of test courses, so I named them based on the course code for these courses. Does it sound reasonable?

scm/github.go
```diff
@@ -335,16 +327,7 @@ func (s *GithubSCM) CreateHook(ctx context.Context, opt *CreateHookOptions) erro
...
 	}
-	var err error
-	// prioritize creating an organization hook
-	if opt.Organization != "" {
-		_, _, err = s.client.Organizations.CreateHook(ctx, opt.Organization, hook)
-		if err != nil {
-			return fmt.Errorf("CreateOrgHook: failed to create GitHub hook for org %s: %w", opt.Organization, err)
-		}
-	} else {
-		_, _, err = s.client.Repositories.CreateHook(ctx, opt.Repository.Owner, opt.Repository.Path, hook)
```
JosteinLindhom, line 346: As we no longer use `opt.Repository` for GitHub we can remove it from `TestCreateHook` in `github_test.go`

```go
// github_test.go line 92:96
opt := &scm.CreateHookOptions{
	URL:        serverURL,
	Secret:     secret,
	Repository: &scm.Repository{Owner: qfTestOrg, Path: "tests"},
}
```
and instead use `opt.Organization`. This way `opt.Valid()` (in the context of this specific test) will return `true` due to `opt.Organization != ""` rather than the repository part of the check:

```go
opt.URL != "" &&
		(opt.Organization != "" || (opt.Repository != nil &&
			opt.Repository.valid()))
```

scm/mock.go
```diff
@@ -0,0 +1,408 @@
...
+	}
+	org, err := s.GetOrganization(ctx, &GetOrgOptions{
+		ID:   opt.Organization.ID,
+		Name: opt.Organization.Name,
+	})
+	if err != nil {
+		return nil, err
+	}
+	repo := &Repository{
+		ID:      uint64(len(s.Repositories) + 1),
```
JosteinLindhom, line 108: Assigning ID like this might end up overwriting another repository:

```go
// s.Repositories
{
    1: repo1
    2: repo2
    3: repo3
}
// Delete repo2
{
    1: repo1
    3: repo3
}
// Create repo, len(s.Repositories) + 1 == 3
{
    1: repo1
    3: repo4
}

Comment (JosteinLindhom) [reply scm/mock.go:108]: In the frontend I ended up with something similar to this for generating IDs:

```go
var IDs = make(map[string]uint64, 0)

func nextID(key string) uint64 {
	if _, ok := IDs[key]; !ok {
		IDs[key] = 1
	} else {
		IDs[key]++
	}
	return IDs[key]
}

nextID("repo") // 1
nextID("repo") // 2
nextID("teams") // 1
// Delete a repository 
nextID("repo") // 3
```

0xf8f8ff: Agree, even if it worked without issues all these years it can still cause unexpected problems for new tests. Same for `Hooks` and `Teams`.

0xf8f8ff: Added a safe ID generator, thank you for the inspiration.

scm/github.go
meling: Let's remove dependencies on gitlab.

meling (approved):

PR #805, opened 2022-10-03
Use testdata folder in mock SCM Clone method

### LINKED ISSUE #799: Clone should use testdata folder with content
The `testdata` folder can be used to hold test data, such as files and folders useful for testing. The go compiler knows about `testdata` and will ignore it when compiling. See `assignments/testdata` for an example. Adding this would allow `MockSCM.Clone()` to use the `testdata` instead of the current approach.

0xf8f8ff:
Resolves #799.


scm/mock.go
```diff
@@ -39,16 +39,21 @@ func NewMockSCMClient() *MockSCM {
 	return s
 }
 
-func (MockSCM) Clone(_ context.Context, opt *CloneOptions) (string, error) {
-	cloneDir := filepath.Join(opt.DestDir, repoDir(opt))
+func (s MockSCM) Clone(ctx context.Context, opt *CloneOptions) (string, error) {
```
meling, line 42: What I meant in #799 was that we should create and populate the `testdata` folder with some content and add it to GitHub, such as in `ci` or `assignments`, so that we could remove the hack below that creates a directory.

internal/qtest/mock_data.go
```diff
@@ -2,6 +2,10 @@ package qtest
 
 import "github.com/quickfeed/quickfeed/qf"
 
+var (
```
meling, line 5: Should be a `const`.

scm/mock.go
meling: That is, when mocking SCM, we could provide some "cloned" content that other tests can rely on.

meling (changes requested):

scm/mock.go
0xf8f8ff: Right, my `scm/testdata` didn't get pushed to GitHub because the folders were empty, added some files now.

0xf8f8ff:
> Not sure what is necessary, but this PR fails the tests:
> 
> ```shell
> --- FAIL: TestRunTests (0.27s)
>     run_tests_test.go:92: failed to clone "tests" repository: organization not found
> FAIL
> FAIL	github.com/quickfeed/quickfeed/ci	38.070s
> 
> --- FAIL: TestRebuildSubmissions (0.04s)
>     rebuild_test.go:165: Failed to rebuild submission: rpc error: code = InvalidArgument desc = failed to rebuild submission failed to clone "tests" repository: organization not found
> FAIL
> FAIL	github.com/quickfeed/quickfeed/web	2.558s
> ```
> 
> Note that I have also modified `TestRebuildSubmissions` and maybe code related to `TestRunTests` in #804. However, these test failures are unrelated to those changes since they are not merged yet. But you may want to check my changes there just in case...

Hmm, I don't get it. Both tests pass for me.

meling:
> > Note that I have also modified `TestRebuildSubmissions` and maybe code related to `TestRunTests` in #804. However, these test failures are unrelated to those changes since they are not merged yet. But you may want to check my changes there just in case...
> 
> Hmm, I don't get it. Both tests pass for me.

I have env variables set, such as 
```go
QF_TEST_ORG=qf101
QF_TEST_USER=meling
GITHUB_ACCESS_TOKEN...
```
When I run without these vars set:
```shell
=== RUN   TestRebuildSubmissions
    rebuild_test.go:77: This test requires that the 'QF_TEST_ORG' is set and that you have access to said GitHub organization
--- SKIP: TestRebuildSubmissions (0.03s)
=== RUN   TestRunTests
    run_tests_test.go:48: This test requires that the 'QF_TEST_ORG' is set and that you have access to said GitHub organization
--- SKIP: TestRunTests (0.00s)
```

0xf8f8ff:
Oh, I see the problem now. My `QF_TEST_ORG` variable was set to the name of the mock test organization.

I wonder if both tests really need an option to run with a real GitHub organization name. I am pretty sure the `TestRebuildSubmissions` can just use the mock organization (instead of `scm.GetTestOrganization(t)` which will make the test skip if this variable is not set). 
Not so sure about the `TestRunTests`. Is it desirable to be able to run the test with real GitHub info?

meling:
Let's make both those tests use the mock organization (and not use `scm.GetTestOrganization(t)`). Instead we should add more tests that actually invoke `RunTests` on "simulated" course tests.

meling (approved):

Issue #806, opened 2022-10-03
Remaining access control checks in web package should be moved to the access control interceptor.


0xf8f8ff:
There are `isValidSubmission` and `isValidSubmissionRequest` methods in the `web` package that check that the user ID in the request has access to the course. The access control interceptor should perform this check instead.

PR #807, opened 2022-10-03
Replace grpc/status and grpc/code with connect equivalents.

### LINKED ISSUE #746: We still use the grpc/status and grpc/codes packages
Maybe it would be better to replace these with their connect equivalents and corresponding error handling.

0xf8f8ff:
Resolves #746.


database/gormdb.go
```diff
@@ -20,7 +19,7 @@ var (
 	ErrEmptyGroup = errors.New("cannot create group without users")
 	// ErrDuplicateGroup is returned when trying to create a group with the same
 	// name as a previously registered group.
-	ErrDuplicateGroup = status.Error(codes.InvalidArgument, "group with this name already registered")
+	ErrDuplicateGroup = connect.NewError(connect.CodeInvalidArgument, errors.New("group with this name already registered"))
```
meling, line 22: Seems like we are checking for duplicate groups in both the database package and the web package. See `web/groups.go`:
```go
func (s *QuickFeedService) checkGroupName(courseID uint64, groupName string) error {
```
I don't think this error would ever happen, given the other check. I suggest that we just make it a normal error, since it would not be pushed out to the frontend anyway.

web/access_control.go
```diff
@@ -2,15 +2,14 @@ package web
...
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
-
+	"github.com/bufbuild/connect-go"
 	"github.com/quickfeed/quickfeed/qf"
 	"github.com/quickfeed/quickfeed/web/auth"
 )
 
-var ErrMissingInstallation = status.Error(codes.PermissionDenied, "github application is not installed on the course organization")
+var ErrMissingInstallation = connect.NewError(connect.CodePermissionDenied, errors.New("github application is not installed on the course organization"))
```
meling, line 12: This is use several places where we need to use the `scm`. Suggest that instead the relevant `getSCM*` functions return this error directly so that we can just `return nil, err` instead of:
```go
	return nil, ErrMissingInstallation
```
There is also one case in `rebuild.go` where we don't use this error, but probably should. But would be solved by returning the error from the source directly.

web/quickfeed_service.go
```diff
@@ -69,7 +67,7 @@ func (s *QuickFeedService) GetUserByCourse(_ context.Context, in *connect.Reques
 	user, err := s.db.GetUserByCourse(query, in.Msg.UserLogin)
 	if err != nil {
 		s.logger.Errorf("GetUserByCourse failed: %v", err)
-		return nil, status.Error(codes.FailedPrecondition, "failed to get student information")
+		return nil, connect.NewError(connect.CodeFailedPrecondition, errors.New("failed to get student information"))
```
meling, line 70: Should this be `InvalidArgument` instead?

web/quickfeed_service.go
```diff
@@ -84,7 +82,7 @@ func (s *QuickFeedService) UpdateUser(ctx context.Context, in *connect.Request[q
 	}
 	if _, err = s.updateUser(usr, in.Msg); err != nil {
 		s.logger.Errorf("UpdateUser failed to update user %d: %v", in.Msg.GetID(), err)
-		return nil, status.Error(codes.InvalidArgument, "failed to update user")
+		return nil, connect.NewError(connect.CodeInvalidArgument, errors.New("failed to update user"))
```
meling, line 85: Should this be `InvalidArgument` instead?

web/quickfeed_service.go
meling: Or `NotFound`?

web/quickfeed_service.go
```diff
@@ -391,12 +389,12 @@ func (s *QuickFeedService) GetSubmissionsByCourse(_ context.Context, in *connect
 func (s *QuickFeedService) UpdateSubmission(_ context.Context, in *connect.Request[qf.UpdateSubmissionRequest]) (*connect.Response[qf.Void], error) {
 	if !s.isValidSubmission(in.Msg.SubmissionID) {
 		s.logger.Errorf("UpdateSubmission failed: submission author has no access to the course")
-		return nil, status.Error(codes.PermissionDenied, "submission author has no course access")
+		return nil, connect.NewError(connect.CodeInvalidArgument, errors.New("submission author has no course access"))
```
meling, line 392: Should this be `PermissionDenied`? It got changed to `InvalidArgument`, but in `RebuildSubmissions()` we still use `PermissionDenied`. That said, ideally, `PermissionDenied` should be handled in the interceptors, if we have sufficient information to do so.

web/quickfeed_service.go
```diff
@@ -409,17 +407,17 @@ func (s *QuickFeedService) RebuildSubmissions(_ context.Context, in *connect.Req
 		// Submission ID > 0 ==> rebuild single submission for given CourseID and AssignmentID
 		if !s.isValidSubmission(in.Msg.GetSubmissionID()) {
 			s.logger.Errorf("RebuildSubmission failed: submitter has no access to the course")
-			return nil, status.Error(codes.PermissionDenied, "submitter has no course access")
+			return nil, connect.NewError(connect.CodePermissionDenied, errors.New("submitter has no course access"))
 		}
 		if _, err := s.rebuildSubmission(in.Msg); err != nil {
 			s.logger.Errorf("RebuildSubmission failed: %v", err)
-			return nil, status.Error(codes.InvalidArgument, "failed to rebuild submission "+err.Error())
+			return nil, connect.NewError(connect.CodeInvalidArgument, errors.New("failed to rebuild submission "+err.Error()))
```
meling, line 414: Don't use `err.Error()` like this. Use:
```go
fmt.Errorf("failed to rebuild submission: %w", err)
```
Same issue applies to Line 420.

meling (changes requested):

web/courses_test.go
```diff
@@ -94,7 +92,7 @@ func TestNewCourseExistingRepos(t *testing.T) {
 	if course != nil {
 		t.Fatal("expected CreateCourse to fail with AlreadyExists")
 	}
-	if err != nil && status.Code(err) != codes.FailedPrecondition {
+	if err != nil && connect.CodeOf(err) != connect.CodeFailedPrecondition {
```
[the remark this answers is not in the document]
0xf8f8ff: The `qfService.CreateCourse()` will actually return the `failedPrecondition` code in three different situations:

```
if contextCanceled(ctx) {
			return nil, connect.NewError(connect.CodeFailedPrecondition, ErrContextCanceled)
		}
```

and

```
if err == ErrAlreadyExists || err == ErrFreePlan {
			return nil, connect.NewError(connect.CodeFailedPrecondition, err)
		}
```
Not sure if `CodeAlreadyExists` is a reasonable code to return in case of a free plan. Should we just separate these two cases?

web/quickfeed_service.go
0xf8f8ff: Not sure if I understand. What are alternatives?

web/quickfeed_service.go
0xf8f8ff: Yes, this is a remaining access control check I refer to in Issue #806, that should be moved to the access control interceptor. Changing it back to permission denied for now.

web/quickfeed_service.go
0xf8f8ff: Agree, `NotFound` makes perfect sense here.

web/courses_test.go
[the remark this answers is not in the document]
meling: Agree that we should separate `ErrAlreadyExists` and `ErrFreePlan`. As for the `contextCanceled`, I don't think it makes sense to use `FailedPrecondition`. See #808.

web/quickfeed_service.go
meling: Hmm. I thought this was a failed precondition error. Maybe I made a mistake.

scm/helper.go
```diff
@@ -51,6 +55,8 @@ var (
 	ErrNotMember = errors.New("user is not a member of the organization")
 	// ErrNotOwner indicates that user has no admin rights in the requested organization.
 	ErrNotOwner = errors.New("user is not an owner of the organization")
+	// ErrMissingInstallation indicates that GitHub application is not installed on organization.
+	ErrMissingInstallation = connect.NewError(connect.CodePermissionDenied, errors.New("github application is not installed on the course organization"))
```
meling, line 59: I'm not a fan of making the `scm` package (or any other packages for that matter) depend on the connect package. I would prefer to keep the connect dependencies in the `web` package and its sub-packages.

meling: I'm not sure what the frontend will see if we just return a regular error. But the message itself should be fine. Whether or not we need an error code such as permission denied, I don't know.

web/courses_test.go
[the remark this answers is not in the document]
0xf8f8ff: Changed `ErrAlreadyExists` to return the `CodeAlreadyExists`. I was thinking to limit changes to the context errors to a separate PR though.

scm/helper.go
0xf8f8ff: If an error has no code, connect will attach the default `CodeUnknown` to it. Is it ok?

0xf8f8ff: Or, because there is only one error that the `GetSCM*` methods can return, we can add the code with `connect.NewError(connect.CodeNotFound, err)` in `quickfeed_service.go` before passing it to the frontend.

meling: Yes, absolutely.

web/courses_test.go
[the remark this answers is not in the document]
meling: ok. Sounds good to me!

meling (approved):

PR #819, opened 2022-10-07
Individual submission grading

### LINKED ISSUE #566: Individual approval for group labs
I want to be able to approve a single member of a lab group, without approving the rest of the group.

This semester in DAT520 we will allow the student to work together as a group in the submission of the labs, but they have to meet individually with the TA to have their work approved.

JosteinLindhom:
This PR implements functionality to grade students individually for group assignments.

Closes #566


meling:
@JosteinLindhom: I have been reluctant to include this one, but I want to re-evaluate this after we have merged some of the other changes. To that end, I'm wondering if we could first focus on the changes needed in the database. That is, we don't necessarily need to write the code, but if you could point out in a summary note what changes are necessary.

JosteinLindhom:
> @JosteinLindhom: I have been reluctant to include this one, but I want to re-evaluate this after we have merged some of the other changes. To that end, I'm wondering if we could first focus on the changes needed in the database. That is, we don't necessarily need to write the code, but if you could point out in a summary note what changes are necessary.

We need to either:

1. Create `n` copies of a submission, where `n` is the number of students associated with a submission. For an individual lab, `n = 1`, group lab `n >= 1`. We should create a many-to-one relationship between `Submission` and `BuildInfo`, `Scores` to avoid data duplication. When a submission is built and tested, this method requires us to update `n` submissions with: `score`, `released`, `commitHash`, `approvedDate`. Not sure if there is any easy solution to avoid duplicating these fields.

2. Move the submission status field of `Submission` into its own table in the database (e.g. `Grade`, as proposed in this PR). This way we avoid duplicating data in our database. A `Submission` may have multiple grades, where each grade is related to a single user / enrollment. To update the status of an individual user (or all users), we update the Grades related to the given submission. I do remember there being some challenges with this approach, but I cannot recall exactly what those were as of this writing.

```proto
message Grade {
    uint64 SubmissionID = 1 [(go.field) = { tags: 'gorm:"primaryKey"' }];
    uint64 userID = 2 [(go.field) = { tags: 'gorm:"primaryKey"' }];
    qf.Submission.Status Status = 3;
}
```

meling (changes requested):
Turns out I had comments on this PR that never got sent.

meling:
Let's redo this PR on top of master and resolve the issue.

JosteinLindhom:
closing as this has been implemented and merged in #1104

PR #835, opened 2022-10-10
Update webhook push event handler.

### LINKED ISSUE #803: Webhooks: handlePush can be simplified and tests should be added
The current `web/hooks/handlePush()` should be simplified. 

- [ ] Ignore pushes to non-default branches should be checked before the switch-case: `if !isDefaultBranch(payload)...`
- [ ] Related to the first point; we need to tackle that group repositories with pull requests can push to non-default branches. So group repos needs special handling.
- [ ] `wh.updateLastActivityDate(repo.UserID, course.ID)` and the if necessary `jobOwner, err := wh.db.GetUserByCourse(course, payload.GetSender().GetLogin())` could be done before the switch case. 
- [ ] The `UserRepo` and `GroupRepo` cases could be simplified using a closure function.

0xf8f8ff:
Resolves #803.

Adding relevant tests in a separate PR.

### DIFF
--- a/web/hooks/push.go
+++ b/web/hooks/push.go
@@ -39,6 +39,10 @@ func (wh GitHubWebHook) handlePush(payload *github.PushEvent) {
 	}
 	wh.logger.Debugf("For course(%d)=%v", course.GetID(), course.GetName())
 
+	if repo.IsStudentRepo() {
+		wh.updateLastActivityDate(course, repo, payload.GetSender().GetLogin())
+	}
+
 	ctx := context.Background()
 	sc, err := wh.scmMgr.GetOrCreateSCM(ctx, wh.logger, course.GetOrganizationName())
 	if err != nil {
@@ -67,7 +71,6 @@ func (wh GitHubWebHook) handlePush(payload *github.PushEvent) {
 
 	case repo.IsUserRepo():
 		wh.logger.Debugf("Processing push event for user repo %s", payload.GetRepo().GetName())
-		wh.updateLastActivityDate(repo.UserID, course.ID)
 		assignments := wh.extractAssignments(payload, course)
 		for _, assignment := range assignments {
 			if !assignment.IsGroupLab {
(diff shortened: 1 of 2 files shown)


web/hooks/push.go
```diff
@@ -68,48 +69,23 @@ func (wh GitHubWebHook) handlePush(payload *github.PushEvent) {
...
 		assignments := wh.extractAssignments(payload, course)
 		for _, assignment := range assignments {
-			if !assignment.IsGroupLab {
-				// only run non-group assignments
-				wh.runAssignmentTests(sc, assignment, repo, course, payload)
-			} else {
-				wh.logger.Debugf("Ignoring push to user repo %s for group assignment: %s", payload.GetRepo().GetName(), assignment.GetName())
+			results := wh.runAssignmentTests(sc, assignment, repo, course, payload)
+			// Non-default branch indicates push to a group repo.
+			if !isDefaultBranch(payload) && !assignment.GradedManually() {
```
meling, line 78: If you move this code into `runAssignmentTests` you can remove the `GradedManually()` check (it is already done in `runAssignmentTests`) and we can remove the `results` return value from `runAssignmentTests`.

meling: The loop becomes nice and tidy:
```go
	wh.runAssignmentTests(sc, assignment, repo, course, payload)
}
```

web/hooks/push.go
```diff
@@ -68,48 +69,23 @@ func (wh GitHubWebHook) handlePush(payload *github.PushEvent) {
...
-					wh.handlePullRequestPush(payload, results, assignment, course, repo)
-				}
-			} else {
-				wh.logger.Debugf("Ignoring push to group repo %s for user assignment: %s", payload.GetRepo().GetName(), assignment.GetName())
-			}
-		}
 	default:
 		wh.logger.Debug("Nothing to do for this push event")
 	}
+
```
meling, line 88: Unnecessary blank line

web/hooks/push.go
meling: The check at the end of the `runAssignmentTests` func becomes:
```go
	// Non-default branch indicates push to a group repo.
	if !isDefaultBranch(payload) {
		// Attempt to find the pull request for the branch, if it exists,
		// and then assign reviewers to it, if the branch task score is higher than the assignment score limit
		wh.handlePullRequestPush(payload, results, assignment, course, repo)
	}
```

meling: And you can remove all the returns.

web/hooks/push.go
```diff
@@ -27,13 +27,22 @@ func (wh GitHubWebHook) handlePush(payload *github.PushEvent) {
 	}
 	wh.logger.Debugf("Received push event for repository %v", repo)
 
+	if !(isDefaultBranch(payload) || repo.IsGroupRepo()) {
```
meling, line 30: I felt this "guard" was more difficult to read than using `!isDefaultBranch && !repo.IsGroupRepo`. I think both are correct.

meling (changes requested):
This looks nice! A few minor tweaks and this should be fine, I think...

web/hooks/push.go
0xf8f8ff: Agree, this one is harder to read. Changed to the `!isDefaultBranch && !repo.IsGroupRepo`.

meling (approved):

PR #844, opened 2022-10-12
Allow services to use different types as map key

JosteinLindhom:


web/stream/service.go
```diff
@@ -12,47 +12,52 @@ import (
...
 
+// ID is the allowed type for stream IDs.
+type ID interface {
+	uint64 | string
+}
+
 // Service[T] is a type specific stream service.
 // It also contains a map of streams that are currently connected.
-type Service[T any] struct {
+type Service[T any, V ID] struct {
```
meling, line 32: I guess a common pattern for map-based structs is to have the key `K` first and then the value `V`:
```go
type Service[K ID, V any] struct {
```

web/stream/service.go
```diff
@@ -61,27 +66,27 @@ func (s *Service[T]) SendTo(data *T, userIDs ...uint64) {
 }
 
 // Add adds a new stream for the given user to the service.
```
meling, line 68: I guess it is not longer only `the given user` but also a device... (although a device is perhaps a bit weird...) Would it make sense to say `Add adds a new stream to the service for the given identifier. The identifier may be a user ID or external application ID.` (not sure if the external application ID makes sense...)

meling (changes requested):
Looks great. The only thing to change here is the ordering of K-V pairs. And check that the doc comments makes sense.

meling (approved):

PR #857, opened 2022-10-17
Update scm test helper.

### LINKED ISSUE #856: Rewrite scm.GetTestUser
Rewrite scm.GetTestUser(t) to require access token and fetching the user name from GitHub directly (using s.client.Users.Get(ctx, "")

0xf8f8ff:
Updating `scm.GetTestUser` to use access token and return a test username and test SCM client.

Resolves #856.

### DIFF
--- a/assignments/assignments_test.go
+++ b/assignments/assignments_test.go
@@ -15,7 +15,7 @@ import (
 
 func TestFetchAssignments(t *testing.T) {
 	qfTestOrg := scm.GetTestOrganization(t)
-	s := scm.GetTestSCM(t)
+	s, _ := scm.GetTestUserAndSCM(t)
 
 	course := &qf.Course{
 		Name:             "QuickFeed Test Course",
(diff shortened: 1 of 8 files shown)


assignments/assignments_test.go
```diff
@@ -15,7 +15,7 @@ import (
 
 func TestFetchAssignments(t *testing.T) {
 	qfTestOrg := scm.GetTestOrganization(t)
-	s := scm.GetTestSCM(t)
+	s, _ := scm.GetTestUserAndSCM(t)
```
meling, line 18: Suggest to keep the name `GetTestSCM`, since that's the main purpose. Alternatively, the name should at least mirror the return order: `GetTestSCMAndUser`, but I think it is not necessary...

0xf8f8ff: Renamed to `GetTestSCM`.

scm/env_helper.go
```diff
@@ -44,13 +47,3 @@ func GetWebHookServer(t *testing.T) string {
 	}
```
meling, line 47: Will we need `GetWebHookServer`?

meling: Could it be removed?

0xf8f8ff: I think it can be removed now.

meling: Ok. Let's just remove it here then.

PR #863, opened 2022-10-20
[Frontend] Switch from `grpc-web` to `connect-web`

JosteinLindhom:


Makefile
```diff
@@ -7,21 +7,16 @@
...
 toolsdir			:= bin
 tool-pkgs			:= $(shell go list -f '{{join .Imports " "}}' tools.go)
 tool-cmds			:= $(foreach tool,$(notdir ${tool-pkgs}),${toolsdir}/${tool}) $(foreach cmd,${tool-cmds},$(eval $(notdir ${cmd})Cmd := ${cmd}))
 
 # necessary when target is not tied to a specific file
-.PHONY: devtools download tools grpcweb install ui proto scm version-check
+.PHONY: devtools download tools install ui proto scm version-check
 
-devtools: grpcweb tools
+devtools: tools
```
meling, line 19: We only need `tools` then

buf.gen.ui.yaml
```diff
@@ -8,11 +8,3 @@ plugins:
   - remote: buf.build/bufbuild/plugins/connect-web:v0.2.1-1
     out: public/gen
```
meling, line 9: Should we instead keep the generated proto files in `public/proto`?

public/src/Helpers.ts
```diff
@@ -263,20 +263,19 @@ export const defaultYear = (date: Date): number => {
...
-    return `https://github.com/${course.organizationname}/${user.login}-labs`
+export const userRepoLink = (course: Course, user: User): string => {
+    return `https://github.com/${course.organizationName}/${user.login}-labs`
 }
 
-export const groupRepoLink = (course: Course.AsObject, group: Group.AsObject): string => {
-    course.organizationname
-    return `https://github.com/${course.organizationname}/${slugify(group.name)}`
+export const groupRepoLink = (course: Course, group: Group): string => {
+    return `https://github.com/${course.organizationName}/${slugify(group.name)}`
```
meling, line 275: Do we need the `slugify` here? I don't think we allow a `Group` object to be saved in the database with an illegal group name.

public/src/GRPCManager.ts
```diff
@@ -78,20 +84,20 @@ export class GrpcManager {
         return this.grpcSend<Void>(this.agService.updateCourse, course)
     }
 
-    public getCourse(courseID: number): Promise<IGrpcResponse<Course>> {
-        const request = new CourseRequest()
-        request.setCourseid(courseID)
+    public getCourse(courseID: bigint): Promise<IGrpcResponse<Course>> {
```
meling, line 87: This comment applies to all RPC forwarding functions in `GRPCManager.ts` and `MockGRPCManager.ts`. If you instead keep the arguments to these functions as `number` and convert them to `bigint` in these functions, you can continue to use them as `number` elsewhere in the code.

I guess the drawback is that the response types still contain the `bigint` variants, which would then have to be converted back to `number` if using data from the response type as input to these functions.

I guess it is something to consider.

meling (changes requested):
Mostly this looks fine to me. I think we can remove the `slugify` function. And please take a closer look at the remaining DeepSource issues. There are some that are potential bugs or otherwise a bit problematic.

meling (approved):

PR #875, opened 2022-10-24
Restructure CreateCourse

### LINKED ISSUE #810: Move sentinel errors to scm package
These errors seem to be a better fit for the scm package:
```go
var (
	repoNames = fmt.Sprintf("(%s, %s, %s)",
		qf.InfoRepo, qf.AssignmentRepo, qf.TestsRepo)

	// ErrAlreadyExists indicates that one or more QuickFeed repositories
	// already exists for the directory (or GitHub organization).
	ErrAlreadyExists = errors.New("course repositories already exist for that organization: " + repoNames)
	// ErrFreePlan indicates that payment plan for given organization does not allow private
	// repositories and must be upgraded
	ErrFreePlan = errors.New("organization does not allow creation of private repositories")
	// ErrContextCanceled indicates that method failed because of scm interaction that took longer than expected
	// and not because of some application error
	ErrContextCanceled = errors.New("context canceled because the github interaction took too long. Please try again later")
	// FreeOrgPlan indicates that organization's payment plan does not allow creation of private repositories
	FreeOrgPlan = "free"
)
```
At least some of them. Needs to be considered along with error handling. We should also replace `err == Err...` with `errors.Is()` calls.

### LINKED ISSUE #832: Allow using organizations with Free plan for QuickFeed courses.
Organizations with GitHub's Free plan can now create an unlimited number of private repositories. Because of this, we should enable using free organizations for QuickFeed courses.

### LINKED ISSUE #876: `CreateCourse` has a chain of GitHub interactions that seem easy to refactor into a separate scm method

0xf8f8ff:
The `CreateCourse` method in the `web` package included many calls to GitHub API. These steps have been refactored into a new method of the SCM interface.

Resolves #876, closes #832.
Resolves #810.


scm/github.go
```diff
@@ -661,6 +661,88 @@ func (s *GithubSCM) UpdateIssueComment(ctx context.Context, opt *IssueCommentOpt
...
+	}
+	// set default repository access level for all students to "none"
+	// will not affect organization owners (teachers)
+	orgOptions := &OrganizationOptions{
+		Name:              org.GetName(),
+		DefaultPermission: OrgNone,
+		RepoPermissions:   false,
+	}
+	if err = s.UpdateOrganization(ctx, orgOptions); err != nil {
+		return nil, fmt.Errorf("failed to update permissions for GitHub organization %s: %s", orgOptions.Name, err)
```
meling, line 685: Replace %s with %w for the error.

scm/github.go
```diff
@@ -661,6 +661,88 @@ func (s *GithubSCM) UpdateIssueComment(ctx context.Context, opt *IssueCommentOpt
...
+		repositories = append(repositories, repo)
+	}
+	// create teacher team with course creator
+	teamOpt := &NewTeamOptions{
+		Organization: org.Name,
+		TeamName:     TeachersTeam,
+		Users:        []string{opt.CourseCreator},
+	}
+	if _, err = s.CreateTeam(ctx, teamOpt); err != nil {
+		s.logger.Debugf("failed to create teachers team: %s", err)
```
meling, line 708: Should not both log and return the error.

scm/github.go
```diff
@@ -661,6 +661,88 @@ func (s *GithubSCM) UpdateIssueComment(ctx context.Context, opt *IssueCommentOpt
...
+		return nil, err
+	}
+	// create student team without any members
+	studOpt := &NewTeamOptions{Organization: org.Name, TeamName: StudentsTeam}
+	if _, err = s.CreateTeam(ctx, studOpt); err != nil {
+		s.logger.Debugf("failed to create students team: %s", err)
```
meling, line 714: Should not both log and return the error.

scm/github.go
```diff
@@ -661,6 +661,88 @@ func (s *GithubSCM) UpdateIssueComment(ctx context.Context, opt *IssueCommentOpt
...
+	if err != nil {
+		return nil, fmt.Errorf("createStudentRepo: failed to create repo: %w", err)
+	}
+
+	// add push access to student repo
+	if err = s.UpdateRepoAccess(ctx, repo, student, RepoPush); err != nil {
+		return nil, fmt.Errorf("createStudentRepo: failed to update repo push access: %w", err)
+	}
+	return repo, nil
+}
```
meling, line 744: Remove the method prefix in error message.

scm/github.go
```diff
@@ -661,6 +661,88 @@ func (s *GithubSCM) UpdateIssueComment(ctx context.Context, opt *IssueCommentOpt
...
+		Path:         path,
+		Private:      true,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("createStudentRepo: failed to create repo: %w", err)
+	}
+
+	// add push access to student repo
+	if err = s.UpdateRepoAccess(ctx, repo, student, RepoPush); err != nil {
+		return nil, fmt.Errorf("createStudentRepo: failed to update repo push access: %w", err)
```
meling, line 741: Remove the method prefix in error message.

scm/github.go
```diff
@@ -661,6 +661,88 @@ func (s *GithubSCM) UpdateIssueComment(ctx context.Context, opt *IssueCommentOpt
...
+	repo, err := s.createStudentRepo(ctx, org, qf.StudentRepoName(opt.CourseCreator), opt.CourseCreator)
+	if err != nil {
+		return nil, err
+	}
+	repositories = append(repositories, repo)
+	return repositories, nil
+}
+
+// creates {username}-labs repository and provides pull/push access to it for the given student
+func (s *GithubSCM) createStudentRepo(ctx context.Context, org *qf.Organization, path string, student string) (*Repository, error) {
```
meling, line 727: Remove string after path.

scm/github.go
```diff
@@ -661,6 +661,88 @@ func (s *GithubSCM) UpdateIssueComment(ctx context.Context, opt *IssueCommentOpt
...
+func (s *GithubSCM) CreateCourse(ctx context.Context, opt *NewCourseOptions) ([]*Repository, error) {
+	org, err := s.GetOrganization(ctx, &GetOrgOptions{ID: opt.OrganizationID})
+	if err != nil {
+		return nil, err
+	}
+	repos, err := s.GetRepositories(ctx, org)
+	if err != nil {
+		return nil, err
+	}
+	if IsDirty(repos) {
```
meling, line 674: If we need IsDirty elsewhere, it should obviously be exported, but if not it can be made unexported.

Also, we should move the call to get repositories into isDirty and only pass in the organization.

meling: And ctx

scm/github.go
```diff
@@ -661,6 +661,88 @@ func (s *GithubSCM) UpdateIssueComment(ctx context.Context, opt *IssueCommentOpt
...
+	// will not affect organization owners (teachers)
+	orgOptions := &OrganizationOptions{
+		Name:              org.GetName(),
+		DefaultPermission: OrgNone,
+		RepoPermissions:   false,
+	}
+	if err = s.UpdateOrganization(ctx, orgOptions); err != nil {
+		return nil, fmt.Errorf("failed to update permissions for GitHub organization %s: %s", orgOptions.Name, err)
+	}
+	var repositories []*Repository
```
meling, line 687: Should allocate space for len(RepoPaths)+1

scm/github.go
```diff
@@ -661,6 +661,88 @@ func (s *GithubSCM) UpdateIssueComment(ctx context.Context, opt *IssueCommentOpt
...
+	// create course repos and webhooks for each repo
```
meling, line 688: Webhooks not anymore. Rest of comment should be fixed

web/courses_new.go
```diff
@@ -9,127 +9,42 @@ import (
...
-	qf.AssignmentsRepo: private,
-	qf.TestsRepo:       private,
-}
-
 // createCourse creates a new course for the directory specified in the request
 // and creates the repositories for the course. Requires that the directory
 // does not contain the QuickFeed repositories that will be created.
 func (s *QuickFeedService) createCourse(ctx context.Context, sc scm.SCM, request *qf.Course) (*qf.Course, error) {
-	org, err := sc.GetOrganization(ctx, &scm.GetOrgOptions{ID: request.OrganizationID})
+	courseCreator, err := s.db.GetUser(request.GetCourseCreatorID())
```
meling, line 16: Should we instead get the userID of the course creator from context instead of the request?

web/courses_new.go
```diff
@@ -9,127 +9,42 @@ import (
...
 	if err != nil {
-		return nil, err
-	}
-	if org.GetPaymentPlan() == FreeOrgPlan {
-		return nil, ErrFreePlan
+		return nil, fmt.Errorf("createCourse: failed to get course creator record from database: %w", err)
```
meling, line 18: Should not have createCourse in error messages.

web/courses_new.go
```diff
@@ -9,127 +9,42 @@ import (
...
 	}
-	repos, err := sc.GetRepositories(ctx, org)
+	repos, err := sc.CreateCourse(ctx, &scm.NewCourseOptions{
+		CourseCreator:  courseCreator.Login,
+		OrganizationID: request.OrganizationID,
+	})
 	if err != nil {
+		s.logger.Debugf("createCourse: failed to create course repositories or teams: %w", err)
```
meling, line 25: Should not log error messages that also get returned up the stack. Log only once.

meling: %w does not work in log messages, only in errors

scm/github.go
0xf8f8ff: IsDirty is still used by the API `GetOrganization` method (which is being called when checking the organization before giving a user access to the front-end form to create a new course). I think this part can also be eventually moved to the `scm` package, but probably in another PR. This is why `IsDirty` has to stay exported for now.

I can see two ways to change `IsDirty` in the proposed way so that it still can be used by the API method (for now) :
1. To add it to the general scm methods ( `(scm.SCM).IsDirty(ctx, org)` )
2. Or to pass an scm client to it (`IsDirty(context.Context, scm.SCM, *qf.Organization)`)

Which solution is best or are there any better options?

scm/github.go
```diff
@@ -661,6 +661,88 @@ func (s *GithubSCM) UpdateIssueComment(ctx context.Context, opt *IssueCommentOpt
...
+		return nil, err
+	}
+	// create student team without any members
+	studOpt := &NewTeamOptions{Organization: org.Name, TeamName: StudentsTeam}
+	if _, err = s.CreateTeam(ctx, studOpt); err != nil {
+		s.logger.Debugf("failed to create students team: %s", err)
+		return nil, err
+	}
+	// add student repo for the course creator
+	repo, err := s.createStudentRepo(ctx, org, qf.StudentRepoName(opt.CourseCreator), opt.CourseCreator)
```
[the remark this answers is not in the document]
0xf8f8ff: Absolutely. Fixed.

web/courses_new.go
0xf8f8ff: We do get it from context and adding to the request before calling this method, and it still has to be added to the request as well because the request will be used as a query when creating a database record for the new course.

scm/github.go
```diff
@@ -85,6 +69,22 @@ func (s *GithubSCM) GetOrganization(ctx context.Context, opt *GetOrgOptions) (*q
 		}
 	}
 
+	org := &qf.Organization{
```
meling, line 72: We can return `org` on Line 105, avoiding to "allocate" another `qf.Organization` object.

scm/scm.go
```diff
@@ -96,6 +97,12 @@ func newSCMAppClient(ctx context.Context, logger *zap.SugaredLogger, config *Con
...
 
+// NewCourseOptions contain information about new course.
+type NewCourseOptions struct {
+	OrganizationID uint64
+	CourseCreator  string
+}
+
 // OrganizationOptions contains information on how an organization should be
 // created.
 type OrganizationOptions struct {
```
meling, line 108: Seems like this is no longer used.

scm/github.go
```diff
@@ -661,6 +658,85 @@ func (s *GithubSCM) UpdateIssueComment(ctx context.Context, opt *IssueCommentOpt
...
+	org, err := s.GetOrganization(ctx, &GetOrgOptions{ID: opt.OrganizationID, NewCourse: true})
+	if err != nil {
+		return nil, err
+	}
+
+	// Set restrictions to prevent students from creating new repositories and prevent access
+	// to organization repositories. This will not affect organization owners (teachers).
+	defaultPermissions := OrgNone
+	createRepoPermissions := false
+	if _, _, err = s.client.Organizations.Edit(ctx, org.Name, &github.Organization{
```
meling, line 673:
```go
	// DefaultRepoPermission can be one of: "read", "write", "admin", or "none". (Default: "read").
	// It is only used in OrganizationsService.Edit.
	DefaultRepoPermission *string `json:"default_repository_permission,omitempty"`
	// MembersCanCreateRepos default value is true and is only used in Organizations.Edit.
	MembersCanCreateRepos *bool `json:"members_can_create_repositories,omitempty"`
```
Seems like `DefaultRepoPermission` is not supported by `Organizations.Edit` only `OrganizationsService.Edit`. Thus, I'm not sure if this argument has any effect?? Can you check and add a comment if we still need to use the `DefaultRepoPermission` or alternatively, we maybe need to also invoke `OrganizationsService.Edit`??

meling: Never mind. This seems to be a type or weirdness on the part of the library. `OrganizationsService` is the type, while `Organizations` is a field of that type. So it is probably the same thing.

web/organizations.go
```diff
@@ -8,21 +8,5 @@ import (
 )
 
 func (s *QuickFeedService) getOrganization(ctx context.Context, sc scm.SCM, org string, user string) (*qf.Organization, error) {
```
meling, line 10: This func is now useless and should be removed.

web/courses_new.go
meling: Ok. Got it.

meling (approved):
This looks good. I added a few comments about possible removals; not sure if they have been or will be removed in the next PR. Feel free to merge.

PR #885, opened 2022-10-28
Adds python script that anonymizes the database

JosteinLindhom:


cmd/anonymize/readme.md
```diff
@@ -0,0 +1,22 @@
+# Database Tool
+
+## Installation
+
+    pip install -r requirements.txt
```
meling, line 5: Should be in a shell or bash back ticks:
```shell
% pip install -r requirements.txt
```

meling: Also no need for additional indentation

cmd/anonymize/readme.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Example usage
+
+```bash
+    # Shows available options
```
meling, line 10: No need for indentation I think. You could also add `%` symbol in front of the `python3` lines.

cmd/anonymize/main.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def close(self):
+        self.conn.close()
+
+    def anonymize(self):
+        for statement in self.get_update_list():
+            rows = self.fetch(statement[0])
+            for row in rows:
+                # Generate fake data for each column in the row (row[0] is the value of the first column, passed to the WHERE clause)
+                self.updateCur.execute(statement[1], tuple(func() for func in statement[2]) + (row[0],))
```
meling, line 50: Maybe name `row[0]` as `where_clause` or something.

And as a non-Python expert that `func() for func ...` statement was non-trivial to interpret. Perhaps split into separate lines to make the code more "readable".

cmd/anonymize/main.py
```diff
@@ -0,0 +1,118 @@
...
+        self.conn = sqlite3.connect(db)
+        self.cur = self.conn.cursor()
+        self.updateCur = self.conn.cursor()
+        # User to exclude from anonymization
+        self.excludeUser = 0
+
+    def get_update_list(self):
+        # List of tuples (select statement, update statement, functions to generate fake data)
+        return [
+            ("SELECT DISTINCT organization_id FROM repositories", "UPDATE repositories SET organization_id=? WHERE organization_id=?", (self.fake.random_number,)),
```
meling, line 30: The lines are a bit long here, making it tricky to see the difference between the select, update and fake data.

Not sure what is the best approach in Python to make it more readable.

PR #891, opened 2022-11-01
Refactor scm-related parts of UpdateGroup.

### LINKED ISSUE #890: UpdateGroup calls multiple scm methods that should be refactored into a single scm method.

0xf8f8ff:
Resolves #890.


scm/github.go
```diff
@@ -669,6 +646,42 @@ func (s *GithubSCM) DemoteTeacherToStudent(ctx context.Context, opt *UpdateEnrol
...
 
+// CreateGroup creates team and repository for a new group.
+func (s *GithubSCM) CreateGroup(ctx context.Context, opt *NewTeamOptions) (*Repository, *Team, error) {
+	if !opt.valid() {
+		return nil, nil, ErrMissingFields{
+			Method:  "CreateGroup",
+			Message: fmt.Sprintf("%+v", opt),
+		}
+	}
+	orgOtions := &GetOrgOptions{Name: opt.Organization}
```
meling, line 657: `orgOptions`

scm/github.go
```diff
@@ -669,6 +646,42 @@ func (s *GithubSCM) DemoteTeacherToStudent(ctx context.Context, opt *UpdateEnrol
...
+	org, err := s.GetOrganization(ctx, orgOtions)
+	if err != nil {
+		return nil, nil, err
+	}
+	repoOptions := &CreateRepositoryOptions{
+		Organization: opt.Organization,
+		Path:         opt.TeamName,
+		Private:      true,
+	}
+	repo, err := s.CreateRepository(ctx, repoOptions)
```
meling, line 667: Will we still need this method as part of the SCM interface? Or can we replace it with a direct call to GitHub.

scm/github.go
```diff
@@ -669,6 +646,42 @@ func (s *GithubSCM) DemoteTeacherToStudent(ctx context.Context, opt *UpdateEnrol
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	team, err := s.CreateTeam(ctx, opt)
```
meling, line 672: Will we still need this method as part of the SCM interface? Or can we replace it with a direct call to GitHub.

meling (approved):
Minor typo and some questions.

scm/github.go
0xf8f8ff: It is used in a test, where it seems easy to replace. I think it can be removed from the SCM interface.

scm/github.go
0xf8f8ff: Only used by the scm tool, so should also be easy to remove. Should I remove this and `CreateRepository` in this PR or make a separate PR?

meling: Okay to do separate PR.

PR #903, opened 2022-11-04
Remove scm methods CreateRepository and CreateTeam.

### LINKED ISSUE #902: scm methods `CreateRepository` and `CreateTeam` should be removed

0xf8f8ff:
- add an option when creating a new mock SCM to create the default course repositories and teams to provide assistance to test methods outside of the `scm` package
- remove `CreateRepository` and `CreateTeam` as they are no longer used outside of the `scm` package

Resolves #902.

### DIFF
--- a/scm/github.go
+++ b/scm/github.go
@@ -159,58 +159,6 @@ func (s *GithubSCM) RepositoryIsEmpty(ctx context.Context, opt *RepositoryOption
 	return (err != nil && resp.StatusCode == 404) || (err == nil && len(contents) == 0)
 }
 
-// CreateTeam implements the SCM interface.
-func (s *GithubSCM) CreateTeam(ctx context.Context, opt *TeamOptions) (*Team, error) {
-	if !opt.valid() || opt.TeamName == "" || opt.Organization == "" {
-		return nil, ErrMissingFields{
-			Method:  "CreateTeam",
-			Message: fmt.Sprintf("%+v", opt),
-		}
-	}
-
-	// check that the team name does not already exist for this organization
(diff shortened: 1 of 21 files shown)


scm/github.go
```diff
@@ -668,11 +615,63 @@ func (s *GithubSCM) deleteRepository(ctx context.Context, opt *RepositoryOptions
 	return nil
 }
 
+// createTeam creates a new GitHub team.
+func (s *GithubSCM) createTeam(ctx context.Context, opt *TeamOptions) (*Team, error) {
+	if !opt.valid() || opt.TeamName == "" || opt.Organization == "" {
```
meling, line 620: Should these checks be move to the valid method?

scm/mock.go
```diff
@@ -23,7 +22,9 @@ type MockSCM struct {
 }
 
 // NewMockSCMClient returns a new mock client implementing the SCM interface.
-func NewMockSCMClient() *MockSCM {
+// If withCourse is true, creates default course repositories and teams for a course
+// associated with qtest.MockOrg mock organization.
+func NewMockSCMClient(withCourse bool) *MockSCM {
```
meling, line 27: I don’t like passing bools like this. It has wide-ranging consequences (lots of unrelated changes in other code). 

It is better to add another constructor `NewMockSCMClientWithCourse`… and similar for other mocks that also require with course.

scm/mock.go
```diff
@@ -23,6 +22,8 @@ type MockSCM struct {
 }
 
 // NewMockSCMClient returns a new mock client implementing the SCM interface.
+// If withCourse is true, creates default course repositories and teams for a course
```
meling, line 25: Should be moved to other constructor func

meling (approved):
Minor doc comment needs fixing

PR #904, opened 2022-11-05
Remove RemoteIdentity type and Course.Provider field

### LINKED ISSUE #694: Remove support for multiple providers
As mentioned in #663, "*Each deployment of quickfeed should anyway only support one SCM.*"

We should simplify the codebase accordingly. There is no need to keep two or more providers. Thus, any maps or slices with providers (auth configs, remote IDs etc.) should be removed in the interest of reducing complexity.

However, we should retain the necessary generality in that it should be possible to configure quickfeed to run with either `github` or `gitlab` or possibly others... But not several providers with the same quickfeed deployment/instance/database.

meling:
- Remove database lookup to get access token for scm cmd
- Remove RemoteIdentity
- Renamed User field names to capitalization

Fixes #694


qf/remove_remote_id.go
```diff
@@ -3,13 +3,8 @@ package qf
 // RemoveRemoteID removes user's remote identity before transmitting to client.
 func (u *User) RemoveRemoteID() {
 	if u != nil {
-		voidIDs := make([]*RemoteIdentity, 0)
-		u.RemoteIdentities = voidIDs
-		for _, enrollment := range u.GetEnrollments() {
-			if enrollment.User != nil && enrollment.User.RemoteIdentities != nil {
```
meling, line 9: Ups! These enrollments may also contain the token and remote id.

JosteinLindhom (approved):

0xf8f8ff (approved):

PR #907, opened 2022-11-06
Renamed gorm:uniqueIndex tags

meling:
Just a simple renaming of the uniqueIndex gorm tags.

Based on #904.


0xf8f8ff:
Looks like this change is incompatible with the existing database. On auth, the current user is not found in the database (even though the correct record is present) and QF creates a new one which means all the enrollments and submissions for this user will be lost. Is this intended?

0xf8f8ff (approved):

Issue #864, opened 2022-10-21
Restructuring SCM package


0xf8f8ff:
Currently, most methods in the `scm` package are designed as wrappers around the corresponding methods of the `go-GitHub dependency with added validation checks for the passed arguments. The main business logic happens in the `web` package. However, quite a few of the `scm` methods are only used by one or two API methods in similar situations. It should be reasonable to move instead the business logic that relies on direct interactions with GitHub to the `scm` package to replace the "wrapper" type scm methods.

Here is a table with all methods of the `scm` package and every API method that calls them.

|        scm \ web         | **CreateCourse** | **UpdateCourse** | **GetOrganization** | **UpdateEnrollment** | **UpdateGroup** | **DeleteGroup** | **IsEmptyRepo** |
|--------------------------|-------------------|-------------------|------------|-------------------|------------------|------------------|-----------------|
| **UpdateOrganization**            |       **+**       |                   |            |                   |                  |                  |                 |
| **GetOrganization**             |       **+**       |       **+**       |    **+**        |                   |                  |                  |                 |
| **CreateRepository**           |       **+**       |                   |            |       **+**       |       **+**      |                  |                 |
| **GetRepository**              |                   |                   |            |                   |                  |                  |                 |
| **GetRepositories**             |       **+**       |                   |    **+**   |                   |                  |                  |                 |
| **DeleteRepository**           |                   |                   |            |       **+**       |                  |       **+**      |                 |
| **UpdateRepoAccess**    |       **+**       |                   |            |       **+**       |                  |                  |                 |
| **RepositoryIsEmpty**         |                   |                   |            |                   |                  |                  |      **+**      |
| **CreateTeam**          |       **+**       |                   |            |                   |       **+**      |                  |                 |
| **DeleteTeam**          |                   |                   |            |                   |                  |       **+**      |                 |
| **GetTeam**             |                   |                   |            |                   |                  |                  |                 |
| **GetTeams**            |                   |                   |            |                   |                  |                  |                 |
| **AddTeamRepo**         |                   |                   |            |                   |       **+**      |                  |                 |
| **AddTeamMember**       |                   |                   |            |       **+**       |                  |                  |                 |
| **RemoveTeamMember**    |                   |                   |            |       **+**       |                  |                  |                 |
| **UpdateTeamMember**    |                   |                   |            |                   |       **+**      |                  |                 |
| **UpdateOrgMembership** |                   |                   |            |       **+**       |                  |                  |                 |
[...]

PR #909, opened 2022-11-07
Reorganize SCM validators

### LINKED ISSUE #908: scm option types and corresponding validator methods must be kept close to each other to improve maintainability

0xf8f8ff:
SCM "options" types and corresponding `valid()` methods were defined in different places of the `scm` package. It made it easy to forget to update validators when introducing changes to the "options" interfaces. 
This PR puts validation methods close to the corresponding interface definitions. It also adds a missing validation check and updates the affected tests.

Closes #908.

### DIFF
--- a/scm/github.go
+++ b/scm/github.go
@@ -41,7 +41,7 @@ func NewGithubSCMClient(logger *zap.SugaredLogger, token string) *GithubSCM {
 }
 
 // GetOrganization implements the SCM interface.
-func (s *GithubSCM) GetOrganization(ctx context.Context, opt *GetOrgOptions) (*qf.Organization, error) {
+func (s *GithubSCM) GetOrganization(ctx context.Context, opt *OrganizationOptions) (*qf.Organization, error) {
 	if !opt.valid() {
 		return nil, ErrMissingFields{
 			Method:  "GetOrganization",
@@ -154,7 +154,7 @@ func (s *GithubSCM) GetRepositories(ctx context.Context, org *qf.Organization) (
 	if org.Name != "" {
 		path = org.Name
 	} else {
-		opt := &GetOrgOptions{
+		opt := &OrganizationOptions{
 			ID: org.ID,
 		}
 		org, err := s.GetOrganization(ctx, opt)
@@ -197,7 +197,7 @@ func (s *GithubSCM) RepositoryIsEmpty(ctx context.Context, opt *RepositoryOption
 
 // CreateTeam implements the SCM interface.
 func (s *GithubSCM) CreateTeam(ctx context.Context, opt *TeamOptions) (*Team, error) {
-	if !opt.valid() || opt.TeamName == "" || opt.Organization == "" {
+	if !opt.valid() {
 		return nil, ErrMissingFields{
 			Method:  "CreateTeam",
 			Message: fmt.Sprintf("%+v", opt),
@@ -468,7 +468,7 @@ func (s *GithubSCM) CreateCourse(ctx context.Context, opt *CourseOptions) ([]*Re
 		}
 	}
 	// Get and check the organization's suitability for the course
-	org, err := s.GetOrganization(ctx, &GetOrgOptions{ID: opt.OrganizationID, NewCourse: true})
+	org, err := s.GetOrganization(ctx, &OrganizationOptions{ID: opt.OrganizationID, NewCourse: true})
 	if err != nil {
 		return nil, err
 	}
@@ -532,7 +532,7 @@ func (s *GithubSCM) UpdateEnrollment(ctx context.Context, opt *UpdateEnrollmentO
 			Message: fmt.Sprintf("%+v", opt),
 		}
 	}
-	org, err := s.GetOrganization(ctx, &GetOrgOptions{
+	org, err := s.GetOrganization(ctx, &OrganizationOptions{
 		Name: opt.Organization,
 	})
 	if err != nil {
@@ -569,7 +569,7 @@ func (s *GithubSCM) RejectEnrollment(ctx context.Context, opt *RejectEnrollmentO
 			Message: fmt.Sprintf("%+v", opt),
 		}
 	}
-	org, err := s.GetOrganization(ctx, &GetOrgOptions{ID: opt.OrganizationID})
+	org, err := s.GetOrganization(ctx, &OrganizationOptions{ID: opt.OrganizationID})
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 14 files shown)


scm/github.go
```diff
@@ -575,6 +581,12 @@ func (s *GithubSCM) RejectEnrollment(ctx context.Context, opt *RejectEnrollmentO
 
 // DemoteTeacherToStudent removes user from teachers team, revokes owner status in the organization.
 func (s *GithubSCM) DemoteTeacherToStudent(ctx context.Context, opt *UpdateEnrollmentOptions) error {
+	if !opt.valid() {
+		return ErrMissingFields{
+			Method:  "UpdateEnrollment",
```
meling, line 586: Wrong method name. That said, I prefer to return just an `fmt.Errorf("missing fields: %+v", opt)` instead of `ErrMissingFields`... Do we use (check for) this error type somewhere? If not, maybe we can just remove it...

scm/github.go
```diff
@@ -595,7 +607,7 @@ func (s *GithubSCM) CreateGroup(ctx context.Context, opt *TeamOptions) (*Reposit
 			Message: fmt.Sprintf("%+v", opt),
```
meling, line 607: Does this method need to return both a `Repository` and `Team`? Could we get away with returning only one of them?

scm/mock_test.go
```diff
@@ -939,26 +925,82 @@ func TestMockCreateCourse(t *testing.T) {
...
-			Organization: qtest.MockOrg,
+	tests := []struct {
+		name      string
+		opt       *scm.CourseOptions
+		wantRepos int
+		wantTeams map[uint64]*scm.Team
+		wantErr   bool
+	}{
+		{
+			"invalid opt, missing org ID",
```
meling, line 937: It would be better if these table entries used the field labels like: `name`, `opt` etc.

scm/scm_options.go
```diff
@@ -0,0 +1,151 @@
...
+// GroupOptions contain information about group.
+type GroupOptions struct {
+	OrganizationID uint64
+	RepositoryID   uint64
+	TeamID         uint64
+}
+
+func (opt *GroupOptions) valid() bool {
+	return opt.OrganizationID > 0 && opt.RepositoryID > 0 &&
+		opt.TeamID > 0
```
meling, line 24: We can fit these checks on one line.

scm/scm_options.go
```diff
@@ -0,0 +1,151 @@
...
+// RejectEnrollmentOptions contain information about enrollment.
+type RejectEnrollmentOptions struct {
+	OrganizationID uint64
+	RepositoryID   uint64
+	User           string
+}
+
+func (opt *RejectEnrollmentOptions) valid() bool {
+	return opt.OrganizationID > 0 && opt.RepositoryID > 0 &&
+		opt.User != ""
```
meling, line 47: We can fit these checks on one line.

scm/scm_options.go
```diff
@@ -0,0 +1,151 @@
...
+// RequestReviewersOptions contains information on how to assign reviewers to a pull request.
+type RequestReviewersOptions struct {
+	Organization string
+	Repository   string
+	Number       int
+	Reviewers    []string // Reviewers is a slice of github usernames
+}
+
+func (opt RequestReviewersOptions) valid() bool {
+	return opt.Organization != "" && opt.Repository != "" &&
```
meling, line 149: We can fit these checks on one line.

scm/scm_options.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (opt *RejectEnrollmentOptions) valid() bool {
+	return opt.OrganizationID > 0 && opt.RepositoryID > 0 &&
+		opt.User != ""
+}
+
+// OrganizationOptions contain information about organization.
+type OrganizationOptions struct {
+	ID   uint64
```
meling, line 52: In some other structs we use `OrganizationID`. Should we be consistent?

scm/github.go
0xf8f8ff: We need to save GitHub repository ID and GitHub team ID into the database, so I'm not sure if we can.

scm/scm_options.go
```diff
@@ -0,0 +1,151 @@
...
+	Name string
```
meling, line 53: In some other structs we use `Organization` and here we use `Name`. Should we use `OrganizationName` everywhere to be consistent?

scm/scm_options.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (opt OrganizationOptions) valid() bool {
+	return opt.ID != 0 || opt.Name != ""
+}
+
+// RepositoryOptions is used to fetch a single repository by ID or name.
+// Either ID or both Path and Owner fields must be set.
+type RepositoryOptions struct {
+	ID    uint64
```
meling, line 67: Should we use `RepositoryID` to avoid confusion with other IDs?

web/courses_test.go
```diff
@@ -20,13 +20,13 @@ func TestCreateAndGetCourse(t *testing.T) {
 
 	client, tm, _ := MockClientWithUser(t, db)
 
-	admin := qtest.CreateFakeUser(t, db, 1)
+	admin := qtest.CreateAdminUser(t, db, "fake")
```
meling, line 23: Note that in another PR, I've remove `CreateAdminUser`, replacing it with `CreateFakeUser`... to reduce the number of "almost identical" user creation helpers. So this would probably cause merge problems down the line...

0xf8f8ff: I see. Will "CreateNamedUser" remain? Some tests expect that the user has a non-empty name.

scm/scm_options.go
meling: On second thought. Let's keep these as they are to avoid further polluting this PR with unrelated changes.

scm/scm_options.go
meling: On second thought. Let's keep these as they are to avoid further polluting this PR with unrelated changes.

web/courses_test.go
meling: Yes, that's the only remaining one in addition to the FakeUser one.

scm/scm_options.go
meling: On second thought. Let's keep these as they are to avoid further polluting this PR with unrelated changes.

scm/github.go
0xf8f8ff: Added a sub-issue to #864.

web/courses_test.go
0xf8f8ff: Replaced previously changed to `CreateAdminUser` with `CreateNamedUser` to avoid future merge conflicts.

meling (approved):

PR #914, opened 2022-11-08
Remove `allstudents` team

### LINKED ISSUE #897: The "allstudents" team serves no purpose and should be removed.

0xf8f8ff:
The default `allstudents` team serves no real purpose in the current system. However, it adds an extra call to GitHub API every time the status of enrollment changes. This PR removes this team and the corresponding steps in the enrollment process.

Resolves #897.

### DIFF
--- a/scm/github.go
+++ b/scm/github.go
@@ -388,12 +388,6 @@ func (s *GithubSCM) CreateCourse(ctx context.Context, opt *CourseOptions) ([]*Re
 		return nil, err
 	}
 
-	// Create student team without any members
-	studOpt := &TeamOptions{Organization: org.Name, TeamName: StudentsTeam}
-	if _, err = s.createTeam(ctx, studOpt); err != nil {
-		return nil, err
-	}
-
 	// Create student repository for the course creator
 	repo, err := s.createStudentRepo(ctx, org.Name, opt.CourseCreator)
 	if err != nil {
@@ -418,10 +412,6 @@ func (s *GithubSCM) UpdateEnrollment(ctx context.Context, opt *UpdateEnrollmentO
 	case qf.Enrollment_STUDENT:
 		// Give access to the course's info and assignments repositories
 		if err := s.grantPullAccessToCourseRepos(ctx, org.Name, opt.User); err != nil {
-			return nil, err
-		}
-		// Add student to the organization's "students" team
-		if err := s.addUserToStudentsTeam(ctx, org.Name, opt.User); err != nil {
 			return nil, err
 		}
 		return s.createStudentRepo(ctx, org.Name, opt.User)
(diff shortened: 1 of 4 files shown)


scm/github.go
```diff
@@ -662,18 +648,8 @@ func (s *GithubSCM) grantPullAccessToCourseRepos(ctx context.Context, org, login
...
 }
 
-// addUserToStudentsTeam adds user to the organization's "students" team.
-func (s *GithubSCM) addUserToStudentsTeam(ctx context.Context, org, login string) error {
-	teamMember := &github.TeamAddTeamMembershipOptions{Role: TeamMember}
-	_, _, err := s.client.Teams.AddTeamMembershipBySlug(ctx, org, StudentsTeam, login, teamMember)
-	return err
-}
-
 // promoteToTeacher adds user to the organization's "teachers" team, and removes the user from the "students" team.
```
meling, line 651: Comment needs to be updated

scm/github.go
```diff
@@ -595,14 +581,14 @@ func (s *GithubSCM) createTeam(ctx context.Context, opt *TeamOptions) (*Team, er
 			Name: opt.TeamName,
 		})
 		if err != nil {
-			if opt.TeamName != TeachersTeam && opt.TeamName != StudentsTeam {
+			if opt.TeamName != TeachersTeam {
```
meling, line 584: I’m confused by this code. Why could it happen that the teacher team already exists, if the `team` variable is nil in the above check (Line 578)?

Also, on Line 572, we use `slug.Make` twice, but continue to use both variables being slugged without slugging…

scm/github.go
```diff
@@ -595,14 +581,14 @@ func (s *GithubSCM) createTeam(ctx context.Context, opt *TeamOptions) (*Team, er
...
 				return nil, ErrFailedSCM{
 					Method:   "CreateTeam",
 					Message:  fmt.Sprintf("failed to create GitHub team %s, make sure it does not already exist", opt.TeamName),
 					GitError: fmt.Errorf("failed to create GitHub team %s: %w", opt.TeamName, err),
 				}
 			}
-			// continue if it is one of standard teacher/student teams. Such teams can be safely reused
+			// continue if it is the standard teacher team that can be safely reused
```
meling, line 591: Why is it being reused?

scm/helper.go
```diff
@@ -46,8 +46,6 @@ const (
 
 	// TeachersTeam is the team with all teachers and teaching assistants of a course.
 	TeachersTeam = "allteachers"
```
meling, line 48: Should we also rename the team to just `teachers`?

scm/github.go
0xf8f8ff: Yes, I also don't quite get it what and why is happening in this method. Should I create a new issue to rework it?

scm/github.go
0xf8f8ff: I think the original idea was as follows: we don't want the method to fail if there already is a teacher/student team in the organization and can just use the existing one. But I'm not sure this is what this method currently does.

scm/helper.go
0xf8f8ff: I think we should, but I'm not sure how it would affect demoting/promoting teachers in existing courses.

scm/github.go
0xf8f8ff: Updated slugs. We really want to only use the slugified team name here, but I think it is unnecessary for the organization name.

PR #918, opened 2022-11-10
Un-export token update check

### LINKED ISSUE #812: Unexport UpdateRequired
We currently use `UpdateCookie` only in this construction, with a precondition that an update is required. 
```go
if !u.tm.UpdateRequired(claims) {
	return claims, nil, nil
}
u.logger.Debug("Updating cookie for user ", claims.UserID)
updatedCookie, err := u.tm.UpdateCookie(claims)
```
Instead, this check can be moved inside the `UpdateCookie()` call.
```go
func (tm *TokenManager) UpdateCookie(claims *Claims) (*http.Cookie, error) {
if !tm.UpdateRequired(claims) {
	return nil, nil
}
...
```
Then `UpdateRequired` could be made unexported; it is currently used in tests, so those tests must either be rewritten to interface via the `UpdateCookie` method, or the tests must be moved to the `interceptor` package instead of `interceptor_test`.

Thoughts?

0xf8f8ff:
JWT manager's method `UpdateRequired` is now unexported.

Closes #812.

### DIFF
--- a/web/auth/jwt_test.go
+++ b/web/auth/jwt_test.go
@@ -157,17 +157,12 @@ func TestUpdateTokenList(t *testing.T) {
 	if cookie != nil {
 		t.Error("JWT update required is true, expected false")
 	}
+
 	// Adding user must update manager's update list and database record.
 	if err := manager.Add(admin.ID); err != nil {
 		t.Fatal(err)
 	}
-	cookie, err = manager.UpdateCookie(claims)
-	if err != nil {
-		t.Error(err)
-	}
-	if cookie == nil {
-		t.Error("JWT update required is false, expected true")
-	}
+	// Check database record first.
 	updatedUser, err := db.GetUser(admin.ID)
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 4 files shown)


web/auth/jwt_test.go
```diff
@@ -35,14 +35,24 @@ func TestNewManager(t *testing.T) {
 		Admin:   true,
 		Courses: make(map[uint64]qf.Enrollment_UserStatus, 0),
 	}
-	if manager.UpdateRequired(&user1claims) {
+	cookie, err := manager.UpdateCookie(&user1claims)
+	if err != nil {
+		t.Error(err)
+	}
+	if cookie != nil {
 		t.Error("JWT update required is true, expected false")
```
meling, line 43: The error doesn't make sense anymore.

web/auth/jwt_test.go
```diff
@@ -51,7 +61,11 @@ func TestNewManager(t *testing.T) {
 		UserID: user2.ID,
 		Admin:  false,
 	}
-	if !manager.UpdateRequired(&user2claims) {
+	cookie, err = manager.UpdateCookie(&user2claims)
+	if err != nil {
+		t.Error(err)
+	}
+	if cookie == nil {
 		t.Error("JWT update required is false, expected true")
```
meling, line 69: The error message doesn't make sense anymore.

web/auth/jwt_test.go
```diff
@@ -136,37 +150,58 @@ func TestUpdateTokenList(t *testing.T) {
 		Admin:  false,
 	}
 	// Admin should not be in the token update list.
-	if manager.UpdateRequired(claims) {
+	cookie, err := manager.UpdateCookie(claims)
+	if err != nil {
+		t.Error(err)
+	}
+	if cookie != nil {
 		t.Error("JWT update required is true, expected false")
```
meling, line 158: The error message doesn't make sense anymore.

web/auth/jwt_test.go
```diff
@@ -136,37 +150,58 @@ func TestUpdateTokenList(t *testing.T) {
...
 	}
-	// Removing user must update token list and user record in the database.
-	if err := manager.Remove(admin.ID); err != nil {
+	// UpdateCookie will remove user from token list and update the database record.
+	cookie, err = manager.UpdateCookie(claims)
+	if err != nil {
+		t.Error(err)
+	}
+	if cookie == nil {
+		t.Error("JWT update required is false, expected true")
```
meling, line 179: The error message doesn't make sense anymore.

web/auth/jwt_test.go
```diff
@@ -136,37 +150,58 @@ func TestUpdateTokenList(t *testing.T) {
...
 	if updatedUser.UpdateToken {
 		t.Error("User's 'UpdateToken' field not updated in the database")
 	}
+	// UpdateCookie must return nil and not an updated cookie.
+	cookie, err = manager.UpdateCookie(claims)
+	if err != nil {
+		t.Error(err)
+	}
+	if cookie != nil {
+		t.Error("JWT update required is true, expected false")
```
meling, line 203: The error message doesn't make sense anymore.

meling (approved):
Looks good to me!! Only need some error message updates in the tests.

PR #920, opened 2022-11-11
Highlight cells where the current user has pending reviews

JosteinLindhom:


meling:
Could you fix the deepsource issues. There aren't too many. Also, is this related to any issue?

PR #922, opened 2022-11-14
Add submission date to the submission's build info

### LINKED ISSUE #755: Build date or submission date
We currently only update the "build date" when a student push to their repository. However, it would be nice to distinguish between the actual build date, e.g., for rebuild purposes, and the submission date when the student last updated their code in their repo. This would allow us to rebuild if a test were updated and see that the date actually change in response. As is, the date remains the same for rebuilds.

Conclusion: we need two different date fields in the results view...

Should be done in or on top of PR #704, or after it has been merged.

0xf8f8ff:
`BuildInfo` will have two fields:
- `BuildDate` indicates the date of the latest build and will be updated after a rebuild
- `SubmissionDate` will only update on a new push to the student's repository

Both dates can be seen in the lab result view:
image

Resolves #755.


ci/run_tests_test.go
```diff
@@ -195,18 +196,22 @@ printf "RandomSecret: {{ .RandomSecret }}\n"
 		t.Fatal(err)
 	}
 	if submission.Status == qf.Submission_APPROVED {
-		t.Error("Submission must not be auto approved")
+		t.Error("submission must not be auto approved")
```
meling, line 199: It is okay and expected to use capitalization for `t.Error` and `t.Fatal` etc. 

It is the `return fmt.Errorf()` and `return errors.New()` that should be lowercase; this is because errors returned up from a function call can be nested in a final printing of the error... and then we don't want lots of capitalization.

ci/run_tests_test.go
```diff
@@ -216,29 +221,38 @@ printf "RandomSecret: {{ .RandomSecret }}\n"
 		t.Fatal(err)
 	}
 	if enrollment.RemainingSlipDays(course) == int32(course.SlipDays) || len(enrollment.UsedSlipDays) < 1 {
-		t.Error("Student must have reduced slip days")
+		t.Error("student must have reduced slip days")
 	}
-	if updatedSubmission.BuildInfo.BuildDate != newBuildDate {
-		t.Errorf("Incorrect build date: want %s, got %s", newBuildDate, updatedSubmission.BuildInfo.BuildDate)
+	if diff := cmp.Diff(newSubmissionDate, updatedSubmission.BuildInfo.BuildDate); diff != "" {
+		t.Errorf("build date mismatch: (-want +got):\n%s", diff)
```
meling, line 227: This is nice!

ci/run_tests_test.go
```diff
@@ -216,29 +221,38 @@ printf "RandomSecret: {{ .RandomSecret }}\n"
...
+	}
+	if diff := cmp.Diff(newSubmissionDate, updatedSubmission.BuildInfo.SubmissionDate); diff != "" {
+		t.Errorf("submission date mismatch: (-want +got):\n%s", diff)
 	}
 
-	// When rebuilding after deadline: delivery date and slip days must stay unchanged
+	// When rebuilding after deadline: delivery date and slip days must stay unchanged, build date must update
```
meling, line 233: must be update**d**

qf/submission_test.go
```diff
@@ -11,21 +11,21 @@ import (
 func TestNewestSubmissionDate(t *testing.T) {
 	submission := &qf.Submission{}
 	tim := time.Now()
-	newBuildDate, err := submission.NewestBuildDate(tim)
+	newBuildDate, err := submission.NewestSubmissionDate(tim)
```
meling, line 14: Should also change the `newBuildDate` on the lhs.

meling (approved):

PR #927, opened 2022-11-17
Use Select before updating database entries.

### LINKED ISSUE #570: GORM Update from struct with zero value fields may not work correctly

0xf8f8ff:
Some concerns are stated [here](https://github.com/quickfeed/quickfeed/issues/570#issuecomment-1318468312).

When using `Updates` while updating database entries, the fields that already have non-default values can never be changed back to default values (i.e., `Submission.Approved` could not be updated from `true` to `false`). To prevent this, it is better to use `Select` before updating.

Resolves #570.


database/gormdb_assignment.go
```diff
@@ -135,11 +135,7 @@ func (db *GormDB) UpdateBenchmark(query *qf.GradingBenchmark) error {
 			ID:           query.ID,
 			AssignmentID: query.AssignmentID,
 			ReviewID:     query.ReviewID,
-		}).
-		Updates(&qf.GradingBenchmark{
-			Heading: query.Heading,
-			Comment: query.Comment,
-		}).Error
+		}).Select("*").Updates(query).Error
```
meling, line 138: Hmm. Isn't it more readable/common to say select before where?
```go
Select("*").
Where(&qf.GradingBenchmark{
	ID:           query.ID,
	AssignmentID: query.AssignmentID,
	ReviewID:     query.ReviewID,
}).Updates(query).Error
```

database/gormdb_assignment.go
```diff
@@ -156,8 +152,7 @@ func (db *GormDB) CreateCriterion(query *qf.GradingCriterion) error {
 // UpdateCriterion updates the given criterion
 func (db *GormDB) UpdateCriterion(query *qf.GradingCriterion) error {
 	return db.conn.
-		Where(&qf.GradingCriterion{ID: query.ID, BenchmarkID: query.BenchmarkID}).
-		Updates(&qf.GradingCriterion{Description: query.Description, Comment: query.Comment, Grade: query.Grade, Points: query.Points}).Error
+		Where(&qf.GradingCriterion{ID: query.ID, BenchmarkID: query.BenchmarkID}).Select("*").Updates(query).Error
```
meling, line 155: Long line. Please break up the where clause similar to the `UpdateBenchmark` above.

Also, I think select should be before the where clause.

meling:
```go
	return db.conn.
		Select("*").
		Where(&qf.GradingCriterion{
			ID:          query.ID,
			BenchmarkID: query.BenchmarkID,
		}).Updates(query).Error
```

database/gormdb_enrollment.go
```diff
@@ -38,9 +38,12 @@ func (db *GormDB) RejectEnrollment(userID, courseID uint64) error {
 
 // UpdateEnrollment changes status and display state of the given enrollment.
 func (db *GormDB) UpdateEnrollment(enrol *qf.Enrollment) error {
+	// TODO(vera): Omitting ID because a lot of old tests rely on passing a query with empty ID here
+	// which would set the ID of the Enrollment record to zero.This omission is no longer required
+	// outside the tests and can be safely removed when the tests are rewritten or updated.
 	return db.conn.Model(&qf.Enrollment{}).
-		Where(&qf.Enrollment{CourseID: enrol.CourseID, UserID: enrol.UserID}).
-		Updates(&qf.Enrollment{State: enrol.State, Status: enrol.Status, LastActivityDate: enrol.LastActivityDate}).Error
+		Where(&qf.Enrollment{CourseID: enrol.CourseID, UserID: enrol.UserID}).Select("*").Omit("ID").
```
meling, line 45: This is easier to read imo.
```go
		Select("*").Omit("ID").
		Where(&qf.Enrollment{
			CourseID: enrol.CourseID,
			UserID:   enrol.UserID,
		}).Updates(enrol).Error
```

database/gormdb_group.go
```diff
@@ -74,7 +74,7 @@ func (db *GormDB) UpdateGroup(group *qf.Group) error {
 	}
 
 	tx := db.conn.Begin()
-	if err := tx.Model(group).Updates(group).Error; err != nil {
+	if err := tx.Model(group).Select("*").Updates(group).Error; err != nil {
```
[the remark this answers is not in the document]
meling: Also Line 94 below could be written more cleanly like this (I think):
```go
		Where(&qf.Enrollment{CourseID: group.CourseID}).
		Where("user_id IN (?)", userids).
		Where("status IN (?)", []qf.Enrollment_UserStatus{
			qf.Enrollment_STUDENT,
			qf.Enrollment_TEACHER,
		}).Updates(&qf.Enrollment{GroupID: group.ID})
```

database/gormdb_tasks.go
```diff
@@ -67,7 +67,7 @@ func (db *GormDB) SynchronizeAssignmentTasks(course *qf.Course, taskMap map[uint
 					existingTask.Body = task.Body
 					updatedTasks = append(updatedTasks, existingTask)
 					err = tx.Model(&qf.Task{}).
-						Where(&qf.Task{ID: existingTask.GetID()}).
+						Where(&qf.Task{ID: existingTask.GetID()}).Select("*").
```
meling, line 70: Move select before where.

public/src/overmind/actions.tsx
```diff
@@ -291,7 +291,12 @@ export const approvePendingEnrollments = async ({ state, actions, effects }: Con
         // Clone and set status to student for all pending enrollments.
         // We need to clone the enrollments to avoid modifying the state directly.
         // We do not want to update set the enrollment status before the update is successful.
-        const enrollments = Object.assign({}, state.pendingEnrollments)
+        //const enrollments = Object.assign({}, state.pendingEnrollments)
```
meling, line 294: should remove this line if not necessary anymore

public/src/overmind/actions.tsx
```diff
@@ -291,7 +291,12 @@ export const approvePendingEnrollments = async ({ state, actions, effects }: Con
...
+        const enrollments = state.pendingEnrollments.map(e => {
+            const temp = e.clone()
+            temp.status = Enrollment_UserStatus.STUDENT
```
meling, line 297: Is this necessary, given that Line 300 seems to do the same assignment again??

web/courses.go
```diff
@@ -97,25 +97,22 @@ func (s *QuickFeedService) rejectEnrollment(ctx context.Context, sc scm.SCM, enr
...
+	course, user := query.GetCourse(), query.GetUser()
 
 	// check whether user repo already exists,
 	// which could happen if accepting a previously rejected student
 	repo, err := s.getRepo(course, user.GetID(), qf.Repository_USER)
 	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
 		return fmt.Errorf("failed to get %s repository for %q: %w", course.Code, user.Login, err)
 	}
-
+	// Use enrollment with full updated info to ensure that gorm Select.Updates works correctly.
```
meling, line 110: This has me a bit confused. Does the `db.UpdateEnrollment` API require that we "fetch" from the database (or that we already have the content from the database) that we will pass in as the `query` here? 

I mean, my intuition for an update API is that, given the necessary fields to uniquely fetch an existing row to update, the other fields would be updated, but not those that we didn't pass in.

0xf8f8ff: I think I have answered this question here: https://github.com/quickfeed/quickfeed/issues/570#issuecomment-1318468312. 
Using, for example, `Select("*").Updates(Updates(&qf.Enrollment{State: enrol.State, Status: enrol.Status, LastActivityDate: enrol.LastActivityDate}))` will update all the fields we don't "pass" to their default values. 

And because in different situations we only update one of these values, it is either `State`, `Status` or `LastActivityDate` that would be set in the query, and the two others would be "empty" before and if using `Select("*")` would be updated to their default values.

web/courses.go
```diff
@@ -97,25 +97,22 @@ func (s *QuickFeedService) rejectEnrollment(ctx context.Context, sc scm.SCM, enr
...
+	query.Status = qf.Enrollment_STUDENT
 	s.logger.Debugf("Enrolling student %q in %s; has database repo: %t", user.Login, course.Code, repo != nil)
 	if repo != nil {
 		// repo already exist, update enrollment in database
-		return s.db.UpdateEnrollment(&qf.Enrollment{
-			UserID:   user.ID,
-			CourseID: course.ID,
-			Status:   qf.Enrollment_STUDENT,
```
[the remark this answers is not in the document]
0xf8f8ff: If we want to only update some fields in the database records we have to use gorm's `Select("Field1", "Field2")` and this is something we would like to avoid as these field names are easy to forget in case of changes, am I correct?

database/gormdb_group.go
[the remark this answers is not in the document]
0xf8f8ff: Yes, I am also very confused by Line 84.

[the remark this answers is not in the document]
JosteinLindhom: We set the `group_id` to zero to remove all enrollments from the given group. This way we may safely add all members of the incoming group request on Line 94.

A simplified example:
```go
// Current DB group { ... Enrollments [1, 2, 3] }
// Incoming group request { ... Enrollments [2, 3, 4] }
if err := tx.Exec("UPDATE enrollments SET group_id= ? WHERE group_id= ?", 0, group.ID).Error; err != nil {
    ...
}
// Current DB group { ... Enrollments [] }
query := tx.Model(&qf.Enrollment{}).
		Where(&qf.Enrollment{CourseID: group.CourseID}).
		Where("user_id IN (?) AND status IN (?)", userids, // [2, 3, 4]
			[]qf.Enrollment_UserStatus{qf.Enrollment_STUDENT, qf.Enrollment_TEACHER}).
		Updates(&qf.Enrollment{GroupID: group.ID})
// Current DB group { ... Enrollments [2, 3, 4] }
``` 
Without Line 84, the group would contain `Enrollments [1, 2, 3, 4]`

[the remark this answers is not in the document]
JosteinLindhom: Could update Line 84 to this to make it more GORM-like:

```go
if err := tx.Model(&qf.Enrollment{}).Where("group_id = ?", group.ID).Update("group_id", 0).Error; err != nil {
```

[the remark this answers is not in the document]
meling: Thanks for the explanation @JosteinLindhom. My suggestion is to keep the current `tx.Exec()`, but to add a comment to explain why it is there.

PR #937, opened 2022-11-23
Preload group users

JosteinLindhom:


meling:
This looks like a nice cleanup. Could you create an issue that briefly explains the cleanup being done and link to this PR. Also, would be nice to have some tests for these features.

database/gormdb_group.go
```diff
@@ -136,31 +136,27 @@ func (db *GormDB) DeleteGroup(groupID uint64) error {
...
-		userIds = append(userIds, enrollment.UserID)
-		u, err := db.GetUser(enrollment.UserID)
-		if err != nil {
-			return nil, err
-		}
-		enrollment.User = u
-	}
-	if len(userIds) == 0 {
+	if len(userIDs) == 0 {
 		return nil, errors.New("failed to get next student reviewer: no users in group")
```
meling, line 158: This error message seems a bit too specific. Should be more generic.

database/gormdb_group.go
```diff
@@ -136,31 +136,27 @@ func (db *GormDB) DeleteGroup(groupID uint64) error {
...
+	if err := db.conn.
+		Preload("Enrollments").
+		Preload("Enrollments.UsedSlipDays").
+		Preload("Enrollments.User").
+		Preload("Users", "id IN ?", userIDs).
+		First(&group, groupID).Error; err != nil {
 		if err == gorm.ErrRecordNotFound {
 			return nil, err
 		}
 		return nil, fmt.Errorf("error fetching group record for group with ID %d: %w", groupID, err)
```
meling, line 155: Error message could also be shortened

database/gormdb_group_test.go
```diff
@@ -202,12 +202,6 @@ func TestGormDBCreateAndGetGroup(t *testing.T) {
 				}
 			}
 			group.Enrollments = enrollments
-			for _, usr := range have.Users {
-				usr.Enrollments = nil
-			}
-			for _, e := range have.Enrollments {
-				e.User.Enrollments = nil
-			}
```
meling, line 210: Let me know if tests are sufficient already.

meling (changes requested):
Some changes while we are touching this is to fix the error messages.

database/gormdb_group_test.go
JosteinLindhom: I believe `TestGormDBCreateAndGetGroup` and `TestGetGroupsByCourse` are sufficient already.

meling (approved):

PR #939, opened 2022-11-24
Combine two methods and corresponding request types used to fetch enrollments.

### LINKED ISSUE #933: Combine EnrollmentRequest and EnrollmentStatusRequest
We have these two RPCs and corresponding request message types:
```proto
    rpc GetEnrollmentsByUser(EnrollmentStatusRequest) returns (Enrollments) {}
    rpc GetEnrollmentsByCourse(EnrollmentRequest) returns (Enrollments) {}
```
```proto
message EnrollmentRequest {
    uint64 courseID                         = 1;
    bool ignoreGroupMembers                 = 2;
    bool withActivity                       = 3;
    repeated Enrollment.UserStatus statuses = 4;
}

// EnrollmentStatusRequest is a request for a given user, with a specific enrollment status.
message EnrollmentStatusRequest {
    uint64 userID                           = 1;
    repeated Enrollment.UserStatus statuses = 2;
}
```
We can combine these into a single RPC and single `EnrollmentRequest` message type.
```proto
message EnrollmentRequest {
    uint64 courseID                         = 1;
    uint64 userID                           = 2;
    repeated Enrollment.UserStatus statuses = 3;
}
```
Or optionally:
```proto
message EnrollmentRequest {
    oneof FetchMode {
        uint64 courseID = 1;
        uint64 userID   = 2;
    }
    repeated Enrollment.UserStatus statuses = 3;
}
```
We don't need the `ignoreGroupMembers` (always false) and `withActivity` (always true) flags anymore. Thus, it will suffice to have a single RPC for this:
```proto
    rpc GetEnrollments(EnrollmentRequest) returns (Enrollments) {}
```

0xf8f8ff:
- `GetEnrollmentsByUser` and `GetEnrollmentByCourse` were combined into a single API method `GetEnrollments`
- 'EnrollmentRequest` and `EnrollmentStatusRequest` combined into a single `EnrollmentRequest` type
- `withActivity` Boolean field of the `EnrollmentRequest` was removed because it was always set to True in all use cases
- `ignoreGroupMembers` Boolean field was removed as it was always set to False
- access control roles of the new `GetEnrollments` method changed to {user student teacher} as the "admin" case of the old `GetEnrollmentByUser` was not required.

Resolves #933.


meling:
Looks good so far. Please fix the JavaScript issue found by deepsource. (Only two issues).

The only thing that I couldn't determine just by looking at the PR diffs is whether or not the GetEnrollments() will limit user and student role to their own enrollments, and that it wouldn't be possible to get other user's enrollments. I think it is correct, but just mentioning that I didn't check.

0xf8f8ff:
> The only thing that I couldn't determine just by looking at the PR diffs is whether or not the GetEnrollments() will limit user and student role to their own enrollments, and that it wouldn't be possible to get other user's enrollments. I think it is correct, but just mentioning that I didn't check.

Because of the `oneof` field in the `EnrollmentRequest` access control checks work as follows.
The `user` role only grants access if the `UserID` in the request is the same as in the JWT claims. If the request is of the `_CourseID` type, `GetUserID()` returns 0, and access is never granted.

The `student` and `teacher` roles only grant access if the current user enrolled in the `request.CourseID` with relevant enrollment status. If the request is of the `_userID` type, `GetCourseID` will return 0, and, again, the access would never be granted.

web/interceptor/access_control_test.go
```diff
@@ -149,6 +151,7 @@ func TestAccessControl(t *testing.T) {
 	}
 	for name, tt := range studentAccessTests {
 		t.Run("StudentAccess/"+name, func(t *testing.T) {
+			t.Log("NAME: ", name)
```
meling, line 154: Should be removed, no?

meling (approved):
Minor test log output should be removed

web/interceptor/access_control_test.go
0xf8f8ff: Absolutely. Fixed.

PR #942, opened 2022-11-30
Update GetRepositories request type

### LINKED ISSUE #934: Replace URLRequest with CourseRequest
```proto
message URLRequest {
    uint64 courseID                    = 1;
    repeated Repository.Type repoTypes = 2;
}
```
We currently have this:
```proto
    rpc GetRepositories(URLRequest) returns (Repositories) {}
```
We should instead do:
```proto
    rpc GetRepositories(CourseRequest) returns (Repositories) {}
```
Instead of letting the frontend decide which repo types to include in the response, we should let the server decide based on the user's status (student or teacher) as recorded in the database. Hence, for `GetRepositories` we can reuse the `CourseRequest` message type.

### LINKED ISSUE #941: Repository links in the student view appear to be broken
It looks like the student view repository links to the `Course info`, 'Assignments', and `-labs` repositories aren't working anymore.

0xf8f8ff:
- Use `CourseRequest` instead of `URLRequest` in `GetRepositories`
- Prevent `GetRepositories` from returning an arbitrary group repository for enrollments without a group
- Update the corresponding test

Closes #934. Fixes #941.


public/src/MockGRPCManager.ts
```diff
@@ -656,14 +655,11 @@ export class MockGrpcManager {
 
     // /* REPOSITORY */ //
 
-    public getRepositories(courseID: bigint, types: Repository_Type[]): Promise<IGrpcResponse<Repositories>> {
+    public getRepositories(courseID: bigint): Promise<IGrpcResponse<Repositories>> {
         // TODO
```
meling, line 659: Remove todo?

public/src/MockGRPCManager.ts
```diff
@@ -656,14 +655,11 @@ export class MockGrpcManager {
...
         if (!this.courses.courses.find(c => c.ID === courseID)) {
             return this.grpcSend<Repositories>(null, new Status({ Code: BigInt(Code.Unknown), Error: "Course not found" }))
         }
-        types.forEach(() => {
-            // TODO
-        })
         //const repos = this.repositories.getRepositoriesList().filter(r => r.courseID === courseID && types.includes(r.getType()))
```
meling, line 663: Perhaps this should be replaced with code that returns some dummy repositories for mocking instead of the empty repositories.

web/quickfeed_service.go
```diff
@@ -563,17 +563,27 @@ func (s *QuickFeedService) GetOrganization(ctx context.Context, in *connect.Requ
...
 	course, err := s.db.GetCourse(in.Msg.GetCourseID(), false)
 	if err != nil {
 		s.logger.Errorf("GetRepositories failed: course %d not found: %v", in.Msg.GetCourseID(), err)
 		return nil, connect.NewError(connect.CodeNotFound, errors.New("course not found"))
 	}
 	usrID := userID(ctx)
-	enrol, _ := s.db.GetEnrollmentByCourseAndUser(course.GetID(), usrID)
+	enrol, err := s.db.GetEnrollmentByCourseAndUser(course.GetID(), usrID)
+	if err != nil {
+		s.logger.Error("GetRepositories failed: enrollment for user %d and course %d not found", usrID, course.GetID())
```
meling, line 575: Should also print the root cause error after `...not found: %v"`

web/repositories.go
```diff
@@ -50,3 +50,11 @@ func (s *QuickFeedService) isEmptyRepo(ctx context.Context, sc scm.SCM, request
 	}
 	return isEmpty(ctx, sc, repos)
 }
+
+func repoTypes(enrollment *qf.Enrollment) []qf.Repository_Type {
+	repositories := []qf.Repository_Type{qf.Repository_INFO, qf.Repository_ASSIGNMENTS, qf.Repository_USER, qf.Repository_GROUP}
```
meling, line 55: Better to add line breaks here; one repo type per line.

web/quickfeed_service.go
```diff
@@ -563,17 +563,27 @@ func (s *QuickFeedService) GetOrganization(ctx context.Context, in *connect.Requ
...
+		return nil, connect.NewError(connect.CodeNotFound, errors.New("enrollment not found"))
+	}
 
 	urls := make(map[string]string)
-	for _, repoType := range in.Msg.GetRepoTypes() {
+	for _, repoType := range repoTypes(enrol) {
+		if repoType == qf.Repository_GROUP && enrol.GroupID == 0 {
```
meling, line 581: This check should be done in the helper function `repoTypes()`. We should only return a `qf.Repository_GROUP` if the enrollment has a `GroupID > 0`.

web/repositories.go
```diff
@@ -50,3 +50,11 @@ func (s *QuickFeedService) isEmptyRepo(ctx context.Context, sc scm.SCM, request
...
+	if enrollment.IsTeacher() {
+		repositories = append(repositories, qf.Repository_TESTS)
+	}
```
meling, line 58: Add the following:
```go
if enrollment.GetGroupID() > 0 {
	repositories = append(repositories, qf.Repository_GROUP)
}
```
Given this code, we don't need the length comment explaining that `getRepo()` returns any (the first) group repository for the course.

web/quickfeed_service.go
meling: See code suggestion below.

web/quickfeed_service.go
```diff
@@ -563,17 +563,27 @@ func (s *QuickFeedService) GetOrganization(ctx context.Context, in *connect.Requ
...
+			// If groupID for a group repository type is 0, getRepo() returns the first group repository for the course
+			// because database query looks like this: Repository{type: GROUP, OrganizationID: course.OrgID} which mathes
+			// to any group repository for this course. To prevent this, ignore such cases instead of calling getRepo().
+			continue
+		}
 		var id uint64
 		switch repoType {
 		case qf.Repository_USER:
```
meling, line 589: Referring to Line 596, if you implement my suggestion from [issue #931 (comment)](https://github.com/quickfeed/quickfeed/issues/941#issuecomment-1332652355), you can use:
```go
urls[repoType] = repo.GetHTMLURL()
```
(I think)...

web/repositories.go
meling: Remove `qf.Repository_GROUP` from the initial list.

qf/requests.proto
```diff
@@ -91,7 +86,7 @@ message RepositoryRequest {
 }
 
 message Repositories {
-    map<string, string> URLs = 1;
+    map<uint32, string> URLs = 1;
```
meling, line 89: Should add a comment `// Map key is the Repository.Type`

web/repositories.go
```diff
@@ -50,3 +50,17 @@ func (s *QuickFeedService) isEmptyRepo(ctx context.Context, sc scm.SCM, request
 	}
 	return isEmpty(ctx, sc, repos)
 }
+
+func repoTypes(enrollment *qf.Enrollment) []qf.Repository_Type {
+	repositories := []qf.Repository_Type{
+		qf.Repository_INFO,
+		qf.Repository_ASSIGNMENTS,
+		qf.Repository_USER}
```
meling, line 58: Prefer to keep the final `}` on a separate line.

meling (approved):
LGTM! Only a few minor remaining things.

public/src/MockGRPCManager.ts
meling: @JosteinLindhom: could you take a look at this? Not sure if the mock method is used or needed in the near future, but if not, perhaps we can just remove the line.

meling:
@JosteinLindhom: There are some aspects to this PR that could benefit from your insight before we merge it... Thanks!

JosteinLindhom (approved):

PR #943, opened 2022-12-01
Remove link types

JosteinLindhom:


public/src/Helpers.ts
```diff
@@ -277,11 +271,17 @@ export const userLink = (user: User): string => {
     return `https://github.com/${user.Login}`
 }
 
-export const userRepoLink = (course: Course, user: User): string => {
+export const userRepoLink = (user: User, course?: Course): string => {
```
meling, line 274: These helpers (userLink, UserRepoLink, and groupRepoLink) should be removed or changed to use information obtained from the updated GetRepositories() RPC method. (Maybe we can fix it later, once #942 has been merged.)

public/src/Helpers.ts
```diff
@@ -135,12 +135,12 @@ export const getPassedTestsCount = (score: Score[]): string => {
...
 
-export const isValid = (elm: User | EnrollmentLink): boolean => {
+export const isValid = (elm: User | Enrollment): boolean => {
     if (elm instanceof User) {
         return elm.Name.length > 0 && elm.Email.length > 0 && elm.StudentID.length > 0
     }
-    if (elm instanceof EnrollmentLink) {
-        return elm.enrollment?.user !== undefined && elm.submissions.length > 0
+    if (elm instanceof Enrollment) {
+        return elm.user !== undefined && isValid(elm.user)
```
meling, line 143: I recall that deep source didn't like comparing with `undefined`, but didn't check. Maybe it should be changed to `null` instead?

public/src/Helpers.ts
```diff
@@ -203,33 +203,27 @@ export const isHidden = (value: string, query: string): boolean => {
...
-            return
-        }
-        score += link.submission.score
+    submissions.forEach(submission => {
+        score += submission.score
     })
     return score
 }
 
 /** getNumApproved returns the number of approved submissions in a SubmissionLink[] */
```
meling, line 214: Update comment: Remove `in a SubmissionLink[]`

qf/requests.proto
```diff
@@ -8,8 +8,7 @@ import "qf/types.proto";
 // Message types that are used as requests and responses in API calls.
 
 message CourseSubmissions {
-    Course course                 = 1;  // preloaded assignments
-    repeated EnrollmentLink links = 2;
+    map<uint64, Submissions> submissions = 1;
```
meling, line 11: I assume the map key is a course ID; We should add a comment `// map: CourseID -> Submissions`

web/courses.go
```diff
@@ -184,88 +184,83 @@ func (s *QuickFeedService) getAllCourseSubmissions(request *qf.SubmissionRequest
...
 // for all course groups and all group assignments
-func makeGroupResults(course *qf.Course, assignments []*qf.Assignment) []*qf.EnrollmentLink {
-	enrolLinks := make([]*qf.EnrollmentLink, 0)
+func makeGroupResults(course *qf.Course, submissions []*qf.Submission, assignments []*qf.Assignment) map[uint64]*qf.Submissions {
+	submissionsMap := make(map[uint64]*qf.Submissions)
 	seenGroup := make(map[uint64]bool)
+	om := newOrderMap(course.GetAssignments())
 	for _, enrollment := range course.Enrollments {
-		if seenGroup[enrollment.GroupID] {
+		if seenGroup[enrollment.GroupID] || enrollment.GroupID == 0 {
```
meling, line 206: This could be made simpler with:
```go
skipGroup := map[uint64]bool{
	0: true,  // ignore enrollments without a group
}
```
And then do:
```go
if skipGroup[enrollment.GroupID] {
	continue // include group enrollment only once
}
skipGroup[enrollment.GroupID] = true
```

qf/requests.proto
JosteinLindhom: The map key is actually EnrollmentID or GroupID based on which SubmissionType is sent in the request. ALL and USER are keyed by EnrollmentID, GROUP is keyed by GroupID.

web/courses.go
```diff
@@ -184,88 +184,83 @@ func (s *QuickFeedService) getAllCourseSubmissions(request *qf.SubmissionRequest
 		return nil, err
 	}
 
-	var enrolLinks []*qf.EnrollmentLink
+	var submissionsMap map[uint64]*qf.Submissions
 	switch request.GetType() {
 	case qf.SubmissionRequest_GROUP:
-		enrolLinks = makeGroupResults(course, assignments)
+		submissionsMap = makeGroupResults(course, submissions, course.GetAssignments())
```
meling, line 190: The `course.GetAssignments()` argument is unused in these `make` functions.

web/courses.go
```diff
@@ -385,3 +385,17 @@ func (s *QuickFeedService) acceptRepositoryInvites(ctx context.Context, scmApp s
 	user.RefreshToken = newRefreshToken
 	return s.db.UpdateUser(user)
 }
+
+type orderMap map[uint64]uint32
+
+func newOrderMap(assignments []*qf.Assignment) *orderMap {
```
meling, line 391: This looks neat. Could you add a comment to explain what it does and also add a test with a few use-case examples?

web/courses.go
```diff
@@ -333,30 +328,35 @@ func (s *QuickFeedService) getEnrollmentsWithActivity(courseID uint64) ([]*qf.En
...
-			submission := submissionLink.Submission
-			if submission != nil {
-				if submission.Status == qf.Submission_APPROVED {
-					totalApproved++
-				}
-				if enrol.LastActivityDate == "" {
-					submissionDate, err = submission.NewestSubmissionDate(submissionDate)
-					if err != nil {
-						return nil, err
+		if submissions, ok := submissionsMap.Submissions[enrollment.ID]; ok {
```
meling, line 340: You can add this snippet in the `qf` package:
```go
// For returns the course's submissions for the given ID.
func (cs *CourseSubmissions) For(id uint64) []*Submission {
	submissions := cs.GetSubmissions()
	if submissions == nil {
		return nil
	}
	return submissions[id].GetSubmissions()
}
```
And simplify this code like this:
```go
for _, submission := range submissionsMap.For(enrollment.ID) {
	if submission.Status == qf.Submission_APPROVED {
		totalApproved++
	}
...
```
Maybe `submissions` would then be a better name instead of `submissionsMap`

web/quickfeed_service.go
```diff
@@ -364,8 +364,11 @@ func (s *QuickFeedService) GetSubmissions(ctx context.Context, in *connect.Reque
 	return connect.NewResponse(submissions), nil
 }
 
-// GetSubmissionsByCourse returns all the latest submissions
-// for every individual or group course assignment for all course students/groups.
+// GetSubmissionsByCourse returns a map of submissions for the given course ID.
+// The map is keyed by either the group ID or enrollment ID depending on request type.
+// SubmissionRequest__GROUP returns a map keyed by group ID.
```
meling, line 369: `SubmissionRequest__GROUP` -> `SubmissionRequest_GROUP`

web/quickfeed_service.go
```diff
@@ -364,8 +364,11 @@ func (s *QuickFeedService) GetSubmissions(ctx context.Context, in *connect.Reque
 	return connect.NewResponse(submissions), nil
 }
 
-// GetSubmissionsByCourse returns all the latest submissions
-// for every individual or group course assignment for all course students/groups.
+// GetSubmissionsByCourse returns a map of submissions for the given course ID.
+// The map is keyed by either the group ID or enrollment ID depending on request type.
```
meling, line 368: Comment on the general behavior. I wonder if we should instead add a separate `FetchMode` for `EnrollmentID`??

qf/requests.proto
meling: Hmm. I wonder if we should instead make it `EnrollmentID` always. I mean, the `Enrollment` pointed to by an `EnrollmentID` should have a `GroupID` if the enrollment is for a group, or am I just too tired?

JosteinLindhom: My thinking was that if we provide `SubmissionType_GROUP`, the returned map *only* contains group submissions. This way I can:
- Range over all groups in a course to easily access the group's submissions: `groups.forEach(group => submissionsByGroup[group.ID])`
- Easily access group submissions for a specific enrollment (lookup by `Enrollment.GroupID`)

This way, if we have three enrollments belonging to the same group, they will all share one entry in our state `submissionsByGroup` map of submissions. 

```ts
// Assignments: 1 2 3 4 5(g) 6(g) 7
// Enrollment IDs: 1, 2, 3, 
// Group IDs: 66 (all enrollments above are in this group)
// Numbers in the array represent which assignment the submission belongs to
Type -> SubmissionType_USER
Submissions -> {1: [1, 2, 3, 4, 7], 2: [1, 2, 3, 4, 7], 3: [1, 2, 3, 4, 7]}

// For this type, the key is GroupID, the two others are keyed by EnrollmentID
Type -> SubmissionType_GROUP
Submissions -> {66: [5, 6]}

// Note submissions for assignments 5 and 6, which are identical for all three enrollments, are duplicated
Type -> SubmissionType_ALL
Submissions -> {1: [1, 2, 3, 4, 5, 6, 7], 2: [1, 2, 3, 4, 5, 6, 7], 3: [1, 2, 3, 4, 5, 6, 7]}
```

public/src/Helpers.ts
JosteinLindhom: Apparently this helper is only used in tests in `isValid.test.tsx`. I removed it altogether and updated the test to use `state.isValid` that is derived from `state.self`

public/src/MockGRPCManager.ts
```diff
@@ -457,57 +455,51 @@ export class MockGrpcManager {
 
     public getSubmissionsByCourse(courseID: bigint, type: SubmissionRequest_SubmissionType): Promise<IGrpcResponse<CourseSubmissions>> {
         // TODO: Remove `.clone()` when done migrating to AsObject in state
```
meling, line 457: Can this be removed? There are still some `.clone()` calls, but not sure if those are related to this. Moreover, there are no more `AsObject`.

public/src/MockGRPCManager.ts
```diff
@@ -457,57 +455,51 @@ export class MockGrpcManager {
...
 
-                subLink.submission = submission.clone()
-                subs.push(subLink)
+                submissions.submissions[id.ID.toString()].submissions.push(submission.clone())
             })
-            link.submissions = subs
-            enrollmentLinks.push(link)
         })
-        submissions.links = enrollmentLinks
         // TODO
```
meling, line 503: Remove

database/gormdb_assignment.go
```diff
@@ -180,31 +180,31 @@ func (db *GormDB) updateGradingCriteria(tx *gorm.DB, assignment *qf.Assignment)
...
-	var assignments []*qf.Assignment
-	m := db.conn.Preload("Submissions").
-		Preload("Submissions.Reviews").
-		Preload("Submissions.Reviews.GradingBenchmarks").
-		Preload("Submissions.Reviews.GradingBenchmarks.Criteria").
-		Preload("Submissions.Scores")
+// GetCourseSubmissions returns the latest course submissions of the requested submission type.
+func (db *GormDB) GetCourseSubmissions(courseID uint64, submissionType qf.SubmissionRequest_SubmissionType) ([]*qf.Submission, error) {
+	var assignmentIDs []uint64
+	a := db.conn.Model(&qf.Assignment{}).Where("course_id = ?", courseID)
```
meling, line 186: Would it be better to it like this (to be type safe)?:
```go
a := db.conn.Model(&qf.Assignment{}).Where(&qf.Assignment{CourseID: courseID})
switch submissionType {
case qf.SubmissionRequest_USER:
	a.Where(&qf.Assignment{IsGroupLab: false})
case qf.SubmissionRequest_GROUP:
	a.Where(&qf.Assignment{IsGroupLab: true})
default: // all
}
```
That is, is there a good reason to use the string-based versions?

public/src/Helpers.ts
```diff
@@ -372,3 +312,30 @@ export const sortEnrollments = (enrollments: Enrollment[], sortBy: EnrollmentSor
         return enrollmentCompare(a, b, sortBy, descending)
     })
 }
+
+export class SubmissionsForCourse {
```
meling, line 316: Should this class be colocated with the `SubmissionOwner` type? Three alternatives:

- Separate new file `SubmissionsForCourse.ts`
- In this file (Helpers.ts)
- In state.tsx (where `SubmissionOwner` is defined)

public/src/components/manual-grading/ReviewForm.tsx
```diff
@@ -10,14 +10,13 @@ import ReviewResult from "../ReviewResult"
 const ReviewForm = (): JSX.Element => {
     const state = useAppState()
     const actions = useActions()
-    const courseID = getCourseID()
 
-    if (!state.activeSubmissionLink) {
+    if (!state.currentSubmission) {
```
meling, line 14: Maybe the following could be made into a method on `state`:
```ts
    if (!state.currentSubmission) {
        return <div>No submission selected</div>
    }
    if (!state.selectedAssignment) {
        return <div>No assignment Selected</div>
    }
```
Or similar. It seems this could be useful elsewhere as well. Not sure if we could return one or both of `assignment` and `submission`

public/src/overmind/namespaces/review/actions.ts
```diff
@@ -23,7 +23,11 @@ export const updateReview = async ({ state, actions, effects }: Context): Promis
         const response = await effects.grpcMan.updateReview(review, state.activeCourse)
         if (success(response) && response.data) {
             // Updates the currently selected review with the new data from the server
-            state.review.reviews[state.activeCourse.toString()][state.activeSubmission][state.review.selectedReview] = response.data
+            const reviews = new Map(state.review.reviews)
+            reviews.set(state.activeSubmission, state.review.reviews.get(state.activeSubmission)?.map(r => r.ID === review.ID ? response.data as Review : r) ?? [])
```
meling, line 27: Too long line. Use more lines or make a method if this would be of use in other places.

public/src/overmind/namespaces/review/state.ts
```diff
@@ -7,11 +7,7 @@ export type ReviewState = {
     selectedReview: number
 
     /* Contains all reviews for the different courses, indexed by the course id and submission id */
-    reviews: {
-        [courseID: string]: {
-            [submissionID: number]: Review[]
-        }
-    }
+    reviews: Map<bigint, Review[]>
```
meling, line 10: Nice! So there is actually a `Map` type that can use `bigint` as key.

public/src/Helpers.ts
```diff
@@ -372,3 +312,30 @@ export const sortEnrollments = (enrollments: Enrollment[], sortBy: EnrollmentSor
...
+    userSubmissions: { [key: string]: Submissions } = {}
```
meling, line 317: These could use the `Map` type with the `bigint` IDs as key instead of doing the conversion to string, or did I miss something?

qf/requests.proto
meling: Okay. I guess we should provide a brief documentation of this above the `CourseSubmissions` message type.

public/src/Helpers.ts
JosteinLindhom: Ideally I would like to use `Map` for this. I'm having issues with Overmind not tracking mutations in the map when using `Map.set(x, y)`. I can get it to work if I clone the entire map, mutate the cloned map, then reassign the map in state with the mutated one. This seems to me to be terribly inefficient though.

public/src/Helpers.ts
JosteinLindhom: Currently `GetRepositories` returns the repositories for a single user. We only use the returned repositories in the `CourseUtilityLinks` component.

For enrollments fetched for an entire course we currently need these helper methods to build the links. We *could* add `Repositories` to both the `Group` and `Enrollment` types, and preload the repositories. If we do this we could avoid using these helpers.

meling: Sounds reasonable. We should do it in a separate PR. Please create an issue to do it later.

public/src/Helpers.ts
```diff
@@ -372,3 +313,62 @@ export const sortEnrollments = (enrollments: Enrollment[], sortBy: EnrollmentSor
         return enrollmentCompare(a, b, sortBy, descending)
     })
 }
+
+export class SubmissionsForCourse {
+    userSubmissions: Map<bigint, Submissions> = new Map()
+    groupSubmissions: Map<bigint, Submissions> = new Map()
+
+    /** getSubmissionsForEnrollment returns user submissions for the given enrollment */
+    getSubmissionsForEnrollment(enrollment: Enrollment): Submission[] {
```
meling, line 322: Suggested rename for these three `getSubmissions` funcs:
```ts
ForUser
ForGroup
ForOwner
```

public/src/overmind/actions.tsx
```diff
@@ -913,3 +795,20 @@ export const updateGroupName = ({ state }: Context, name: string): void => {
 export const setConnectionStatus = ({ state }: Context, status: ConnStatus) => {
     state.connectionStatus = status
 }
+
+// setSubmissionOwner sets the owner of the currently selected submission.
+// The owner is either an enrollment or a group.
+export const setSubmissionOwner = ({ state }: Context, { submission, enrollment }: { submission: Submission, enrollment: Enrollment | Group }) => {
```
meling, line 801: Not sure I like the Owner term. What about `setSubmissionType`?

public/src/overmind/actions.tsx
```diff
@@ -913,3 +795,20 @@ export const updateGroupName = ({ state }: Context, name: string): void => {
...
+        // If the submission has a groupID, it is a group submission,
+        // regardless of whether the enrollment object is of group or enrollment type.
+        state.submissionOwner = { type: "GROUP", id: submission.groupID }
+    } else if (enrollment instanceof Group) {
+        // If the enrollment object is a group, it is a group submission.
+        // However, the submission does not have a groupID.
+        // This should not happen, but we handle it anyway.
+        return
+    } else {
+        state.submissionOwner = { type: "ENROLLMENT", id: enrollment.ID }
```
meling, line 812: Doesn't look like this submission type is "checked against". That's okay.

My actual question is: Should this type be "USER" instead of "ENROLLMENT" since the group type is handled separately? Or can this also be a group enrollment?

public/src/overmind/actions.tsx
```diff
@@ -913,3 +795,20 @@ export const updateGroupName = ({ state }: Context, name: string): void => {
...
+// The owner is either an enrollment or a group.
+export const setSubmissionOwner = ({ state }: Context, { submission, enrollment }: { submission: Submission, enrollment: Enrollment | Group }) => {
+    if (submission.groupID > 0) {
+        // If the submission has a groupID, it is a group submission,
+        // regardless of whether the enrollment object is of group or enrollment type.
+        state.submissionOwner = { type: "GROUP", id: submission.groupID }
+    } else if (enrollment instanceof Group) {
+        // If the enrollment object is a group, it is a group submission.
+        // However, the submission does not have a groupID.
+        // This should not happen, but we handle it anyway.
```
meling, line 809: We don't actually handle it since we just return. Maybe just say ignore it.

public/src/overmind/actions.tsx
JosteinLindhom: We can rename the type `"ENROLLMENT"` to `"USER"`. The `type` determines which map to look for submissions in. If the `enrollment` is of the type `Enrollment`, we look for the `id` in `userSubmissions: Map<bigint, Submissions>`. If we're dealing with a `Group`, `groupSubmissions: Map<bigint, Submissions>` is used instead.

meling:
I've studied the SubmissionOwner thing, and it adds unnecessary complexity, IMO. Will it help to solve our problems if we add the following:
```proto
    uint64 EnrollmentID                    = 3;
```
To the `Submission` message type?

That way, we can do things like:
```ts
        if (isGroup(submission)) {
            const clone = new Map(this.groupSubmissions)
            this.groupSubmissions = clone.set(submission.groupID, new Submissions({ submissions }))
        } else {
            const clone = new Map(this.userSubmissions)
            this.userSubmissions = clone.set(submission.EnrollmentID, new Submissions({ submissions }))
        }
```

Moreover, would it help to add a field to the following:
```proto
message CourseSubmissions {
    SubmissionRequest.SubmissionType Type = 1;
    map<uint64, Submissions> submissions  = 1;
}
```

public/src/overmind/state.tsx
```diff
@@ -297,46 +310,46 @@ export const state: State = {
...
-        }
-        return -1
+        return sortedSubmissions as Group[] | Enrollment[]
     }),
+    activeSubmission: 0n,
     activeEnrollment: null,
-    activeSubmissionLink: null,
-    currentSubmission: derived(({ activeSubmissionLink }: State) => {
-        return activeSubmissionLink?.submission ?? null
+    currentSubmission: derived(({ activeSubmission, submissionsForCourse, submissionOwner }: State) => {
```
meling, line 340: It may be a bit inefficient, but couldn't we replace `submissionOwner` and the `ForOwner` method here with a new method on `submissionsForCourse` that examines all submissions in both user and group maps for a specific submission ID. For example, `submissionsForCourse.Submission(activeSubmission)`?

PR #944, opened 2022-12-03
Removes unused courseID argument in UpdateSubmissions database method

### LINKED ISSUE #924: UpdateSubmissions has unused courseID argument
```go
func (db *GormDB) UpdateSubmissions(courseID uint64, query *qf.Submission) error {
```
The method does not use the `courseID` argument in the update query. Similarly, the 
```go
func (s *QuickFeedService) updateSubmissions(request *qf.UpdateSubmissionsRequest) error {
```
and the `qf.UpdateSubmissionsRequest` probably does not need `courseID`, since `AssignmentID` should be unique.

Double check why this field is there.

meling:
This removes the unused argument courseID to the database method UpdateSubmissions (plural).

Fixes #924

### DIFF
--- a/web/courses.go
+++ b/web/courses.go
@@ -292,7 +292,6 @@ func (s *QuickFeedService) updateSubmissions(request *qf.UpdateSubmissionsReques
 	}, false); err != nil {
 		return err
 	}
-
 	query := &qf.Submission{
 		AssignmentID: request.AssignmentID,
 		Score:        request.ScoreLimit,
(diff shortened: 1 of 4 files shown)


web/courses.go
```diff
@@ -292,7 +292,6 @@ func (s *QuickFeedService) updateSubmissions(request *qf.UpdateSubmissionsReques
 	}, false); err != nil {
```
meling, line 292: @0xf8f8ff and @JosteinLindhom: The `getAssignmentWithCourse()` call above seems to be here only to check that the course and assignment exist. Is this necessary? 

The `UpdateSubmissions()` database call (below) also checks via the `Where` clause that the assignment exists.
```go
return db.conn.
	Model(query).
	Where("assignment_id = ?", query.AssignmentID).
	Where("score >= ?", query.Score).
```

0xf8f8ff: Looks absolutely unnecessary to me. I think it should be ok to just remove this method.

0xf8f8ff (approved):

web/courses.go
JosteinLindhom: Agreed

JosteinLindhom (approved):

PR #947, opened 2022-12-06
Serialize timestamp fields in proto messages to time.Time for database compatibility.

### LINKED ISSUE #432: BuildDate is string in score.proto

0xf8f8ff:
An attempt to serialize `timestampbb.Timestamp` fields in protobuf message types to `time.Time` to allow storing in the database as `datetime` type.

There is an issue when using gorm's `Assign` and other methods that use a map. It looks like these methods aren't working with custom serializers, so the timestamp field values must be pre-converted to `time.Time` in cases like this.

Fixes #432.


qf/submission.go
```diff
@@ -14,10 +14,10 @@ func (s *Submission) IsApproved() bool {
 // NewestSubmissionDate returns the submission's build date if newer than the provided submission date.
 // Otherwise, the provided submission date is returned, i.e., if it is newer.
 func (s *Submission) NewestSubmissionDate(submissionDate time.Time) (t time.Time, err error) {
-	if s == nil || s.BuildInfo == nil {
+	if s == nil || s.BuildInfo == nil || s.BuildInfo.BuildDate == nil || s.BuildInfo.SubmissionDate == nil {
```
meling, line 17: Do we need both submissions date and build date here?

qf/submission.go
```diff
@@ -14,10 +14,10 @@ func (s *Submission) IsApproved() bool {
...
 		return t, ErrMissingBuildInfo
 	}
-	currentSubmissionDate, err := time.Parse(TimeLayout, s.BuildInfo.SubmissionDate)
+	currentSubmissionDate := s.BuildInfo.SubmissionDate.AsTime()
```
meling, line 20: If instead we use the getter methods for build info and submission date I think we maybe don’t need the nil checking above?

qf/submission.go
```diff
@@ -14,10 +14,10 @@ func (s *Submission) IsApproved() bool {
...
 	if err != nil {
```
meling, line 21: This check is not necessary anymore

qf/submission.go
0xf8f8ff: We can totally skip the nil checking. I assumed the point was to return with an error immediately in case `BuildInfo` is missing.

qf/submission.go
0xf8f8ff: Not really, removed the build date one.

database/timestamp.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"google.golang.org/protobuf/types/known/timestamppb"
+	"gorm.io/gorm/schema"
+)
+
+var ErrUnsupportedType = errors.New("unsupported type")
+
+// TimestampSerializer is a GORM serializer that allows the serialization and deserialization of the
+// google.protobuf.Timestamp protobuf message type.
+type TimestampSerializer struct{}
```
meling, line 17: It would be great if you could add a comment here and maybe in other places about the advantages of using `time.Time` as the internal database representation.

kit/score/score.proto
```diff
@@ -2,27 +2,30 @@ syntax = "proto3";
...
-    string BuildDate = 3;
-    string BuildLog = 4;
-    int64 ExecTime = 5;
-    string SubmissionDate = 6;
+    uint64 ID           = 1;
+    uint64 SubmissionID = 2 [(go.field) = { tags: 'gorm:"foreignKey:ID"' }];
+    string BuildLog     = 3;
+    int64 ExecTime      = 4;
+
+    google.protobuf.Timestamp BuildDate      = 5 [(go.field) = { tags: 'gorm:"serializer:timestamp;type:time"' }];
```
meling, line 29: See also comment in `types.proto`. Could you also add a doc comment at the top of the file that explains the tag (and other tags):
```proto
[(go.field) = { tags: 'gorm:"serializer:timestamp;type:time"' }];
```
Can be copied from the `types.proto` file.

meling: Note that we should use the `'` single quote ticks instead of the `"` because then we can use the `"` inside the tag description... I made this change in a commit just now.

public/src/Helpers.ts
```diff
@@ -28,8 +29,7 @@ export enum ConnStatus {
 const months = ["January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"]
 
 /** Returns a string with a prettier format for a deadline */
-export const getFormattedTime = (deadline_string: string): string => {
-    const deadline = new Date(deadline_string)
+export const getFormattedTime = (deadline: Date): string => {
```
meling, line 32: We can replace this `deadline: Date` argument and do the following:
```go
export const getFormattedTime = (timestamp: Timestamp | undefined): string => {
    if (!timestamp) {
        return "N/A"
    }
    const deadline = timestamp.toDate()
    return `${deadline.getDate()} ${months[deadline.getMonth()]} ${deadline.getFullYear()} ${deadline.getHours()}:${deadline.getMinutes() < 10 ? "0" : ""}${deadline.getMinutes()}`
}
```
Then we can simplify the use of all uses of this function; we always do this (or equal alternative):
```typescript
assignment.deadline ? getFormattedTime(assignment.deadline.toDate()) : "N/A"
```
Which would become:
```typescript
getFormattedTime(assignment.deadline)
```

public/src/components/CourseLabs.tsx
```diff
@@ -41,7 +41,7 @@ const CourseLabs = (): JSX.Element => {
                             {assignmentStatusText(assignment, submission)}
                         </div>
                         <div className="col-4 text-center">
-                            {getFormattedTime(assignment.deadline)}
+                            {assignment.deadline ? getFormattedTime(assignment.deadline.toDate()) : "N/A"}
```
meling, line 44: See suggestion in `Helpers.ts`

public/src/components/LabResultTable.tsx
```diff
@@ -84,13 +84,13 @@ const LabResultTable = ({ submission, assignment }: lab): JSX.Element => {
...
                                     <td colSpan={2}>Approved</td>
-                                    <td>{getFormattedTime(submission.approvedDate)}</td>
+                                    <td>{getFormattedTime(submission.approvedDate.toDate())}</td>
                                 </tr>
                                 : null
                         }
                         <tr>
                             <td colSpan={2}>Deadline</td>
-                            <td>{getFormattedTime(assignment.deadline)}</td>
+                            <td>{assignment.deadline ? getFormattedTime(assignment.deadline.toDate()) : "N/A"}</td>
```
meling, line 93: See suggestion in `Helpers.ts`

public/src/components/Members.tsx
```diff
@@ -46,7 +46,7 @@ const Members = (): JSX.Element => {
         data.push(enrollment.user ? enrollment.user.Name : "")
         data.push(enrollment.user ? enrollment.user.Email : "")
         data.push(enrollment.user ? enrollment.user.StudentID : "")
-        data.push(enrollment.lastActivityDate)
+        data.push(enrollment.lastActivityDate ? getFormattedTime(enrollment.lastActivityDate.toDate()) : "N/A")
```
meling, line 49: See suggestion in `Helpers.ts`

public/src/components/dashboard/SubmissionsTable.tsx
```diff
@@ -57,7 +57,7 @@ const SubmissionsTable = (): JSX.Element => {
                                 <span className="badge ml-2 float-right"><i className="fa fa-users" title="Group Assignment" /></span> : null}
                         </td>
                         <td><ProgressBar assignmentIndex={assignment.order - 1} courseID={courseID.toString()} submission={submission} type={Progress.OVERVIEW} /></td>
-                        <td>{getFormattedTime(assignment.deadline)}</td>
+                        <td>{assignment.deadline}</td>
```
meling, line 60: This is interesting. Don't we need to format the timestamp here? If the default formatter is nice, then maybe we don't need the `getFormattedTime()`; hehe...

public/src/components/dashboard/SubmissionsTable.tsx
```diff
@@ -17,18 +17,18 @@ const SubmissionsTable = (): JSX.Element => {
             assignments.push(...state.assignments[courseID])
         }
         assignments.sort((a, b) => {
-            if (b.deadline > a.deadline) {
+            if ((a.deadline && b.deadline) && (b.deadline > a.deadline)) {
```
meling, line 20: I was going to suggest this as more readable code:
```ts
            if (!a.deadline || !b.deadline) {
                return 0
            }
            if (a.deadline < b.deadline) {
                return -1
            }
            if (a.deadline > b.deadline) {
                return 1
            }
```

But then copilot suggested this:
```typescript
            if (a.deadline && b.deadline) {
                return a.deadline.toDate().getTime() - b.deadline.toDate().getTime()
            }
            return 0
```
I also tried:
```ts
       return a.deadline.seconds - b.deadline.seconds
```
But `seconds` are `bigint` (not a number) and didn't work with the sort api.

public/src/components/dashboard/SubmissionsTable.tsx
```diff
@@ -38,15 +38,15 @@ const SubmissionsTable = (): JSX.Element => {
             }
             // Submissions are indexed by the assignment order - 1.
             const submission = submissions[assignment.order - 1] ?? new Submission()
-            if (submission.status !== Submission_Status.APPROVED) {
+            if (submission.status !== Submission_Status.APPROVED && assignment.deadline) {
                 const deadline = timeFormatter(assignment.deadline)
                 if (deadline.daysUntil > 3 && submission.score >= assignment.scoreLimit) {
```
meling, line 43: What about moving this logic into the `timeFormatter(assignment, submission)`?

public/src/components/dashboard/SubmissionsTable.tsx
```diff
@@ -57,7 +57,7 @@ const SubmissionsTable = (): JSX.Element => {
                                 <span className="badge ml-2 float-right"><i className="fa fa-users" title="Group Assignment" /></span> : null}
                         </td>
                         <td><ProgressBar assignmentIndex={assignment.order - 1} courseID={courseID.toString()} submission={submission} type={Progress.OVERVIEW} /></td>
-                        <td>{getFormattedTime(assignment.deadline)}</td>
+                        <td>{assignment.deadline}</td>
                         <td>{deadline.message ? deadline.message : '--'}</td>
```
meling, line 61: We should move this check and set the `deadline.message` to `--` in `timeFormatter()`. See also check Line 46.

public/src/components/dashboard/SubmissionsTable.tsx
```diff
@@ -38,15 +38,15 @@ const SubmissionsTable = (): JSX.Element => {
             }
             // Submissions are indexed by the assignment order - 1.
             const submission = submissions[assignment.order - 1] ?? new Submission()
-            if (submission.status !== Submission_Status.APPROVED) {
+            if (submission.status !== Submission_Status.APPROVED && assignment.deadline) {
                 const deadline = timeFormatter(assignment.deadline)
                 if (deadline.daysUntil > 3 && submission.score >= assignment.scoreLimit) {
                     deadline.className = "table-success"
                 }
                 if (!deadline.message) {
```
meling, line 46: We should always set `deadline.message` to a valid string, e.g., `--` if there is no message (see Line 61).

public/src/components/dashboard/SubmissionsTable.tsx
meling: That is, it becomes just:
```ts
   <td>{deadline.message}</td>
```

public/src/components/dashboard/SubmissionsTable.tsx
meling: That said since we also use the `Deadline` (interface type) for other formatted stuff, perhaps it would also make sense to add the actual deadline to the `Deadline` type, as `deadline.deadline`. (Not sure it is needed though, if it already works correctly as it is.)

qf/assignment.go
```diff
@@ -15,13 +15,8 @@ const (
 // SinceDeadline returns the duration since the deadline.
 // A positive duration means the deadline has passed, whereas
 // a negative duration means the deadline has not yet passed.
-func (a *Assignment) SinceDeadline(now time.Time) (time.Duration, error) {
-	deadline, err := time.ParseInLocation(TimeLayout, a.GetDeadline(), now.Location())
-	if err != nil {
-		// this should not happen if deadlines are parsed and recorded correctly
-		return zero, err
```
meling, line 22: You can remove `zero` on Line 12.

qf/submission.go
meling: This is what I meant:
```go
	currentSubmissionDate := s.GetBuildInfo().GetSubmissionDate().AsTime()
	if currentSubmissionDate.IsZero() {
		return t, ErrMissingBuildInfo
	}
```
This is possible since even `s == nil` will work fine since `GetBuildInfo()` checks for this and returns `nil` and so on. All the way to `AsTime()`, which will return 0 for each component of `timestamppb.Timestamp` instead of `nil`, which should result in returning the zero time.

PS: I didn't test the code, so please add a test for this... to be sure.

meling: I fixed this by removing the need for even returning an error. Leaving the comment unresolved FYI.

database/timestamp.go
meling: I noticed that the PR comment explains this but won't hurt to have it here as well.

public/src/components/dashboard/SubmissionsTable.tsx
meling: Seems like there is prior art for doing it as copilot suggested. See `Helpers.ts`:
```ts
        case EnrollmentSort.Activity:
            if (a.lastActivityDate && b.lastActivityDate) {
                return sortOrder * (a.lastActivityDate.toDate().getTime() - b.lastActivityDate.toDate().getTime()) 
            }
            return 0               
```

meling (changes requested):
This is looking very nice!

public/src/Helpers.ts
```diff
@@ -28,8 +29,7 @@ export enum ConnStatus {
...
     return `${deadline.getDate()} ${months[deadline.getMonth()]} ${deadline.getFullYear()} ${deadline.getHours()}:${deadline.getMinutes() < 10 ? "0" : ""}${deadline.getMinutes()}`
```
meling, line 33: Here is a more readable version of this (from #704), please use this:
```ts
const months = ["January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"]
const minutes = date.getMinutes()
const zero = minutes < 10 ? "0" : ""
return `${date.getDate()} ${months[date.getMonth()]} ${date.getFullYear()} ${date.getHours()}:${zero}${minutes}`
```
(I also moved the `months` into the func, since it was only used here... not sure if that is smart or not though...)

public/src/Helpers.ts
```diff
@@ -41,8 +41,8 @@ export interface Deadline {
 
 /** Utility function for LandingPageTable functionality. To format the output string and class/css based on how far the deadline is in the future */
```
meling, line 42: Better formatting:
```ts
/**
 * Utility function for LandingPageTable to format the output string and class/css
 * depending on how far into the future the deadline is.
 *
 * layoutTime = "2021-03-20T23:59:00"
 */
```

public/src/components/dashboard/SubmissionsTable.tsx
0xf8f8ff: > This is interesting. Don't we need to format the timestamp here? If the default formatter is nice, then maybe we don't need the `getFormattedTime()`; hehe...

Yes, my bad, we do need to format it here as well.

database/gormdb.go
```diff
@@ -42,12 +43,15 @@ type GormDB struct {
 // NewGormDB creates a new gorm database using the provided driver.
 func NewGormDB(path string, logger *zap.Logger) (*GormDB, error) {
 	conn, err := gorm.Open(sqlite.Open(path), &gorm.Config{
-		Logger: NewGORMLogger(logger),
+		Logger:                 NewGORMLogger(logger),
+		SkipDefaultTransaction: true,
```
JosteinLindhom, line 47: Having this set to `true` may lead to some unexpected results. 
This is especially true for nested types, e.g `Submission{Scores:{}}`. When creating a submission, the submission will be created successfully even though an error occurs when creating the scores. The current behaviour is that the entire creation is rolled back if at any point an error occurs. The same goes for deleting and updating.

public/src/components/dashboard/SubmissionsTable.tsx
0xf8f8ff: I'm not sure I understand the logic on Line 46 well enough to mess with it. Would I need to change the check to `if deadline.message == "--"` ? Or is it ok to always render the rest?

public/src/Helpers.ts
JosteinLindhom: Doubt we need to worry about the following, but I'll mention it nonetheless:

Moving `months` inside the function will cause `months` to be recreated every time the function is called. To avoid this we can create a local closure over the function:

```ts
export const getFormattedTime = (() => {
    const months = ["January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"]
    return (deadline: Timestamp) => {
        const date = deadline.toDate()
        const zero = date.getMinutes() < 10 ? "0" : ""
        return `${date} ${months[date.getMonth()]} ${date.getFullYear()} ${date.getHours()}:${zero}${date.getMinutes()}`
    }
})()
```

Where `months` is no longer in the global namespace, nor is it recreated when calling `getFormattedTime`

meling: That is interesting.

meling (approved):
Let's leave the remaining issues in `SubmissionTable` for another day.

PR #964, opened 2022-12-19
Testing with github app

### LINKED ISSUE #649: Update the Pull request support to use the new GitHub web app approach
Look for the string: `TODO(espeland): Update this for GitHub web app.`

These instances should be updated for the new GitHub web app implemented by @0xf8f8ff. If my recall is correct, the current approach creates PR comments as the course creator (typically the teacher). We should avoid this and let the actual TA or student's named be used on comments.

@oleespe Please add your insights regarding what is needed for updating to use the web app approach.

meling:
- Reorganized .env and public/.env handling
- Added CreateQuickFeedApp helper for creating a github app for testing purposes.
- Added two helpers GetAppSCM and GetSCMManager

This PR confirms that the App owner is used as the creator of the issues (and most probably pull requests) instead of the course creator.

Fixes #649.


JosteinLindhom:
The issues with invalid webhook URLs can be avoided like this in the template:

```go
...
{{if .WebhookActive}}
"hook_attributes": {
	"active": {{.WebhookActive}},
	"url": "{{.WebhookURL}}",
},
{{end}}
...
{{if .WebhookActive}}
"default_events": [
	"push",
	"pull_request",
	"pull_request_review"
]
{{end}}
...
```

We can only omit the `hook_attributes` if we do not include any `default_events`. The `WebhookActive` can be set to false using the `IsLocal` introduced in #966.

```go
if env.IsLocal(domain) {
	// Disable webhook for localhost, or any other non-public domain
	data.WebhookURL = ""
	data.WebhookActive = false
}
```

assignments/tasks_test.go
```diff
@@ -43,6 +43,11 @@ func TestSynchronizeTasksWithIssues(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// Deleting issues failed with:
```
meling, line 46: @JosteinLindhom I would like to also fix this one before merging this one... let me know if you have found any clues regarding this. I'll see if I can figure it out as well...

PR #976, opened 2023-01-07
Simplify Organization message type

### LINKED ISSUE #971: Chore: Simplify Organization message type
The `Organization` message type is not stored in the database and is only a response message. We no longer use the `PaymentPlan` field and as far as I can tell we don't use the `Avatar` field either (for the organization). Moreover, we should rename the `ID` and `name` fields to comply with our convention. Hence, I propose to change the following:
```proto
message Organization {
    uint64 ID          = 1;
    string name        = 2;
    string avatar      = 3;
    string paymentPlan = 4;
}
```
To:
```proto
message Organization {
    uint64 ScmOrganizationID          = 1;
    string ScmOrganizationName        = 2;
}
```
We can also replace the `OrgRequest` message with `Organization`; keeping only a single message type to represent both a request and response message.

meling:
This removes PaymentPlan and avatar from the Organization message type, since these fields were not used (anymore).

We also replace OrgRequest with just Organization, since both hold almost the same information.

Fixes #971


JosteinLindhom (approved):
Looks good!

PR #977, opened 2023-01-07
Removed deprecated yaml assignmentid field

meling:
We no longer support the `assignmentid` field in `assignment.yaml` files; users should use the `order` field to define the order number of an assignment. See the dat320 and dat520 courses for an example. @JosteinLindhom We just need to make sure that the dat310 course and possibly others don't use `assignmentid` before merging and using this code.

### DIFF
--- a/assignments/assignments_parser.go
+++ b/assignments/assignments_parser.go
@@ -16,7 +16,6 @@ const defaultAutoApproveScoreLimit = 80
 // Note that the struct can be private, but the fields must be
 // public to allow parsing.
 type assignmentData struct {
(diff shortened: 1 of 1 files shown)


meling:
@JosteinLindhom ping; I don't think there is much to review here, except checking that dat310 isn't using `assignmentid`.

JosteinLindhom:
> @JosteinLindhom ping; I don't think there is much to review here, except checking that dat310 isn't using `assignmentid`.

We used `assignmentid` for the first assignment in DAT310, but switched to `order` for subsequent assignments.

JosteinLindhom (approved):

PR #980, opened 2023-01-11
Fixes for manual grading

JosteinLindhom:
This PR includes some fixes for manual grading in the frontend.

- Fetch reviewer from course enrollments (previously from `users`). Context: teachers that are not admins do not have access to the `GetUsers` endpoint. As a consequence, `users` will not be populated for non-admin teachers
- Fixed bug where `isAuthor` used review ID rather than reviewer ID

### DIFF
--- a/public/src/overmind/state.tsx
+++ b/public/src/overmind/state.tsx
@@ -234,13 +234,12 @@ export const state: State = {
         if (!activeCourse || !courseEnrollments[activeCourse.toString()]) {
             return {}
         }
-        const teachers = courseEnrollments[activeCourse.toString()].filter(isTeacher)
         const teachersMap: { [userID: string]: User } = {}
-        for (const teacher of teachers) {
-            if (teacher.user) {
-                teachersMap[teacher.userID.toString()] = teacher.user
+        courseEnrollments[activeCourse.toString()].forEach(enrollment => {
+            if (isTeacher(enrollment) && enrollment.user) {
+                teachersMap[enrollment.userID.toString()] = enrollment.user
             }
-        }
+        })
         return teachersMap
     }),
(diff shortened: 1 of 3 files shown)


public/src/overmind/state.tsx
```diff
@@ -226,6 +230,19 @@ export const state: State = {
     users: {},
     allUsers: [],
     courses: [],
+    courseTeachers: derived(({ courseEnrollments, activeCourse }: State) => {
+        if (!activeCourse || !courseEnrollments[activeCourse.toString()]) {
+            return {}
+        }
+        const teachers = courseEnrollments[activeCourse.toString()].filter(isTeacher)
```
meling, line 237: Here you iterate through all users enrolled and then iterate again through the teachers. Wouldn’t it be better to just iterate once over all enrolled users and check if it is a teacher and directly add it to the teachersMap?

JosteinLindhom: Fixed.

meling: What I meant is that `filter(isTeacher)` is actually an iteration over `courseEnrollments`... My proposal was to do:
```ts
for (const enrolled of courseEnrollments[activeCourse.toString()]) {
  if (isTeacher(enrolled)) {
...
```

JosteinLindhom: Appears I forgot to actually push the fix before I left the office.. 🤡 The "fix" I refer to does what you propose, except to use 
```ts
courseEnrollments[activeCourse.toString()].forEach(enrollment => {
    ...
})

meling: sounds good

meling (approved):

PR #983, opened 2023-02-08
Fixed GroupForm bug

### LINKED ISSUE #982: Teachers are not able to see students when editing groups
A change introduced in efb362b74403fc3937cf584788b70597937a8f71 enabled teachers to view *only other teachers* in the group form.

Sometimes a teacher needs to edit student groups. The change prevent teachers from seeing any students in the list of enrollments without a group. As a consequence, a teacher is not able to add students to student groups.

A fix for this issue could be to allow teachers to see both teachers *and* students in the group form.

JosteinLindhom:
Fixes #982.

### DIFF
--- a/public/src/components/group/GroupForm.tsx
+++ b/public/src/components/group/GroupForm.tsx
@@ -94,6 +94,25 @@ const GroupForm = (): JSX.Element | null => {
         )
     })
 
+    const toggleEnrollmentType = () => {
+        if (hasTeacher(enrollmentType)) {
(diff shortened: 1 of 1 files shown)


meling:
This is okay, but maybe we should add a toggle button:
- students only
- teachers only

JosteinLindhom:
> This is okay, but maybe we should add a toggle button:
> 
> * students only
> * teachers only

Added a toggle for teachers. I also fixed a bug where students would not show in the list of available users if removed from a group.

meling (approved):
LGTM

PR #990, opened 2023-03-27
Results overlap fix

JosteinLindhom:
This PR fixes an issue where a submission / review would overlap with the submissions table in both `Results` and `Review` views.

- Both columns in these views are now the same size
- Assignments that overflow the horizontal width of the submissions table are not shown
    - In these cases the column is horizontally scrollable by clicking and dragging


meling (approved):
Looks good to me!! (But I haven't seen it in action...)

But take a look at my comment about `table-fix` name...

PR #993, opened 2023-04-20
Remove grpc manager

### LINKED ISSUE #970: Remove GRPCManager and (maybe also) MockGRPCManager?
The legacy GRPCManager (and its reflection-based approach) should be replaced with the idiomatic connect-web approach. See the [connect-web documentation](https://connect.build/docs/web/using-clients) for the idiomatic way to make calls from the client. There are several variants available. It should be implemented after merging current changes to the frontend.

JosteinLindhom:
This PR removes both `GRPCManager` and `MockGrpcManager`.

Closes #970


public/src/overmind/actions.tsx
```diff
@@ -54,36 +55,43 @@ export const receiveSubmission = ({ state }: Context, submission: Submission): v
...
-        state.enrollments = result.data.enrollments
-        for (const enrollment of state.enrollments) {
-            state.status[enrollment.courseID.toString()] = enrollment.status
+    const response = await effects.client.getEnrollments({
+        FetchMode: {
+            case: "userID",
+            value: state.self.ID,
         }
+    })
+    if (response.error) {
```
meling, line 74: Should this error be shown to the user somewhere in the frontend?

public/src/client.ts
```diff
@@ -0,0 +1,121 @@
...
+            case MethodKind.ServerStreaming:
+                return createServerStreamingFn(transport, service, method);
+            default:
+                return null;
+        }
+    }) as ResponseClient<T>;
+}
+
+/**
+  * UnaryFn is the method signature for a unary method of a ResponseClient.
```
meling, line 52: Nit: Indentation should be same...
```ts
/**
 * UnaryFn is the method signature for a unary method of a ResponseClient.
 */
```

public/src/client.ts
```diff
@@ -0,0 +1,121 @@
...
+    error: ConnectError | null
+    message: T
+}
+
+/**
+ * ResponseClient is a simple client that supports unary and server-streaming
+ * methods. Methods will produce a promise for the response message,
+ * or an asynchronous iterable of response messages.
+ */
+export type ResponseClient<T extends ServiceType> = {
```
meling, line 24: Not sure it makes sense, but is there a test case for this stuff that we could add to show how this is actually used.

public/src/__tests__/updateEnrollment.test.tsx
```diff
@@ -1,18 +1,53 @@
...
+import { CallOptions } from "@bufbuild/connect"
+import { Response } from "../client"
+import { initializeOvermind } from "./TestHelpers"
 
 
 describe("UpdateEnrollment", () => {
-    const mockedOvermind = initializeOvermind({})
+    const mockedOvermind = initializeOvermind({}, {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        getEnrollments: jest.fn(async (request: PartialMessage<EnrollmentRequest>, _?: CallOptions | undefined): Promise<Response<Enrollments>> => {
```
meling, line 21: It looks a bit awkward, but if this is how it needs to be then that's how it needs to be... just wish it was more readable.

public/src/overmind/actions.tsx
meling: Maybe we could have an error log, but not sure it makes sense to expose such errors to the user (other than us)...

public/src/overmind/actions.tsx
```diff
@@ -279,88 +304,96 @@ export const getRepositories = async ({ state, effects }: Context): Promise<void
...
-        const response = await effects.grpcMan.getRepositories(courseID)
-        if (response.data) {
-            state.repositories[courseID.toString()] = response.data.URLs
+        const response = await effects.client.getRepositories({ courseID })
+        if (response.error) {
+            return
         }
-    }))
-}
+        state.repositories[courseID.toString()] = response.message.URLs
```
meling, line 312: rename blank line

public/src/overmind/actions.tsx
```diff
@@ -279,88 +304,96 @@ export const getRepositories = async ({ state, effects }: Context): Promise<void
...
 }
 
 /** getSubmissions fetches all submission for the current user by Course ID and stores them in state */
 // TODO: Currently not used, see refreshSubmissions.
 export const getSubmissions = async ({ state, effects }: Context, courseID: bigint): Promise<void> => {
-    const result = await effects.grpcMan.getSubmissions(courseID, state.self.ID)
-    if (result.data) {
-        state.submissions[courseID.toString()] = result.data.submissions
-    }
+    const result = await effects.client.getSubmissions({
```
meling, line 372: rename to response for consistency? Or delete method if not used?

public/src/overmind/actions.tsx
```diff
@@ -384,43 +417,59 @@ export const loadCourseSubmissions = async ({ state, actions }: Context, courseI
...
 export const getGroupsByCourse = async ({ state, effects }: Context, courseID: bigint): Promise<void> => {
     state.groups[courseID.toString()] = []
-    const res = await effects.grpcMan.getGroupsByCourse(courseID)
-    if (res.data) {
-        state.groups[courseID.toString()] = res.data.groups
-    }
+    const response = await effects.client.getGroupsByCourse({ courseID })
+    if (response.error) { return }
+    state.groups[courseID.toString()] = response.message.groups
+
```
meling, line 455: remove blank line

public/src/overmind/actions.tsx
```diff
@@ -490,52 +555,79 @@ export const rebuildSubmission = async ({ state, actions, effects }: Context, {
...
-        const enrollments = await effects.grpcMan.getEnrollmentsByUser(state.self.ID)
-        if (enrollments.data) {
-            state.enrollments = enrollments.data.enrollments
-        }
+    const response = await effects.client.createEnrollment({
+        courseID: courseID,
+        userID: state.self.ID,
+    })
+    if (!success(response)) {
+        // Alert user that enrollment failed
```
meling, line 572: Missing alert?

public/src/overmind/actions.tsx
```diff
@@ -490,52 +555,79 @@ export const rebuildSubmission = async ({ state, actions, effects }: Context, {
...
+        return
     }
+    const deleteResponse = await effects.client.deleteGroup({
+        courseID: group.courseID,
+        groupID: group.ID,
+    })
+    if (success(deleteResponse)) {
+        state.groups[group.courseID.toString()] = state.groups[group.courseID.toString()].filter(g => g.ID !== group.ID)
+    }
+
```
meling, line 618: remove blank line

public/src/__tests__/updateEnrollment.test.tsx
JosteinLindhom: I've introduced a `mock` function that should make mocking individual rpc methods easier and more readable.

public/src/__tests__/TestHelpers.ts
```diff
@@ -72,12 +73,35 @@ export const setupDrivers = (path?: string): ThenableWebDriver[] => {
...
+type Methods = UnaryApiClient["client"];
+
+/** mock is a helper function that takes a method and a mocked function to run in place of the method.
+ *  It returns a function that can be used to replace the method in the ApiClient.
+ * @param _method the method to mock
+ * @param mockFn the function to run in place of the method
+ * @example: mock("getSubmission", async (req) => { return { error: null, message: new Submission() } })
+*/
+export function mock<T extends keyof Methods>(
+    _method: T,
```
meling, line 101: Why use `_method` here and in a few other places?

public/src/__tests__/TestHelpers.ts
```diff
@@ -72,12 +73,35 @@ export const setupDrivers = (path?: string): ThenableWebDriver[] => {
...
+    mockFn: (req: Parameters<Methods[T]>[0]) => ReturnType<Methods[T]>
+): Methods[T] {
+    return async function (...args: any[]): Promise<any> {
+        return mockFn(args);
+    };
+}
```
meling, line 107: need line break.

public/src/overmind/actions.tsx
```diff
@@ -190,84 +215,86 @@ export const updateEnrollment = async ({ state, effects }: Context, { enrollment
...
-            if (status === Enrollment_UserStatus.NONE) {
-                // If the enrollment is rejected, remove it from state
-                enrollments.splice(found, 1)
-            } else {
-                // If the enrollment is accepted, update the enrollment in state
-                enrollments[found].status = status
-            }
+    // Send updated enrollment to server
+    const response = await effects.api.client.updateEnrollments({ enrollments: [temp] })
+    if (success(response)) {
```
meling, line 241: Seems to break the pattern (now adopted elsewhere) of checking for `if (response.error)` and returning early and keeping the "happy path" non-indented below the error-guard.

public/src/overmind/actions.tsx
```diff
@@ -190,84 +215,86 @@ export const updateEnrollment = async ({ state, effects }: Context, { enrollment
...
+    // We do not want to update set the enrollment status before the update is successful.
+    const enrollments = state.pendingEnrollments.map(e => {
+        const temp = e.clone()
+        temp.status = Enrollment_UserStatus.STUDENT
+        return temp
+    })
+
+    // Send updated enrollments to server
+    const response = await effects.api.client.updateEnrollments({ enrollments })
+    if (success(response)) {
```
meling, line 270: Suggest to flip the if, and use the common pattern `if (response.error)` and keep the happy path non-indented after the `await ... ` and a return.

public/src/overmind/actions.tsx
```diff
@@ -382,45 +369,58 @@ export const loadCourseSubmissions = async ({ state, actions }: Context, courseI
...
 }
 
 export const getGroupsByCourse = async ({ state, effects }: Context, courseID: bigint): Promise<void> => {
     state.groups[courseID.toString()] = []
-    const res = await effects.grpcMan.getGroupsByCourse(courseID)
-    if (res.data) {
-        state.groups[courseID.toString()] = res.data.groups
-    }
+    const response = await effects.api.client.getGroupsByCourse({ courseID })
+    if (response.error) { return }
```
meling, line 405: Prefer to maintain a consistent pattern like:
```ts
if (response.error) {
   return
}
```

public/src/overmind/actions.tsx
```diff
@@ -459,17 +465,23 @@ export const setSelectedSubmission = ({ state }: Context, submission: Submission
...
-    const response = await effects.grpcMan.getSubmission(courseID, submission.ID)
-    if (!response.data || !success(response)) {
+    const response = await effects.api.client.getSubmission({
+        CourseID: courseID,
+        FetchMode: {
+            case: "SubmissionID",
+            value: submission.ID,
+        },
+    })
+    if (!success(response)) {
```
meling, line 475: I think this should be replaced with:
```ts
    if (response.error) {
        return
    }
```

public/src/overmind/actions.tsx
```diff
@@ -478,10 +490,14 @@ export const rebuildSubmission = async ({ state, actions, effects }: Context, {
     if (!(submission && state.selectedAssignment && state.activeCourse)) {
         return
     }
-    const response = await effects.grpcMan.rebuildSubmission(state.selectedAssignment.ID, submission.ID, state.activeCourse)
+    const response = await effects.api.client.rebuildSubmissions({
+        courseID: state.activeCourse,
+        assignmentID: state.selectedAssignment.ID,
+        submissionID: submission.ID,
+    })
     if (success(response)) {
```
meling, line 498:
```ts
    if (response.error) {
        return
    }
```

public/src/overmind/actions.tsx
```diff
@@ -490,52 +506,76 @@ export const rebuildSubmission = async ({ state, actions, effects }: Context, {
...
-    if (success(response)) {
-        const enrollments = await effects.grpcMan.getEnrollmentsByUser(state.self.ID)
-        if (enrollments.data) {
-            state.enrollments = enrollments.data.enrollments
-        }
+    const response = await effects.api.client.createEnrollment({
+        courseID,
+        userID: state.self.ID,
+    })
+    if (!success(response)) {
```
meling, line 522:
```ts
    if (response.error) {
        return
    }
```

public/src/overmind/actions.tsx
```diff
@@ -490,52 +506,76 @@ export const rebuildSubmission = async ({ state, actions, effects }: Context, {
 
 /* rebuildAllSubmissions rebuilds all submissions for a given assignment */
 export const rebuildAllSubmissions = async ({ effects }: Context, { courseID, assignmentID }: { courseID: bigint, assignmentID: bigint }): Promise<boolean> => {
-    const response = await effects.grpcMan.rebuildSubmissions(assignmentID, courseID)
+    const response = await effects.api.client.rebuildSubmissions({
+        courseID,
+        assignmentID,
+    })
     return success(response)
```
meling, line 513: Looks like we might end up removing the `success` function; in which case this should return `!response.error`

public/src/overmind/actions.tsx
```diff
@@ -490,52 +506,76 @@ export const rebuildSubmission = async ({ state, actions, effects }: Context, {
...
+        return
     }
+    const enrollments = await effects.api.client.getEnrollments({
+        FetchMode: {
+            case: "userID",
+            value: state.self.ID,
+        }
+    })
+
+    if (enrollments.error) { return }
```
meling, line 532:
```ts
    if (response.error) {
        return
    }
```

public/src/overmind/actions.tsx
```diff
@@ -490,52 +506,76 @@ export const rebuildSubmission = async ({ state, actions, effects }: Context, {
...
+    state.enrollments = enrollments.message.enrollments
+
 }
 
 export const updateGroupStatus = async ({ effects }: Context, { group, status }: { group: Group, status: Group_GroupStatus }): Promise<void> => {
     const oldStatus = group.status
     group.status = status
-    const response = await effects.grpcMan.updateGroup(group)
+    const response = await effects.api.client.updateGroup(group)
     if (!success(response)) {
```
meling, line 541:
```ts
    if (response.error) {
        return
    }
```

public/src/overmind/actions.tsx
```diff
@@ -490,52 +506,76 @@ export const rebuildSubmission = async ({ state, actions, effects }: Context, {
...
+    }
+
+    if (!confirm(`Warning! Group repository is not empty! Do you still want to delete group, github team and group repository?`)) {
+        return
+    }
+    const deleteResponse = await effects.api.client.deleteGroup({
+        courseID: group.courseID,
+        groupID: group.ID,
+    })
+    if (success(deleteResponse)) {
```
meling, line 565:
```ts
    if (response.error) {
        return
    }
```

public/src/overmind/actions.tsx
```diff
@@ -547,44 +587,44 @@ export const createOrUpdateCriterion = async ({ effects }: Context, { criterion,
...
         const index = benchmark.criteria.findIndex(c => c.ID === criterion.ID)
         if (index > -1) {
             benchmark.criteria[index] = criterion
         }
     } else {
-        const response = await effects.grpcMan.createCriterion(criterion)
-        if (success(response) && response.data) {
-            benchmark.criteria.push(response.data)
+        const response = await effects.api.client.createCriterion(criterion)
+        if (success(response)) {
```
meling, line 597:
```ts
    if (response.error) {
        return
    }
```

public/src/overmind/actions.tsx
```diff
@@ -547,44 +587,44 @@ export const createOrUpdateCriterion = async ({ effects }: Context, { criterion,
     }
 
     // Existing criteria have a criteria id > 0, new criteria have a criteria id of 0
-    if (criterion.ID && success(await effects.grpcMan.updateCriterion(criterion))) {
+    if (criterion.ID && success(await effects.api.client.updateCriterion(criterion))) {
```
meling, line 590: We probably want to follow the pattern also here and move the call into the if statement and do:
```ts
    if (response.error) {
        return
    }
```

public/src/overmind/actions.tsx
```diff
@@ -547,44 +587,44 @@ export const createOrUpdateCriterion = async ({ effects }: Context, { criterion,
...
         const index = assignment.gradingBenchmarks.indexOf(benchmark)
         if (index > -1) {
             assignment.gradingBenchmarks[index] = benchmark
         }
     } else {
-        const response = await effects.grpcMan.createBenchmark(bm)
-        if (success(response) && response.data) {
-            assignment.gradingBenchmarks.push(response.data)
+        const response = await effects.api.client.createBenchmark(benchmark)
+        if (success(response)) {
```
meling, line 613:
```ts
    if (response.error) {
        return
    }
```

public/src/overmind/actions.tsx
```diff
@@ -547,44 +587,44 @@ export const createOrUpdateCriterion = async ({ effects }: Context, { criterion,
...
+            benchmark.criteria.push(response.message)
         }
     }
 }
 
 export const createOrUpdateBenchmark = async ({ effects }: Context, { benchmark, assignment }: { benchmark: GradingBenchmark, assignment: Assignment }): Promise<void> => {
     // Check if this need cloning
     const bm = benchmark.clone()
-    if (benchmark.ID && success(await effects.grpcMan.updateBenchmark(bm))) {
+    if (benchmark.ID && success(await effects.api.client.updateBenchmark(bm))) {
```
meling, line 606: Same thing as above.

public/src/overmind/actions.tsx
```diff
@@ -547,44 +587,44 @@ export const createOrUpdateCriterion = async ({ effects }: Context, { criterion,
...
+            assignment.gradingBenchmarks.push(response.message)
         }
     }
 }
 
 export const createBenchmark = async ({ effects }: Context, { benchmark, assignment }: { benchmark: GradingBenchmark, assignment: Assignment }): Promise<void> => {
     benchmark.AssignmentID = assignment.ID
-    const response = await effects.grpcMan.createBenchmark(benchmark)
+    const response = await effects.api.client.createBenchmark(benchmark)
     if (success(response)) {
```
meling, line 622:
```ts
    if (response.error) {
        return
    }
```

public/src/overmind/actions.tsx
```diff
@@ -642,6 +680,23 @@ export const startSubmissionStream = ({ actions, effects }: Context) => {
     })
 }
 
+export const updateAssignments = async ({ effects }: Context, courseID: bigint): Promise<void> => {
+    const response = await effects.api.client.updateAssignments({ courseID })
+    if (success(response)) {
```
meling, line 685:
```ts
    if (response.error) {
        return
    }
```

public/src/overmind/actions.tsx
```diff
@@ -642,6 +680,23 @@ export const startSubmissionStream = ({ actions, effects }: Context) => {
...
+        // alert user if update was successful
+    } else {
+        // alert user if update failed
+    }
+}
+
+export const getGroup = async ({ state, effects }: Context, enrollment: Enrollment): Promise<void> => {
+    const response = await effects.api.client.getGroup({ courseID: enrollment.courseID, groupID: enrollment.groupID })
+    if (success(response)) {
```
meling, line 694:
```ts
    if (response.error) {
        return
    }
```

public/src/overmind/actions.tsx
```diff
@@ -685,8 +740,15 @@ export const fetchUserData = async ({ state, actions }: Context): Promise<boolea
...
-        const status = await effects.grpcMan.getEnrollmentsByUser(state.self.ID, [Enrollment_UserStatus.TEACHER])
-        if (status.data?.enrollments.find(enrol => enrol.courseID === BigInt(courseID) && hasTeacher(enrol.status))) {
+        const response = await effects.api.client.getEnrollments({
+            FetchMode: {
+                case: "userID",
+                value: state.self.ID,
+            },
+            statuses: [Enrollment_UserStatus.TEACHER],
+        })
+        if (response.error) { return }
```
meling, line 750:
```ts
    if (response.error) {
        return
    }
```

meling (changes requested):
Looks nice; sorry for the noisy single comments. I'm not sure it was finished from your side, but I at least see that there are some inconsistencies in how we handle errors, which I feel this PR is a good opportunity to fix them herein.

I only looked at the code, but does the alert handler thing work?

public/src/__tests__/TestHelpers.ts
JosteinLindhom: `method` is not used in the function, but exists to provide better auto completion. Underscoring the parameter tells the compiler that not using the parameter is intentional.

In our `tsconfig` we've enabled `"noUnusedParameters": true,`. With this enabled we get compile errors if there are any unused parameters in our code. ESLint also reports any unused parameters / variables. Using underscores will bypass these restrictions.

Omitting `method: T` altogether would work as long as we provide the method as a type argument:

```ts
mock<"getSubmission">(async () => { ... }) // mockFn is typed correctly
```

However, when adding the method as a type argument you would not get auto completion / the list of all available methods.

public/src/overmind/actions.tsx
```diff
@@ -158,12 +174,18 @@ export const updateSubmission = async ({ state, effects }: Context, { owner, sub
...
+    const response = await effects.api.client.updateSubmission({
+        courseID: state.activeCourse,
+        submissionID: submission.ID,
+        status: submission.status,
+        released: submission.released,
+        score: submission.score,
+    })
+
+    if (response.error) {
         /* If the update failed, revert the submission status */
```
meling, line 189: Is this accurate? There isn't any code that does the revert herein.

Also we should remove the blank line on 187.

public/src/overmind/actions.tsx
```diff
@@ -382,45 +367,60 @@ export const loadCourseSubmissions = async ({ state, actions }: Context, courseI
...
+        }
+    })
+    const groupResponse = await effects.api.client.getSubmissionsByCourse({
+        CourseID: courseID,
+        FetchMode: {
+            case: "Type",
+            value: SubmissionRequest_SubmissionType.GROUP
         }
+    })
+    if (userResponse.error || groupResponse.error) {
```
meling, line 386: What will happen if the first query worked and the second failed? Should this instead be two separate `if (xxxResponse.error)` checks, so that the second call will not be called if the first failed? I'm not sure what is the right semantics here.

public/src/overmind/actions.tsx
```diff
@@ -478,64 +490,95 @@ export const rebuildSubmission = async ({ state, actions, effects }: Context, {
...
+    const response = await effects.api.client.rebuildSubmissions({
+        courseID: state.activeCourse,
+        assignmentID: state.selectedAssignment.ID,
+        submissionID: submission.ID,
+    })
+    if (response.error) {
+        return
     }
-
+    // TODO: Alerting is temporary due to the fact that the server no longer returns the updated submission.
```
meling, line 501: Maybe there is an issue for this in the issue tracker, and if not, maybe create one (and reference it in the TODOs here).

meling (approved):
Looks good now. I added a few more comments that you may fix before merging. Also, I wonder if the `success()` and `alertHandler()` helpers can be deleted now?

public/src/overmind/actions.tsx
JosteinLindhom: Removed. If the update fails, we do not update the state at all. The submission we've mutated on line 178 is a clone and can safely be ignored.

JosteinLindhom:
> Looks good now. I added a few more comments that you may fix before merging. Also, I wonder if the `success()` and `alertHandler()` helpers can be deleted now?

The `success()` helper has been removed.

The `alertHandler` is passed to the client when we initialize Overmind. This way the client has access to an action that can mutate the state.

```ts
export const onInitializeOvermind = async ({ actions, effects }: Context) => {
    // Initialize the API client. *Must* be done before accessing the client.
    effects.api.init(actions.alertHandler)
    ...
}
```

When an error occurs, the client calls `alertHandler`. We specify how to handle any errors inside `alertHandler`, e.g. push an alert to state.

It might make more sense if this action is named `errorHandler` instead.

meling:
> When an error occurs, the client calls `alertHandler`. We specify how to handle any errors inside `alertHandler`, e.g. push an alert to state.
> 
> It might make more sense if this action is named `errorHandler` instead.

Sounds good to me. Feel free to merge this PR and the other approved PR at your leisure.

Issue #109, opened 2019-02-06
Frontend: Add a title bar to display maintenance messages


meling:
We should add a title bar on the frontend that can display messages about upcoming/planned maintenance ... This should be initiated by an administrator through a separate frontend page for entering maintenance messages. This message should not be stored in the database, but should be kept in main memory on server. 

We should also push such messages to inactive users, not currently interacting with AG. (I'm not sure how to implement this, but I assume this is possible somehow...)

When an administrator enters information about down time, the backend should "talk" to the NGIX proxy (or whichever proxy we use) to redirect users to a static page describing AG downtime. Maybe grpc-web-proxy can be used for this as well??!

PR #995, opened 2023-05-09
Auto dismiss alerts

JosteinLindhom:
This PR adds auto dismissable alerts. 

We can now create alerts that are removed after a set delay by setting the optional `delay` property an alert.

```ts
// The following alert will be dismissed after 10 seconds, or when the user interacts with it
actions.alert({ text: "Alert", color: Color.RED, delay: 10000 })
// Will be dismissed only when a user interacts with it
actions.alert({ text: "Alert", color: Color.RED })
```

Alerts that are automatically dismissed after a set delay will display an animation indicating when it will disappear.


public/src/components/alerts/Alert.tsx
```diff
@@ -0,0 +1,72 @@
...
+ *  @param alert - The alert to be displayed
+ */
+const Alert = ({ alert }: { alert: AlertType }): JSX.Element => {
+    const circleRef = useRef<SVGCircleElement>(null);
+    const actions = useActions();
+
+    useEffect(() => {
+        let id: ReturnType<typeof setTimeout>;
+        if (alert.delay) {
+            // If the optional delay property is set, the alert will be removed after the delay
```
meling, line 22: Maybe we don't need this comment here; since it is already documented in the top-level.

public/src/components/alerts/Alert.tsx
```diff
@@ -0,0 +1,72 @@
...
+                <svg viewBox="0 0 50 50" style={{ width: 20, height: 20, marginRight: 20 }}>
+                    <circle ref={circleRef} cx={25} cy={25} r={20} strokeWidth={5} fill="none" stroke="#000" />
+                </svg>
+            )}
+            {alert.text}
+        </div>
+    );
+};
+
+export default Alert;
```
meling, line 72: add line break.

meling (approved):
Minor suggestions. Please merge at your leisure.

Edit: Looked into existing issues, but couldn't find anyone quickly that is related, except maybe #109. Maybe there are some issues where this is part of the solution.

public/src/Helpers.ts
```diff
@@ -271,6 +271,10 @@ export const validateGroup = (group: CourseGroup): { valid: boolean, message: st
     return { valid: true, message: "" }
 }
 
+export const newID = (): number => {
```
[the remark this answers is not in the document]
JosteinLindhom: Rewritten to return an auto-incrementing ID instead.

meling (approved):
Minor comments. Please merge at your convenience.

meling:
@JosteinLindhom Ready to merge?

PR #1008, opened 2023-11-13
Add submission stream write deadline controller

### LINKED ISSUE #1007: Increase deadline for streaming requests
Currently *all* requests have a read/write timeout of 2 minutes. 

https://github.com/quickfeed/quickfeed/blob/c174b03ca445e4cc15a199d1a736d464ad8c53af/web/server.go#L45-L52

As of Go 1.20 we have access to the [http.ResponseController](https://pkg.go.dev/net/http#ResponseController). We can use this to set deadlines on a per-request basis. Streaming requests in particular could benefit from a longer deadline.

Hopefully `connect-go` will add functionality to set deadlines per-endpoint in the not too distant future.

JosteinLindhom:
Fixes #1007 

This PR adds a wrapper that increases the write deadline for streaming requests.


web/webserver.go
```diff
@@ -50,3 +52,17 @@ func (s *QuickFeedService) RegisterRouter(tm *auth.TokenManager, authConfig *oau
 
 	return router
 }
+
+// controller is a wrapper for the QuickFeedService handler that sets a write deadline for the submission stream.
+// TODO: Remove this when connect-go finally supports deadlines.
+// TODO: https://github.com/connectrpc/connect-go/issues/604
+func controller(h http.Handler) http.Handler {
+	timeout := 15 * time.Minute
```
meling, line 60: We should not hard code the timeout herein. It becomes slow to test. If you pass the timeout as an argument to `contoller` it can more easily be tested.

Moreover, it would be nice to have a test for this, so that Codecov can be made happy as well.

web/webserver.go
```diff
@@ -50,3 +52,17 @@ func (s *QuickFeedService) RegisterRouter(tm *auth.TokenManager, authConfig *oau
 
 	return router
 }
+
+// controller is a wrapper for the QuickFeedService handler that sets a write deadline for the submission stream.
+// TODO: Remove this when connect-go finally supports deadlines.
```
meling, line 57: Once we merge this PR, we should add an issue to remove it again once the referenced issue has been fixed in `connectrpc`.

web/webserver.go
```diff
@@ -34,10 +35,11 @@ func (s *QuickFeedService) RegisterRouter(tm *auth.TokenManager, authConfig *oau
 	router.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		http.ServeFile(w, r, public+"/assets/index.html")
 	}))
-	router.Handle(s.NewQuickFeedHandler(tm))
+	paths, handler := s.NewQuickFeedHandler(tm)
```
meling, line 38: Can we add a test for `RegisterRouter` so that we can make Codecov happy?

web/webserver.go
```diff
@@ -50,3 +52,17 @@ func (s *QuickFeedService) RegisterRouter(tm *auth.TokenManager, authConfig *oau
...
+
+// controller is a wrapper for the QuickFeedService handler that sets a write deadline for the submission stream.
+// TODO: Remove this when connect-go finally supports deadlines.
+// TODO: https://github.com/connectrpc/connect-go/issues/604
+func controller(h http.Handler) http.Handler {
+	timeout := 15 * time.Minute
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == qfconnect.QuickFeedServiceSubmissionStreamProcedure {
+			control := http.NewResponseController(w)
+			control.SetWriteDeadline(time.Now().Add(timeout))
```
meling, line 64: What should we do if this returns an error? If there is nothing to do, then we should do `_ = control.SetWriteDeadline...`

meling (changes requested):
Suggest that we complete this PR and follow up with an issue to remove this PR again once `connectrpc` has been upgraded with the relevant functionality.

PR #1009, opened 2023-11-20
Update approvelist to match new format

JosteinLindhom:
- Combine first and last name to form full name
- Column for approval is now E (from B)


cmd/approvelist/main.go
```diff
@@ -223,6 +224,17 @@ func fileName(courseCode, suffix string) string {
 }
 
 func loadApproveSheet(courseCode string) (approveMap map[string]int, sheetName string, err error) {
+	// The approve sheet is a single sheet Excel file with five columns:
+	//
+	// 		First name | Last name | Student number    | Candidate number | Approval
+	// 		-----------+-----------+-------------------+------------------+----------
+	// 		<first>    | <last>    | <student_no>      | <candidate_no>   | <approved>
+	//      John       | Doe       | 123456            |                  |
```
meling, line 232: Indentation is different on this line.

cmd/approvelist/main.go
```diff
@@ -234,8 +246,9 @@ func loadApproveSheet(courseCode string) (approveMap map[string]int, sheetName s
 	sheetName = f.GetSheetName(f.GetActiveSheetIndex())
 	approveMap = make(map[string]int)
 	for i, row := range f.GetRows(sheetName) {
-		if i > 0 && row[0] != "" {
-			approveMap[row[0]] = i + 1
+		if i > 0 && row[0] != "" && row[1] != "" {
```
meling, line 249: What happens if only one of `row[0]` and `row[1]` is non-empty?

cmd/approvelist/main.go
```diff
@@ -20,10 +20,11 @@ import (
...
 const (
-	srcSuffix = "-original.xlsx"
-	dstSuffix = "-approve-list.xlsx"
-	pass      = "Godkjent"
-	fail      = "Ikke godkjent"
+	srcSuffix      = "-original.xlsx"
+	dstSuffix      = "-approve-list.xlsx"
+	pass           = "Godkjent"
+	fail           = "Ikke godkjent"
+	approvalColumn = "E"
```
meling, line 27: Would it make sense to instead calculate this by reading the header line?

meling:
Is the reason for this change that they changed the spreadsheet format?

JosteinLindhom:
> Is the reason for this change that they changed the spreadsheet format?

Yes. The current format is as follows:

| Fornavn | Etternavn | Studentnr. | Kandidatnr. | Godkjenning |
|----------|------------|-------------|--------------|---------------|
| Luke | Skywalker | 123456 | | Godkjent |
| Anakin | Skywalker | 654321 | | Ikke godkjent |

The format last year was:

| Student | Godkjenning |
|----------|---------------|
| Luke Skywalker (123456) | Godkjent |
|Anakin Skywalker (654321) | Ikke godkjent |

It would probably make more sense to look at student IDs rather than name with the current format.

cmd/approvelist/main.go
JosteinLindhom: Alternate solution provided in 51f3e1d.

meling:
> It would probably make more sense to look at student IDs rather than name with the current format.

Yes, that sounds like a better approach. Did all students add their student IDs?

meling:
Is there more to do here? I think using student IDs is the best solution for matching.

PR #1010, opened 2023-11-22
Migrate to remote plugins

JosteinLindhom:
We previously used the "name" and "remote" keys to specify whether or not a plugin was local or remote. As of April 30, 2023 these were deprecated.

This PR simply migrates our generate files to use the new "plugin" keys.

See <https://buf.build/docs/migration-guides/migrate-remote-generation-alpha#migrate-to-remote-plugins> for more info.


meling (approved):
LGTM!

Issue #1022, opened 2024-01-05
Use organization webhook events to accept member invites


JosteinLindhom:
The [member invited](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=member_invited#organization) event could (possibly) be used to trigger logic to accept invites on behalf of users.

To recieve these events we must subscribe to `Organization` events in our GitHub app:

![image](https://github.com/quickfeed/quickfeed/assets/1346291/23960bff-ebed-44aa-8692-c60ba42303dd)

Currently QuickFeed trigger users to be invited to the organization when they are added to teams as part of group creation.

meling:
Maybe the GitHub app manifest needs to be updated to select organization.

PR #1028, opened 2024-01-10
Use webhook to listen for (and accept) organization invites

JosteinLindhom:
- docs: add important info about returned refresh token
- feat: add webhook handler for accepting organization invites (#1022)


web/hooks/organization.go
```diff
@@ -0,0 +1,47 @@
...
+func (wh GitHubWebHook) handleOrgInvite(e *github.OrganizationEvent) {
+	// check that invitee is enrolled in course
+	invitee := e.GetInvitation().GetLogin()
+	user, err := wh.db.GetUserByCourse(&qf.Course{ScmOrganizationID: uint64(e.GetOrganization().GetID())}, invitee)
+	if err != nil {
+		wh.logger.Error("Invite: Failed to find user:", err)
+		return // course does not exist, or user not enrolled
+	}
+
+	invite := e.GetInvitation()
```
meling, line 20: This check should be moved first in this method. We shouldn't print `No invite` in the log; I think we can just ignore nil invites. Does this actually happen?

web/hooks/organization.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+
+	"github.com/google/go-github/v45/github"
+	"github.com/quickfeed/quickfeed/qf"
+	"github.com/quickfeed/quickfeed/scm"
+)
+
+func (wh GitHubWebHook) handleOrgInvite(e *github.OrganizationEvent) {
+	// check that invitee is enrolled in course
+	invitee := e.GetInvitation().GetLogin()
```
meling, line 13: Should reuse the `e.GetInvitation()` below.

web/hooks/organization.go
```diff
@@ -0,0 +1,47 @@
...
+		Owner:        orgName,
+		RefreshToken: user.GetRefreshToken(),
+	})
+	if err != nil {
+		wh.logger.Error(err)
+	}
+	// The new refresh token needs to be stored in the DB
+	user.RefreshToken = newRefreshToken
+	if err := wh.db.UpdateUser(user); err != nil {
+		wh.logger.Errorf("failed to update refresh token for user (id=%d, name=%s): %w", user.ID, user.Name, err)
```
meling, line 45: `s/failed/Failed`

web/hooks/organization.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	// AcceptInvitations consumes the current refresh token and returns a new one
+	newRefreshToken, err := sc.AcceptInvitations(context.Background(), &scm.InvitationOptions{
+		Login:        user.GetLogin(),
+		Owner:        orgName,
+		RefreshToken: user.GetRefreshToken(),
+	})
+	if err != nil {
+		wh.logger.Error(err)
```
meling, line 40: Should this error lead to return? If not, we should add a comment explaining why not.

web/hooks/organization.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/quickfeed/quickfeed/qf"
+	"github.com/quickfeed/quickfeed/scm"
+)
+
+func (wh GitHubWebHook) handleOrgInvite(e *github.OrganizationEvent) {
+	// check that invitee is enrolled in course
+	invitee := e.GetInvitation().GetLogin()
+	user, err := wh.db.GetUserByCourse(&qf.Course{ScmOrganizationID: uint64(e.GetOrganization().GetID())}, invitee)
+	if err != nil {
+		wh.logger.Error("Invite: Failed to find user:", err)
```
meling, line 16: Elsewhere we log it like this:
```go
		user, err := wh.db.GetUserByCourse(course, login)
		if err != nil {
			wh.logger.Errorf("Failed to find user %s in course %s: %v", login, course.GetName(), err)
			return
		}
```
Suggest to prefix it with `Invite:` and replace `course.GetName()` with `course.GetScmOrganizationName()`.

web/hooks/organization.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/google/go-github/v45/github"
+	"github.com/quickfeed/quickfeed/qf"
+	"github.com/quickfeed/quickfeed/scm"
+)
+
+func (wh GitHubWebHook) handleOrgInvite(e *github.OrganizationEvent) {
+	// check that invitee is enrolled in course
+	invitee := e.GetInvitation().GetLogin()
+	user, err := wh.db.GetUserByCourse(&qf.Course{ScmOrganizationID: uint64(e.GetOrganization().GetID())}, invitee)
```
meling, line 14: To avoid the somewhat long line; I suggest to do:
```go
org := e.GetOrganization()
course := &qf.Course{
  ScmOrganizationID: uint64(org.GetID()),
  ScmOrganizationName: org.GetLogin(),
}
user, err := wh.db.GetUserByCourse(course, invitee)
```
Then you can reuse `course.GetScmOrganizationName()` below.

scm/scm.go
```diff
@@ -49,6 +49,9 @@ type SCM interface {
 
 	// AcceptInvitations accepts course invites on behalf of the user.
 	// A new refresh token for the user is returned, which may be used in subsequent requests.
+	//
+	// IMPORTANT: The provided refresh token will be invalidated after this call.
```
meling, line 53: Suggest to remove `Important:` here.

meling (changes requested):
Is it possible to add some tests for this?

JosteinLindhom:
> Is it possible to add some tests for this?

A test would need:

- A GitHub app
- An organization where the GitHub app is installed
- A user that has authorized the GitHub app
- A way to fetch a refresh token, likely through login of some sort

Then we could test that our token exchange works as expected.

Additionally:
- Adding the user to the organization to generate invites
- Accept the invite(s)

The test should also remove the user from the organization after running.

The hardest part is getting a refresh token. I'm not sure we can generate a refresh token from a personal access token.
A test would likely require some OAuth flow / device flow for the GitHub app to retrieve a refresh token.

We will also have to run the test with a server that is reachable by GitHub in some way. Might be possible to do so using ngrok or similar services. This is required to be able to receive webhook events.

On the other hand, the functionality introduced in this PR is not critical. Should the invite accept fail, the user will have to manually accept.

PR #1029, opened 2024-01-10
Installing GitHub App on an organization will automatically create a course

JosteinLindhom:
Installing a GitHub App on an organization will trigger QuickFeed to automatically create a course for the organization.


web/hooks/installation.go
```diff
@@ -0,0 +1,61 @@
...
+	"context"
+
+	"github.com/google/go-github/v45/github"
+	"github.com/quickfeed/quickfeed/qf"
+	"github.com/quickfeed/quickfeed/scm"
+)
+
+func (wh GitHubWebHook) handleInstallationCreated(event *github.InstallationEvent) {
+	installerID := uint64(event.GetSender().GetID())
+	user, err := wh.db.GetUserByRemoteIdentity(installerID)
```
meling, line 13: `s/user/courseCreator/`.

Should this user be checked if it is admin and allowed to create courses, ref. `interceptor/access_control.go`:
```json
	"CreateCourse":           {admin},
```
There is also this in `web/interceptor/tokens.go`:
```go
	"CreateCourse": // The signed in user gets the teacher role in the new course.
	func(ctx context.Context, tm *auth.TokenManager, _ userIDs) error {
		claims, ok := auth.ClaimsFromContext(ctx)
		if !ok {
			return fmt.Errorf("CreateCourse: missing claims in context")
		}
		return tm.Add(claims.UserID)
	},
```
Or can we trust that the event is from GitHub and representing the course creator?

meling (changes requested):
This is just a partial review; too tired to continue.

web/hooks/installation.go
JosteinLindhom: I believe we can trust that the event is coming from GitHub if we use [webhook secrets](https://docs.github.com/en/webhooks/using-webhooks/validating-webhook-deliveries).

web/hooks/installation.go
```diff
@@ -0,0 +1,90 @@
...
+	if err := wh.db.CreateCourse(courseCreator.ID, course); err != nil {
+		wh.logger.Errorf("Could not create database record for course %s: %v", orgName, err)
+	}
+
+	if err := wh.tm.Add(courseCreator.ID); err != nil {
+		wh.logger.Errorf("Could not add user %s for token refresh: %v", courseCreator.Login, err)
+	}
+}
+
+func defaultYear() uint32 {
```
meling, line 76: Add test.

web/hooks/installation.go
```diff
@@ -0,0 +1,90 @@
...
+	now := time.Now()
+	if now.Month() <= 11 && now.Day() <= 31 && now.Month() > 10 {
+		return uint32(now.Year() + 1)
+	}
+	return uint32(now.Year())
+}
+
+func defaultTag() string {
```
meling, line 84: Add test

meling (changes requested):

meling (approved):

PR #1046, opened 2024-02-07
feat: Add context logging support

### LINKED ISSUE #489: Add a logging interceptor
We do a fair bit of logging for each gRPC call, in particular errors. However, to ensure that we don't forget to log in certain places we should add a logging interceptor so that all gRPC calls and errors are logged in a single place.

See [grpc-ecosystem/go-grpc-middleware](https://github.com/grpc-ecosystem/go-grpc-middleware/tree/master/logging/zap) for reference.

### LINKED ISSUE #670: Custom logger with context
It would be nice to have the logger emit some context that belong to the job owner, especially for long running requests such as running a test scenario.

Specifically, we could add the job owner and course code as context to each log message; use separate columns for each.

This should be implemented as a custom logger in a new package `internal/logger` or reusing `log`. Our logger should be similar to the zap logger we currently use; but using fewer methods (only the ones we need), and just forward to the zap logger.

This has at least two advantages:
- It will be easier to replace zap logger with the next best thing in the future.
- The different packages such as `scm` and others will not have to depend on the zap logger, requiring only internal dependency for the logger.

meling:
This adds a simple interceptor-based context logging feature that can be used in RPC calls to fetch the context logger to log with session context such as source method, userID, courseID etc. We can add more global context if necessary, but most often we
probably want to add logging information via the actual log msg.

This version just uses the current zap logger to avoid introducing another logging framework now. We can decide later whether or not to replace zap with slog.

Fixes #670
Fixes #489

### DIFF
--- a/web/interceptor/logging.go
+++ b/web/interceptor/logging.go
@@ -1,0 +1,67 @@
+package interceptor
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"connectrpc.com/connect"
+	"github.com/quickfeed/quickfeed/database"
+	"github.com/quickfeed/quickfeed/qf"
+	"github.com/quickfeed/quickfeed/web/auth"
+	"go.uber.org/zap"
+)
+
+type ContextLoggingInterceptor struct {
+	db     database.Database
+	logger *zap.Logger
+}
+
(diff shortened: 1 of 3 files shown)


meling:
Closing. Superseded by #1566

Issue #1049, opened 2024-02-16
Improve coloring scheme


meling:
See screenshot; the red color can be a bit hard to distinguish from the green for some color blind people. Suggest to make the bar a bit wider -- that should at least help me.
![Screenshot 2024-02-16 at 11 09 50](https://github.com/quickfeed/quickfeed/assets/810999/32e5148f-7480-4e11-ace4-7d1ebb5999c5)

PR #1055, opened 2024-02-28
Add UI settings

JosteinLindhom:
- feat(state): add settings namespace to Overmind configuration
- feat(ui): set up route and button for a settings page
- feat(ui): add settings components for changing UI elements
- feat(ui): add theme provider that update CSS variables when state changes
- fix: add missing settings route

### DIFF
--- a/public/src/App.tsx
+++ b/public/src/App.tsx
@@ -9,6 +9,7 @@ import AdminPage from './pages/AdminPage'
 import Loading from './components/Loading'
 import Dashboard from './components/Dashboard'
 import AboutPage from './pages/AboutPage'
+import Settings from './components/settings/Settings'
 
 const App = (): JSX.Element => {
     const state = useAppState()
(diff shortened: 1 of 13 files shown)


meling:
Should this be linked with #1049 and taken out of draft mode?

JoachimTislov (approved):
Looks good to me, but I would recommend setting a required base width to prevent users from sabotaging themselves.

And while you're at it, you could add an option where a user can choose between dark and light mode for the whole website.

Update: I don't know if you want to resolve the 16 deepSource issues in this branch, but this one is reasonable to do `Avoid .bind() or local functions in JSX properties
JS-0417`, since catagorized as `Major`

public/src/components/settings/ColorPicker.tsx
```diff
@@ -0,0 +1,87 @@
...
+                        ))}
+                    </div>
+                </button>
+            ))}
+            <button
+                className="btn uniform-button m-2"
+                onClick={() => setCustomColors(!customColors)}
+            >
+                Custom Colors
+            </button>
```
Glattshaved, line 65: Consider removing this button and just having the colorpicker show all the time for a more intuitive setup?

Setting ``const [customColors] = React.useState(true)`` in line 17

Glattshaved (commented):
Agree, setting the bar base width to some fixed value could be helpful to avoid users removing it.

Otherwise looks good.

JosteinLindhom:
Closing in as most of the changes here will be included in an upcoming Tailwind migration

Issue #1031, opened 2024-01-11
Chore: Removed unused arguments in CreateFakeUser


meling:
We should remove the unused argument in this function:
```go
func CreateFakeUser(t *testing.T, db database.Database, _ uint64) *qf.User {
```

PR #1060, opened 2024-03-18
Replace unused args in CreateFakeUser, rename CreateNamedUser

### LINKED ISSUE #1032: Chore: Replace existing uses of CreateNamedUser with CreateFakeCustomUser
```go
func CreateNamedUser(t *testing.T, db database.Database, _ uint64, name string) *qf.User {
```
Should be replaced with:
```go
func CreateFakeCustomUser(t *testing.T, db database.Database, user *qf.User) *qf.User {
```
Which would look something like:
```diff
-	admin := qtest.CreateNamedUser(t, db, 1, "admin")
+	admin := qtest.CreateFakeCustomUser(t, db, &qf.User{Name: "admin", Login: "admin"})
```
Since `CreateNamedUser` already ignores the `ScmRemoteID`.

JosteinLindhom:
Closes #1032 and #1031

### DIFF
--- a/database/db_user_test.go
+++ b/database/db_user_test.go
@@ -73,7 +73,7 @@ func TestDBUpdateAccessTokenUserGetAccessToken(t *testing.T) {
 	)
 	db, cleanup := qtest.TestDB(t)
 	defer cleanup()
-	wantUser := qtest.CreateFakeUser(t, db, remoteID)
+	wantUser := qtest.CreateFakeUser(t, db)
 
 	cachedAccessToken := wantUser.GetRefreshToken()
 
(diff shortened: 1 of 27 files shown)


database/db_user_test.go
```diff
@@ -73,7 +73,7 @@ func TestDBUpdateAccessTokenUserGetAccessToken(t *testing.T) {
 	)
 	db, cleanup := qtest.TestDB(t)
 	defer cleanup()
-	wantUser := qtest.CreateFakeUser(t, db, remoteID)
+	wantUser := qtest.CreateFakeUser(t, db)
```
meling, line 76: can also remove the `remoteID` const, I think.

meling (approved):
Looks good to me! Minor `remoteID` const that can be removed.

PR #1069, opened 2024-04-11
Fix IsEmptyRepo

JosteinLindhom:
- test: mock implementation of RepositoryIsEmpty
- fix: return nil when no repos found
- test: add test for IsEmptyRepo
- fix: require user confirmation only if isEmptyRepo returns an error (indicating non-empty repo)

### DIFF
--- a/web/repositories_test.go
+++ b/web/repositories_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/quickfeed/quickfeed/internal/qtest"
 	"github.com/quickfeed/quickfeed/qf"
+	"github.com/quickfeed/quickfeed/web"
 	"google.golang.org/protobuf/testing/protocmp"
 )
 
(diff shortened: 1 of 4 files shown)


web/repositories_test.go
```diff
@@ -191,3 +192,120 @@ func TestGetRepositories(t *testing.T) {
...
+			request: &qf.RepositoryRequest{
+				CourseID: course.ID, // 1
+				UserID:   123,
+			},
+			// lookup for invalid user should return no repositories
+			wantErr: false,
+		},
+		{
+			name: "user has no repositories",
+			ctx:  context.Background(),
```
meling, line 280: Can avoid these in the struct, and just use `context.Background()` directly in the call.

web/repositories_test.go
```diff
@@ -191,3 +192,120 @@ func TestGetRepositories(t *testing.T) {
...
+	}
+
+	// UpdateGroup to trigger repository creation
+	_, err := client.UpdateGroup(context.Background(), qtest.RequestWithCookie(group, "cookie"))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Group now has a repository
+	// TODO: Although the repository is created, it is empty.
```
meling, line 305: Probably want to remove the TODOs

meling (changes requested):
This review might be mute by now, if you are refactoring the isEmpty methods... as we discussed. I'm submitting the things you already know about for your recollection.

public/src/overmind/utils/errors.ts
```diff
@@ -0,0 +1,30 @@
...
+export function promptOnErrorResponse<T extends AnyMessage>(response: Response<T>, errorCode: Code, message: string) {
+    if (response.error) {
+        if (response.error.code === errorCode) {
+            if (confirm(message)) {
+                return null
+            }
+        }
+    }
+    return response.error
+}
```
meling, line 30: If you add these lines in your user preferences, you should avoid these symbols:

```json
  "files.insertFinalNewline": true,
  "files.trimFinalNewlines": true,
  "files.trimTrailingWhitespace": true,
```
They are already listed in `quickfeed.code-workspace`, but they won't have effect unless you start vscode using that workspace. Alternatively, maybe we can also add these lines to `.vscode/settings.json`, but not sure...

public/src/overmind/utils/errors.ts
```diff
@@ -0,0 +1,30 @@
+import { Code } from "@bufbuild/connect"
+import { AnyMessage } from "@bufbuild/protobuf"
+import { Response } from "../../client"
+
+/** Prompt contains the messages to display to the user when prompting for confirmation. */
+export namespace Prompt {
+    export const GroupDelete = "Are you sure you want to delete this group?"
+    export const GroupRepoNotEmpty = "Warning: The group repository is not empty. Do you still want to delete the group, GitHub team, and group repository?"
+    export const EnrollmentRepoNotEmpty = "Warning: The enrollment repository is not empty. Do you still want to delete the enrollment and enrollment repository?"
```
meling, line 9: Not sure if `enrollment` is a "user-facing" term. Does this message apply to both group and user repos? If only user, we could just go with that.

PR #1078, opened 2024-04-19
fix: avoid searching for go.mod or git repo in init()

### LINKED ISSUE #1073: Avoid calling go list to find quickfeed's root folder
Calling `go list` to find quickfeed's root folder in the init function is problematic because not all (binary) tools (like `approvelist`) should live in the quickfeed folder structure. Currently, we get this:
```console
$ approvelist -h
2024/04/19 11:41:44 running: go list -m -f {{.Dir}} github.com/quickfeed/quickfeed
go: cannot match "github.com/quickfeed/quickfeed" without -versions or an explicit version: go.mod file not found in current directory or any parent directory; see 'go help modules'
2024/04/19 11:41:44 Failed to set QUICKFEED variable: exit status 1
```
when running an installed version of `approvelist` from another folder.

meling:
This moves the search for quickfeed's root folder out of the env package's init() function, since that requires that commands such as approvelist must be started from within a quickfeed folder. For situations where it is desirable to start quickfeed command itself outside the folder, requires that the QUICKFEED variable is set manually.

Fixes #1073

### DIFF
--- a/ci/parse_script_test.go
+++ b/ci/parse_script_test.go
@@ -34,7 +34,7 @@ func testRunData(qfTestOrg string) *RunData {
 }
 
 func TestLoadRunScript(t *testing.T) {
-	t.Setenv("QUICKFEED_REPOSITORY_PATH", filepath.Join(env.Root(), "testdata", "courses"))
+	t.Setenv("QUICKFEED_REPOSITORY_PATH", env.TestdataPath())
 	runData := &RunData{
 		Course: &qf.Course{
 			ID:                  1,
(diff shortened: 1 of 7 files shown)


JosteinLindhom:
If we add these two:

```go
// path.go

func PublicDir() string {
	return filepath.Join(Root(), "public")
}

func DatabasePath() string {
	return filepath.Join(Root(), "qf.db")
}
```

and use them during startup

```go
// main.go

func main() {
	var (
		dbFile   = flag.String("database.file", env.DatabasePath(), "database file")
		public   = flag.String("http.public", env.PublicDir(), "path to content to serve")
		...
      )
      flag.Parse
      ...
}
```

we should be able to run QuickFeed from outside the root dir.

JosteinLindhom:
This one has to be updated aswell, otherwise we will create `/internal/config/certs/...` dirs all over the place.

```go
func CertPath() string {
	certPath := os.Getenv("QUICKFEED_CERT_PATH")
	if certPath == "" {
                // previous
                // certPath = defaultCertPath
		certPath = filepath.Join(Root(), defaultCertPath)
	}
	return certPath
}
```

PR #1080, opened 2024-04-22
fix: show percentage contribution of a test instead of it weight

### LINKED ISSUE #1070: Display percentage impact of individual tests based on score and weight
Results for a submission currently show:

- Total score of all tests, expressed as a percentage ranging from 0-100
- Score for individual tests, expressed as number of test cases passed out of total number of test cases, e.g. 10/10
- Weight for each individual test

Test Name | Score | Weight
-- | -- | --
TestA | 1/1 | 20
TestB | 20/20 | 50
TestC | 10/20 | 50

We should consider also showing the impact each test on the score in percentages.
Score for each test is calculated as follows:

```python
(score / maxScore) * (weight / totalWeight)
# TestA
(1 / 1) * (20 / 120) -> ~16.667%
# TestB
(20 / 20) * (50 / 120) -> ~41.667%
#TestC
(10 / 20) * (50 / 120) -> ~20.833%
```

We could then display the percentage in the table in some form

Test Name | Score | Weight | % of total
-- | -- | -- | --
TestA | 1/1 | 20 | 16.7%
TestB | 20/20 | 50 | 41.7%
TestC | 10/20 | 50 | 20.8% / 41.7%

JosteinLindhom:
This PR removes the `Weight` column in lab results and replaces it with `% of total`. Each test now shows how much it contributes to the total max score in %.

resolves #1070 

TODO:
- [x] Figure out how to display the percentage to users
- [ ] Add toggle to hide passed tests ?
- [x] Show weight on hover ?

First attempt looks like this:

image

How it's currently displayed:

image


meling:
How about adding a third column; here are some options:
```text
Score  | %       | % of total
Points | %       | % of total
Score  | Score % | % of total
```
I also suggest to fix the width of these columns and lock the three score/percentage columns to the right-hand side when resizing the window.

meling (approved):

PR #1083, opened 2024-05-03
Remove GitHub teams

JosteinLindhom:


kit/score/score.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.33.0
+// 	protoc-gen-go v1.32.0
```
meling, line 3: Perhaps you should upgrade your `protoc-gen-go` plugin to avoid going backwards in version numbers.

doc/teacher.md
```diff
@@ -104,25 +104,23 @@ This approach prevents accidentally revealing commit history from old course ins
...
 
 ## Student Enrollments
 
 Students enroll in your course by logging in on QuickFeed with their GitHub accounts, find the course among the course cards and click the `Enroll` link.
 You can access the full list of students (both already enrolled and those waiting for approval) in the `Members` menu of your course page, and accept their enrollments.
 
 After a student's enrollment has been accepted into a course, the student will have access to the course's `assignment` repository and the student's personal repository, e.g., `student-labs`.
-All students in a course will be added to the `allstudents` team in the course's GitHub organization.
 
 Note that the student may receive three invitation emails from `quickfeed-uis[bot]`.
```
meling, line 115: Is this still correct that students get three invitation emails?

qf/types.proto
```diff
@@ -55,7 +55,6 @@ message Group {
     uint64 ID          = 1;
     string name        = 2 [(go.field) = { tags: 'gorm:"uniqueIndex:group"' }];
     uint64 courseID    = 3 [(go.field) = { tags: 'gorm:"uniqueIndex:group"' }];
-    uint64 ScmTeamID   = 4;
```
meling, line 58: Will this be a problem for the database? Or can we just update the database table to ignore this attribute?

web/groups.go
```diff
@@ -119,31 +117,19 @@ func (s *QuickFeedService) updateGroup(ctx context.Context, sc scm.SCM, request
 		return fmt.Errorf("failed to get %s repository for group %q: %w", course.Code, group.Name, err)
 	}
 	if repo == nil {
-		// no group repository exists; create new repository for group
-		if request.Name != "" && newGroup.ScmTeamID < 1 {
```
meling, line 123: Should we still check for `request.Name != ""` and update `newGroup.Name` as before? 

We will get here only if `repo == nil` meaning that there is no repo recorded in the database, and thus we haven't created it yet. Thus, we can still change the name according to the `request`... Or perhaps I'm missing something?

web/groups.go
```diff
@@ -119,31 +117,19 @@ func (s *QuickFeedService) updateGroup(ctx context.Context, sc scm.SCM, request
...
-			// update group name only if team not already created on SCM
-			newGroup.Name = request.Name
-		}
-		repo, team, err := createRepoAndTeam(ctx, sc, course, newGroup)
+		repo, err := createRepo(ctx, sc, course, newGroup)
 		if err != nil {
 			return err
 		}
 		s.logger.Debugf("Creating group repo in the database: %+v", repo)
```
meling, line 124: Perhaps we should instead log two events:
`Created group repo on SCM: %+v`
and
`Created group repo in database: %+v`

(meaning we log the successful events after the creation instead of saying `Creating`.)

meling (changes requested):
Mostly minor stuff

doc/teacher.md
JosteinLindhom: I don't think we need to add students to the `info` repository as it's a public repository.

web/groups.go
JosteinLindhom: I don't think it's necessary as a similar check is being done in `s.newGroup`.

meling (approved):

PR #1090, opened 2024-05-25
Refactor/simplify GetRepository (do not merge)

meling:
This tries to make the GetRepository method return a single
repository for a query since most use cases will only need a single
repository. However, this was not an easy rewrite and it remains
unclear if this will be sufficient for all cases, e.g., the is
empty test may require check for multiple repositories.
Therefore, I'm creating this as a separate branch so that I can
go back and reexamine the situation using the GetRepositories()
method instead.


meling:
Closing this because we have already removed the `GetRepositoryByRemoteID(uint64) (*pb.Repository, error)` method even though the database method still returns multiple repositories, it is now much cleaner overall.

PR #1093, opened 2024-05-26
Mock the GithubSCM implementation instead of implementing a MockSCM

### LINKED ISSUE #1087: Replace SCM mock tests
Our SCM mock tests is not testing the logic of our `GithubSCM`, but rather the `MockSCM` type and its implementation.
```go
type MockSCM struct {
	Repositories  map[uint64]*Repository
	Organizations map[uint64]*qf.Organization
	Issues        map[uint64]*Issue
	IssueComments map[uint64]string
}
```
Instead we should implement a mock that tests the logic implemented by `GithubSCM`, by wrapping/mocking the `httpClient`.
```go
func NewGithubSCMClient(logger *zap.SugaredLogger, token string) *GithubSCM {
	src := oauth2.StaticTokenSource(
		&oauth2.Token{AccessToken: token},
	)
	httpClient := oauth2.NewClient(context.Background(), src)
	return &GithubSCM{
		logger:      logger,
		client:      github.NewClient(httpClient),
		clientV4:    githubv4.NewClient(httpClient),
		token:       token,
		providerURL: "github.com",
	}
}
```
Not entirely sure how that would work, but right now, it feels kinda useless that we are essentially testing an alternative SCM implementation.

### LINKED ISSUE #1097: chore: remove unused scm.Authorization
```go
// Authorization stores information about user scopes
type Authorization struct {
	Scopes []string
}
```

### LINKED ISSUE #1100: chore: remove unused field CreateRepositoryOptions.Permission
```go
// CreateRepositoryOptions contains information on how a repository should be created.
type CreateRepositoryOptions struct {
	Organization string
	Path         string
	Private      bool
	Permission   string // Default permission level for the given repo. Can be "read", "write", "admin", "none".
}
```

meling:
This builds on #1092.

Fixes #1087 
Fixes #1097 
Fixes #1100


scm/github.go
```diff
@@ -430,8 +246,7 @@ func (s *GithubSCM) DemoteTeacherToStudent(ctx context.Context, opt *UpdateEnrol
 	if !opt.valid() {
 		return fmt.Errorf("missing fields: %+v", opt)
 	}
-	role := OrgMember
-	_, _, err := s.client.Organizations.EditOrgMembership(ctx, opt.User, opt.Organization, &github.Membership{Role: &role})
+	_, _, err := s.client.Organizations.EditOrgMembership(ctx, opt.User, opt.Organization, &github.Membership{Role: github.String(OrgMember)})
```
JosteinLindhom, line 249: We could create vars for memberships similar to the ones for push and pull access in helpers.go

meling: This is coming in #1114.

web/rebuild_test.go
```diff
@@ -59,14 +64,64 @@ func TestSimulatedRebuildWorkPoolWithErrCount(t *testing.T) {
 	}
 }
 
+// prepareGitRepo creates copies src/repo folder to dst and initializes
+// dst/repo as a git repository and adds a single file lab1/lab1.go.
+func prepareGitRepo(src, dst, repo string) error {
```
JosteinLindhom, line 69: This function is identical to the one found in `fetcher_test.go`. Maybe these could be move to the fileop package?

Issue #1098, opened 2024-06-11
chore: consider to remove scm.GetRepositories


meling:
```go
// Get repositories within organization.
GetRepositories(context.Context, *qf.Organization) ([]*Repository, error)
```
This method is only used in `GetOrganization()`, in a few tests, and in the `cmd/scm` tool, which isn't maintained.

Moreover, if we want to keep it for some reason, it is unnecessary to pass `qf.Organization` to this method since it only needs the organization name.

JosteinLindhom:
#1115 changed the method signature to take a `string` instead of `qf.Organization`, and updated the implementation accordingly. Callers are now expected to pass an organization name.

JoachimTislov:
> and in the cmd/scm tool, which isn't maintained.

I don't think so.. theres no cmd/scm folder, but the a cmd/qcm folder though

It is also used in GetOrganization:73
```go
// If getting organization for the purpose of creating a new course,
// ensure that the organization does not already contain any course repositories.
if opt.NewCourse {
	repos, err := s.GetRepositories(ctx, orgName)
	if err != nil {
		// this code path can only happen if there is an issue with accessing GitHub since
		// we already checked that the organization exists; returning the underlying error.
		return nil, err
	}
	if isDirty(repos) {
		return nil, E(op, M("%s: course repositories %s: %w", orgName, repoNames, ErrAlreadyExists))
	}
}
```

The "isDirty" is only used here btw... so if the you don't want to limit users from creating a new organization since it already exists a course then we can remove the all of GetRepositories and isDirty

JosteinLindhom:
> Update: The GetRepositories is used in the frontend: [#972](https://github.com/quickfeed/quickfeed/issues/972)

Note that `scm.GetRepositories` differs from the `GetRepositories` used in the frontend. The one used in the frontend is the RPC method.

JoachimTislov:
> > Update: The GetRepositories is used in the frontend: [#972](https://github.com/quickfeed/quickfeed/issues/972)
> 
> Note that `scm.GetRepositories` differs from the `GetRepositories` used in the frontend. The one used in the frontend is the RPC method.

Yes, I mixed those two...

meling:
We can unexport this method so it is no longer part of the SCM interface. It is now only used by the `TestMockGetRepositories` and `GithubSCM.GetOrganization`.

PR #1110, opened 2024-06-27
Rename SCM fields

### LINKED ISSUE #1094: chore: rename RepositoryOptions.Path to Repo
It is confusing that we currently use this:
```go
type RepositoryOptions struct {
	ID    uint64
	Path  string
	Owner string
}
```
But all uses of `Path` is passed to arguments of `repo string` type (for all GitHub specific methods). We should rename `Path` to `Repo`.

JosteinLindhom:
closes #1094 

also changes signature of GetRepositories as suggested in #1098, although it may be more sensible to remove it. `initDatabase` in `tasks_test.go` needs some rewriting if we remove it.


meling:
@JosteinLindhom Suggest to start over from master when we have merged #1114.

meling:
Superseded by #1115

PR #1116, opened 2024-07-02
Remove CreateCourse RPC method

### LINKED ISSUE #1112: chore: remove CreateCourse from the RPC service
We there no longer call CreateCourse from the frontend. We instead create the course as part of the installation (web hook handling the `github.InstallationEvent`) of the quickfeed app on the course's organization. Hence, we can remove the following RPC method from the `quickfeed.proto` file:

```proto
rpc CreateCourse(Course) returns (Course) {}
```

And the corresponding implementation in `quickfeed_service.go`:
```go
func (s *QuickFeedService) CreateCourse(ctx context.Context, in *connect.Request[qf.Course]) (*connect.Response[qf.Course], error) {
```

And this helper in `courses_new.go` (the whole file can be deleted):
```go
func (s *QuickFeedService) createCourse(ctx context.Context, sc scm.SCM, request *qf.Course) (*qf.Course, error) {
```

To avoid conflicts, this should not be done until we have merged pending SCM PRs.

meling:
This removes the CreateCourse RPC method and updates the tests accordingly. Specifically, we no longer need to test the course creation via the RPC interface; we can just create the course in the database using the qtest.CreateCourse test helper (there are only two exceptions that required a few minor tweaks.)

Fixes #1112


public/src/components/admin/CreateCourse.tsx
```diff
@@ -0,0 +1,45 @@
...
+        await actions.getCourses()
+        const c = state.courses.find(c => c.ScmOrganizationName === orgName)
+        if (c) {
+            await actions.getEnrollmentsByUser()
+            setCourse(c)
+        } else {
+            actions.alert({ text: "Course not found", color: Color.YELLOW, delay: 5000 })
+        }
+    }, [orgName])
+
```
meling, line 25: VSCode complains on this line with this message:
```
React Hook React.useCallback has missing dependencies: 'actions' and 'state.courses'. Either include them or remove the dependency array.eslintreact-hooks/exhaustive-deps
const orgName: string
```
The QuickFix suggest making this line be something like:
```tsx
    }, [actions, orgName, state.courses])
```
There were other QuickFix suggestions as well.

PR #1123, opened 2024-07-11
Show submissions for group assignments where user not part of a group

### LINKED ISSUE #1047: Allow teachers to run tests from their -labs repository
It looks like teachers can't create their own group and thus needs to create and join the course as a "simulated" student. This is inconvenient and unnecessary in most situations. We should allow teachers to run tests from a teacher's `-labs` repository also for group labs. 

This is important for teachers debugging their tests.

JosteinLindhom:
- fix: refreshCourseSubmissions now fetches `ALL` submissions for enrollments
- feat(ui): the result view shows submissions for group labs for users not part of a group

Resolves #1047

### DIFF
--- a/public/src/components/ComponentsHelpers.ts
+++ b/public/src/components/ComponentsHelpers.ts
@@ -38,7 +38,16 @@ const generateRow = (enrollment: Enrollment | Group, assignments: AssignmentsMap
             // we should exit early without adding to the row.
             return
         }
-        if (isGroupLab) {
+
+        if (isGroupLab && isEnrollment && !enrollment.groupID) {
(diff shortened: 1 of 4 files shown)


public/src/components/ComponentsHelpers.ts
```diff
@@ -38,7 +38,16 @@ const generateRow = (enrollment: Enrollment | Group, assignments: AssignmentsMap
             // we should exit early without adding to the row.
             return
         }
-        if (isGroupLab) {
+
+        if (isGroupLab && isEnrollment && !enrollment.groupID) {
```
meling, line 42: Shouldn't this be `... && enrollment.groupID === 0` or something?

PR #1145, opened 2024-10-09
ui: student view now show both solo and group submissions for a given assignment

JosteinLindhom:
- **feat: add SubmissionsForUser class to manage user and group submissions**
- **refactor: use new SubmissionForCourse class in state, actions**
- **feat: enhance Lab component to differentiate between group and individual submissions**
- **refactor: simplify ProgressBar component by removing assignmentIndex prop**
- **feat: replace CourseLabs component with AssignmentCard for better submission handling**
- **feat: update NavBarLabs to handle submissions for group and individual labs**
- **style: add submission-icon styles for improved visual presentation**
- **fix: correct user lookup in ManageSubmissionStatus component**
- **refactor: simplify ProgressBar tests by removing assignmentIndex prop and using SubmissionsForUser helper**
- **feat: update StudentPage to include route for group labs**


public/src/Helpers.ts
```diff
@@ -501,3 +502,98 @@ export class SubmissionsForCourse {
...
+
+        for (const submissions of this.groupSubmissions.values()) {
+            const submission = submissions.find(s => s.ID === submissionID)
+            if (submission) {
+                return submission
+            }
+        }
+    }
+
+    update(courseID: bigint, submission: Submission) {
```
meling, line 550: Does this actually update anything? Looks like `subs` should be returned or something. Probably just my lack of typescript knowledge.

public/src/Helpers.ts
```diff
@@ -501,3 +502,98 @@ export class SubmissionsForCourse {
...
+        if (submission.groupID > 0) {
+            const subs = this.groupSubmissions.get(courseID) ?? []
+            const index = subs.findIndex(s => s.ID === submission.ID)
```
meling, line 553: The logic from 553-558 is identical to 561-566. So could probably be moved out of the if-else expression; keeping only a single line in each branch.

public/src/Helpers.ts
```diff
@@ -501,3 +502,98 @@ export class SubmissionsForCourse {
...
+            if (index === -1) {
+                return
+            } else {
+                subs[index] = submission
+            }
+        }
+    }
+
+    receive(submission: Submission) {
```
meling, line 570: Maybe this could have a comment to explain that it updates the state with a new submission?

public/src/Helpers.ts
```diff
@@ -30,18 +30,19 @@ export enum ConnStatus {
 
 const months = ["January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"]
 
-/** Returns a string with a prettier format for a deadline */
-export const getFormattedTime = (timestamp: Timestamp | undefined): string => {
+/** Returns a string with a prettier format for a timestamp */
+export const getFormattedTime = (timestamp: Timestamp | undefined, offset?: boolean): string => {
```
meling, line 34: When do we use the offset and when not? Should probably add a doc comment about it.

public/src/components/navbar/NavBarLabs.tsx
```diff
@@ -30,14 +38,24 @@ const NavBarLabs = (): JSX.Element | null => {
...
-                <ProgressBar courseID={state.activeCourse.toString()} assignmentIndex={index} type={Progress.NAV} />
-            </div>
-        )
+    const labLinks = state.assignments[state.activeCourse.toString()]?.map(assignment => {
+        const submissions = state.submissions.ForAssignment(assignment)
+        if (!submissions) {
+            return null
+        }
+        return submissions.map(submission => {
+            const groupSubmission = submission.groupID !== 0n
```
meling, line 47: Lines 47-48: Possible logic in common with other parts (see below).

meling (changes requested):
Only checked code/logic.

public/src/Helpers.ts
JosteinLindhom: The `receive` method explicitly updates the array for a given key (courseID), while this method does not.

Since arrays are reference types, using `<Map>.get(courseID)` returns the *reference* to the array for the given course ID. 
Modifying `subs` will therefore modify the same array that is stored in our map.

example:

```js
let a = [1, 2, 3]
let b = a
b[0] = 20
console.log(a) // [20, 2, 3]
```

The method is not being used anywhere however. 
Might be better to remove it and rename `receive` to `update` as `receive` is used to update a submission when the submission stream sends a submission.

meling: The reason I thought it wouldn't work was that the `subs` is a `const` which shouldn't be updatable... but maybe you can still update it as a container even though it is a const variable.

But removing it and renaming `receive` is always nice for reducing maintenance overhead.

public/src/components/navbar/NavBarLabs.tsx
```diff
@@ -12,15 +12,23 @@ const NavBarLabs = (): JSX.Element | null => {
...
         return null
     }
 
-    const submissionIcon = (assignment: Assignment) => {
-        const submission = state.submissions[state.activeCourse.toString()][assignment.order - 1]
+    const submissionIcon = (assignment: Assignment, submission: Submission) => {
+        const icon: JSX.Element | null = null
+        if (assignment.isGroupLab) {
+            if (submission.groupID !== 0n) {
+                return <i className={"fa fa-users"} title={"Group assignment"} />
```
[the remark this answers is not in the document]
JosteinLindhom: Says `Group submission` now!

public/src/Helpers.ts
JosteinLindhom: The `const` means I cannot reassign the reference to some other reference, but I can modify the contents of the referenced array.

meling:
Did a quick pass and it looks good. But could you take a look at the DeepSource issues; there are a few simple fixes. You don't have to fix all.

PR #1147, opened 2024-10-25
Upgrade bufbuild/protobuf and connectrpc/connect[-web] packages

### LINKED ISSUE #1144: chore: import @bufbuild/connect instead of @bufbuild/connect-web
We are importing `@bufbuild/connect-web` a few places; this has been deprecated. We should instead import `@bufbuild/connect`.
See for example in index.d.ts:
```ts
/**
 * @deprecated please import from \@bufbuild/connect instead
 */
Code, 
```

JosteinLindhom:
- **chore: change import paths from "bufbuild/{connect/connect-web}" to "connectrpc/{connect/connect-web}". Fixes #1144**
- **chore: update buf.gen.yaml to version 2 and modify plugin configuration**
- **chore: regenerate proto files with protoc-gen-es v2.2.0**
- **refactor: update ResponseClient to be compatible with v2**
- **refactor: rewrite overmind components to be compatible with updated proto types**
- **refactor: update mock function signature to accept multiple parameters**
- **refactor: make tests compatible with updated v2 proto types**
- **refactor: update components to be compatible with v2 proto types**
- **refactor: update Helpers.ts to use new protobuf functions and types**
- **refactor: update import path for quickfeed service**


JosteinLindhom:
Need to figure out how to omit the `patch` import in `types.proto` for code generated for the frontend.

I think we managed to do this previously with `--exclude-path patch`, but that no longer seems to work.
We can manually remove the imports generated in the `.ts` files, but I'd like for us to *not* have to do that.

image

meling:
@JosteinLindhom Here is a possible solution to the problem:
```yaml
version: v2
plugins:
- remote: buf.build/bufbuild/es
  out: public/proto
  opt: target=ts
  include_imports: true
- remote: buf.build/bufbuild/connect-web
  out: public/proto
  opt: target=ts
  include_imports: true
```
(not sure if both `include_imports` are necessary...)

This will generate an unnecessary file, but at least it removes the compile errors due to the unnecessary import. The other option is to add a sed command to remove the unnecessary line:
```typescript
import { file_patch_go } from "../patch/go_pb";
```

public/src/__tests__/updateEnrollment.test.tsx
```diff
@@ -8,29 +8,47 @@ import { Route, Router } from "react-router"
...
 
 
 describe("UpdateEnrollment", () => {
     const api = new ApiClient()
     api.client = {
         ...api.client,
+        getEnrollments: mock("getEnrollments", async (request) => {
+            const enrollments: Enrollment[] = []
+            MockData.mockedEnrollments().enrollments.forEach(e => {
+                if (request.FetchMode?.case === "courseID") {
```
meling, line 26: Two questions.
1. Can this be expressed more compactly with a switch case?
2. Are there other cases than the ones covered? Because all cases does the same: `enrollments.push(e)`.

public/package.json
```diff
@@ -3,9 +3,11 @@
   "version": "0.2.0",
   "private": true,
   "dependencies": {
-    "@bufbuild/connect": "^0.8.4",
-    "@bufbuild/connect-web": "^0.8.4",
-    "@bufbuild/protobuf": "^1.2.0",
+    "@bufbuild/protobuf": "^2.2.0",
+    "@bufbuild/protoc-gen-es": "^2.2.0",
+    "@connectrpc/connect": "^2.0.0-rc.1",
```
meling, line 8: Looks like there is now a released version of these;
- https://www.npmjs.com/package/@connectrpc/connect
- https://www.npmjs.com/package/@connectrpc/connect-web

meling (changes requested):
Looks like there still needs some work to weed out a few more conversions and get rid of some compile issues (test not passing). Also two minor deepsource issues.

public/src/__tests__/setup.jest.js
```diff
@@ -0,0 +1,6 @@
+// For some reason the new proto types cause issues with Jest
+// complaining about TextEncoder and TextDecoder not available
+// on globalThis. This setup file fixes that issue for now.
```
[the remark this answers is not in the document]
JosteinLindhom: Appears to still be necessary.
I mean [globalThis](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis)

Without this setup we get this error:

```ts
Test suite failed to run

    TypeError: globalThis.TextEncoder is not a constructor

       6 | import { enumDesc, fileDesc, messageDesc } from "@bufbuild/protobuf/codegenv1";
       7 | import type { Timestamp } from "@bufbuild/protobuf/wkt";
    >  8 | import { file_google_protobuf_timestamp } from "@bufbuild/protobuf/wkt";
```

public/src/Helpers.ts
```diff
@@ -449,7 +450,7 @@ export class SubmissionsForCourse {
 
     /** ForGroup returns group submissions for the given group or enrollment */
     ForGroup(group: Group | Enrollment): Submission[] {
-        if (group instanceof Group) {
+        if (group.$typeName === "qf.Group") {
```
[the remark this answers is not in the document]
JosteinLindhom: Seems there is a `isMessage` function we can use:

```ts
if (isMessage(group, GroupSchema)) {
    ...
}
```

Note that the `$typeName` is constrained to only these values in this instance: 

```ts
$typeName: "qf.Group" | "qf.Enrollment"
```

Comparing it to any other string will give you an error, e.g:

```ts
if (group.$typeName === "hello")
// This comparison appears to be unintentional because the types '"qf.Group" | "qf.Enrollment"' and '"hello"' have no overlap.ts(2367)
```

public/src/components/Results.tsx
```diff
@@ -86,8 +87,8 @@ const Results = ({ review }: { review: boolean }) => {
             className: `${getSubmissionCellColor(submission, owner)} ${isSelected ? "selected" : ""} ${willBeReleased ? "release" : ""} ${pending ? "pending-review" : ""}`,
             onClick: () => {
                 actions.setSelectedSubmission(submission)
-                if (owner instanceof Enrollment) {
-                    actions.setActiveEnrollment(owner.clone())
+                if (owner.$typeName === "qf.Enrollment") {
```
meling, line 90: @JosteinLindhom Can this also use `isMessage(owner, EnrollmentSchema)`? If so, you should do a search for `"qf."` or `$typeName ===` and check for other places where you could use it... I didn't comment on all cases where we do string-based comparisons.

public/src/client.ts
```diff
@@ -1,72 +1,50 @@
...
 /**
  * Create a ResponseClient for the given service, invoking RPCs through the
  * given transport.
  */
-export function createResponseClient<T extends ServiceType>(
-    service: T,
-    transport: Transport,
-    errorHandler: (payload?: { method: string; error: ConnectError; } | undefined) => void
-) {
+export function createResponseClient<T extends DescService>(service: T, transport: Transport, errorHandler: (payload?: { method: string; error: ConnectError; } | undefined) => void): ResponseClient<T> {
```
meling, line 28: This was a very long line. Should we keep the formatting with one parameter per line as before?

public/src/client.ts
```diff
@@ -1,72 +1,50 @@
...
-) => Promise<Response<O>>;
-
-function createUnaryFn<I extends Message<I>, O extends Message<O>>(
-    transport: Transport,
-    service: ServiceType,
-    method: MethodInfo<I, O>,
-    errorHandler: (payload?: { method: string; error: ConnectError; } | undefined) => void
-): UnaryFn<I, O> {
+type UnaryFn<I extends DescMessage, O extends DescMessage> = (request: MessageInitShape<I>, options?: CallOptions) => Promise<Response<MessageShape<O>>>;
+export function createUnaryFn<I extends DescMessage, O extends DescMessage>(transport: Transport, method: DescMethodUnary<I, O>, errorHandler: (payload?: { method: string; error: ConnectError; } | undefined) => void): UnaryFn<I, O> {
```
meling, line 44: Long line here as well and below... Same as above.

public/src/__tests__/updateEnrollment.test.tsx
JosteinLindhom: Even better, it can be removed altogether!

We used to have to do `getCourse` and `getEnrollments` separately.
Following #1160 this is no longer the case, and only `getCourse` needs to be mocked here.

public/src/components/Results.tsx
JosteinLindhom: Yes, converted all instances of `$typeName ===` now.

Makefile
```diff
@@ -47,6 +48,9 @@ proto:
 # TODO(meling): Split the proto target to avoid generating too new typescript... Need to fix #1147 first; after which we should merge this target with the proto target.
 proto-ui: $(protopatch)
 	buf generate --template buf.gen.ui.yaml --exclude-path patch
+	@echo "Removing protopatch imports from $(proto_ts)"
+	@perl -i -ne "print unless /import\s*\{\s*file_patch_go\s*\}\s*from\s*[\"'].*patch\/go_pb[\"'];/" $(proto_ts)
```
meling, line 52: @JosteinLindhom I keep forgetting that `sed` is not a great choice for cross-platform commands because macOS sed and GNU sed is interpreting the `-i` option differently. Does WSL have `perl` installed by default?

PR #1149, opened 2024-11-07
feat(kit): methods for adding test details to score objects

JosteinLindhom:
The added methods `Errorf`, `Error` and `Fatalf` wrap `testing.T` to add messages to `score.TestDetails` while still
outputting test errors to stderr.

Currently users need to study the entire build log and identify the failed test, and failed test cases.
By using these helpers we may associate a tests' output with its score object to display it along with the test score in the frontend.

### DIFF
--- a/kit/score/score_test.go
+++ b/kit/score/score_test.go
@@ -1,6 +1,10 @@
 package score_test
 
 import (
+	"encoding/json"
+	"os"
+	"strings"
+	"sync"
 	"testing"
 
 	"github.com/quickfeed/quickfeed/kit/score"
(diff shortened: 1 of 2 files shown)


kit/score/score_test.go
```diff
@@ -32,3 +36,56 @@ func TestNormalize(t *testing.T) {
...
+
+func TestScoreDetails(t *testing.T) {
+	sc := &score.Score{}
+
+	messages := []string{"first", "second", "third"}
+	expectedMessages := messages[:2]
+
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+	// run the test in a goroutine to avoid Fatalf
```
meling, line 48: This type of testing could benefit from the test framework that Russ talks about in the testing talk… the one that takes can check exit codes and output strings etc.

kit/score/score_test.go
```diff
@@ -32,3 +37,103 @@ func TestNormalize(t *testing.T) {
...
+				if test.fatal == i {
+					sc.Fatalf(subT, m)
+				} else {
+					sc.Errorf(subT, m)
+				}
+			}
+		}()
+		wg.Wait()
+		sc.Print(t)
+		out := make([]byte, 1024)
```
meling, line 108: Add new line here

kit/score/score_test.go
```diff
@@ -32,3 +37,103 @@ func TestNormalize(t *testing.T) {
...
+		n, _ := r.Read(out)
+
```
meling, line 110: Remove new line here

kit/score/score_test.go
```diff
@@ -32,3 +37,103 @@ func TestNormalize(t *testing.T) {
...
+		if err := json.Unmarshal(out[:n], parsedScore); err != nil {
+			t.Fatalf("Failed to unmarshal score: %v", err)
+		}
+
+		gotMessages := parseTestDetails(parsedScore.TestDetails)
+		if diff := cmp.Diff(test.expected, gotMessages); diff != "" {
+			t.Errorf("TestDetails mismatch (-want +got):\n%s", diff)
+		}
+	}
+	os.Stdout = originalStdout
```
meling, line 121: Remove if deferred

kit/score/score_test.go
```diff
@@ -32,3 +37,103 @@ func TestNormalize(t *testing.T) {
...
+		sc.Print(t)
+		out := make([]byte, 1024)
+		n, _ := r.Read(out)
+
+		parsedScore := &score.Score{}
+		if err := json.Unmarshal(out[:n], parsedScore); err != nil {
+			t.Fatalf("Failed to unmarshal score: %v", err)
+		}
+
+		gotMessages := parseTestDetails(parsedScore.TestDetails)
```
meling, line 116: Can't we also check the `dir/file.go:123: ` part? One approach could be to filter the start of the path and only check the last part of it: `<filtered path>/quickfeed/kit/score/score_test.go:109`. It will require that we fix the line numbers if we update the test. It can be a hassle if we change it a lot... However, we at least check that the line numbers get extracted correctly by the `parseTestDetails` function, without having to write a separate test for it. If we do used fix line numbers, then we should have a comment saying that they must be updated if the line numbers change.

meling (changes requested):
I know this is still in draft, but thought I'd review it since I noticed it when publishing my own changes to kit.

meling:
We should revisit this in the context of recent updates related to `tests.json`. There is also some interesting new things coming in the [testing package in Go 1.25](https://tip.golang.org/doc/go1.25#testingpkgtesting).

PR #1152, opened 2024-11-15
Enrollment.TotalApproved should only count submissions for an assignment once

JosteinLindhom:


cmd/approvelist/main.go
```diff
@@ -76,8 +76,8 @@ func main() {
 		}
 		quickfeedStudents[studID] = student
 
-		submissions := courseSubmissions.For(enroll.ID)
-		numApproved := numApproved(enroll.UserID, submissions)
+		enroll.CountApprovedSubmissions(courseSubmissions.For(enroll.GetID()), false)
```
meling, line 79: Does it matter for approve list if we include the `withActivityDate` information? I think we can drop the second argument to `CountApprovedSubmissions`.

qf/enrollment.go
```diff
@@ -136,3 +138,30 @@ func (m *Enrollments) UserIDs() []uint64 {
 	}
 	return userIDs
 }
+
+func (m *Enrollment) CountApprovedSubmissions(submissions []*Submission) {
```
meling, line 142: Assuming our discussed simplification of the `LastActivityDate` can be done; I think we can rename this method to `UpdateTotalApproved` and we should add a simple doc comment saying something like:

```go
// UpdateTotalApproved updates the total approved assignments for the current enrollment 
// based on the provided submissions.
```

meling: Postponed to separate issue.

Issue #1102, opened 2024-06-23
chore: avoid accessing protobuf message type fields directly


meling:
We should ideally avoid accessing fields of protobuf messages directly; we should instead use the getter methods. Below is an example of a field accessed in two different ways in our codebase:
```go
	err := sc.DemoteTeacherToStudent(ctx, &scm.UpdateEnrollmentOptions{
		Organization: course.GetScmOrganizationName(),    // correct
		User:         user.GetLogin(),
		Status:       qf.Enrollment_STUDENT,
	})
```
```go
	scmRepo, err := sc.UpdateEnrollment(ctx, &scm.UpdateEnrollmentOptions{
		Organization: course.ScmOrganizationName,         // incorrect
		User:         user.GetLogin(),
		Status:       qf.Enrollment_STUDENT,
	})
```

adit-yaa:
If this hasn't been taken up yet by anyone, I can work on the cleanup.

meling:
> If this hasn't been taken up yet by anyone, I can work on the cleanup.

I'm not aware of anyone working on this. Feel free to submit a PR.

Ideally, you should also add tests that demonstrate that the fix avoids the potential problem due to a nil reference. For example, a `course` reference as in the `course.ScmOrganizationName` example above should be nil in the test, which **would cause a panic** due to the nil pointer deference, and fixing it should then **not cause a panic.**

adit-yaa:
Yes, will do.

JosteinLindhom:
#1142 and #1143  are somewhat related to this issue.

PR #1154, opened 2024-11-21
#1102: refactor code to avoid accessing protobuf message type fields directly and add UT

adit-yaa:
Removes all direct access to protobuf message type fields and adds UTs to test out the functions.

Resolves: [#1102](https://github.com/quickfeed/quickfeed/issues/1102)


qf/group_test.go
```diff
@@ -19,3 +19,210 @@ func TestGetUserSubset(t *testing.T) {
...
+			g2: &qf.Group{
+				Users: []*qf.User{
+					{ID: 1, Name: "user-1"},
+					{ID: 2, Name: "user-2"},
+				},
+			},
+			want: true,
+		},
+		{
+			name: "Different order same users",
```
adit-yaa, line 171: This test is also failing because of reflect.DeepEqual(). Should we handle this case also?

adit-yaa, line 171: This test is failing because of `reflect.DeepEqual()`. As this also compares if the order is same. I am not sure if we should handle this case or not?

adit-yaa: I am quite new to the project so I might need help figuring some things out.

adit-yaa: @meling @JosteinLindhom can you please suggest?

qf/group_test.go
```diff
@@ -19,3 +19,349 @@ func TestGetUserSubset(t *testing.T) {
...
+		},
+		{
+			name:  "Nil group",
+			group: nil,
+			want:  nil,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
meling, line 54: Just a quick comment about this line; since Go. 1.22, this is no longer necessary and can be removed from here and other tests below. For details see this [blog post](https://go.dev/blog/loopvar-preview).

kit/score/parse.go
```diff
@@ -49,16 +49,16 @@ func (sc *Score) isValid(secret string) error {
...
-	if sc.MaxScore <= 0 {
+	if sc.GetMaxScore() <= 0 {
 		return test.ErrMsg(tName, ErrMaxScore.Error())
 	}
-	if sc.Weight <= 0 {
+	if sc.GetWeight() <= 0 {
 		return test.ErrMsg(tName, ErrWeight.Error())
 	}
-	if sc.Score < 0 || sc.Score > sc.MaxScore {
+	if sc.GetScore() < 0 || sc.Score > sc.MaxScore {
```
meling, line 58: Missed two fields in the second part of the or.

adit-yaa: Oh I see, thanks for pointing it out.

meling:
I just recently came across [this blog preview](https://opaque-preview.stapelberg.ch/go.dev/blog/protobuf-opaque), which is related to this work. Seems like there is a tool available to automate some of this translation work. Might be worth to try it out.

meling:
Hi @adit-yaa. Just wanted to check in with you on this. Are there more outstanding changes to include? Is it ready to be reviewed?

adit-yaa:
Hi @meling, the only remaining changes here are adding cases in UT for nil values, which would usually cause panic but using these functions it should not.

meling:
First of all, thanks for working on this... Let me know when you will be able to finalize the remaining UTs for nil.

We have two students who will be working on this project over the next 4-5 months, and this PR touches many parts. Hence, it would be great to merge this before they get going on their own larger PRs.

PS: Looks like there is also one failing test:
```
--- FAIL: TestGroup_ContainsAll/Different_order_same_users (0.00s)
```
And some minor DeepSource issues that are unrelated to your changes. They get flagged because you made changes to a file that hasn't been checked by DeepSource in a while (and they've added more checks since then...) Feel free to fix those issues as well.

meling:
Closing in favor of #1193

PR #1160, opened 2024-12-13
Improve data fetching for courses

### LINKED ISSUE #1151: Enrollment `TotalApproved` should only count submissions for an assignment once
If a student has made a submission for an assignment in both individual and group repos, these will be counted twice.

https://github.com/quickfeed/quickfeed/blob/8d2194656db3511a9b8cb7d09a919939387b89cb/web/courses.go#L333-L345

Example:

Course has 5 assignments that are auto-approved.
Student pushes all 5 assignments to both group and individual repositories with passing scores.
`enrollment.TotalApproved` will then return `10`, not `5` as I would expect.

To solve this problem we should count number of approved submissions similarly to how we do it in `approvelist`:
https://github.com/quickfeed/quickfeed/blob/8d2194656db3511a9b8cb7d09a919939387b89cb/cmd/approvelist/main.go#L176-L190

JosteinLindhom:
- **fix: enrollment `TotalApproved` should only count submissions for an assignment once Fixes #1151**
- **feat: add CountApprovedSubmissions method to track unique approved submissions count**
- **refactor: remove unused numApproved function**
- **refactor: move TestCountApproved from web package to qf package**
- **refactor: remove withActivityDate parameter from CountApprovedSubmissions method at it is unnecessary**
- **feat: add isTeacher and courseStatus helpers for user role and status checks**
- **feat: add GetCourseByStatus method to fetch course with preloaded data based on enrollment status**
- **fix: GetCourse now simply returns the course with no preloaded data**
- **refactor: remove unnecessary parameter from GetCourse calls**
- **refactor: simplify getEnrollmentsWithActivity method by removing unnecessary logic and using GetCourseByStatus**
- **refactor: remove unnecessary isTeacher parameter from getEnrollmentsWithActivity method**
- **refactor: remove getEnrollmentsByCourse method and streamline GetEnrollments logic**
- **refactor: remove unnecessary preload parameter from GetCourse calls in multiple methods**
- **refactor: update GetCourse method to use GetCourseByStatus for improved course retrieval**
- **refactor: userGroup is now derived from enrollments**
- **refactor: replace getEnrollmentsByCourse with getCourseData for improved course data retrieval**
- **refactor: simplified `fetchUserData` such that it does not preemptively fetch *all* data**
- **refactor: update Groups component to use users instead of enrollments for member display**
- **feat: preload group users in getEnrollments.**
- **fix(tests): update tests to use getCourseData instead of the removed getEnrollmentsByCourse**


web/quickfeed_service.go
```diff
@@ -102,8 +102,9 @@ func (s *QuickFeedService) UpdateCourse(ctx context.Context, in *connect.Request
 }
 
 // GetCourse returns course information for the given course.
-func (s *QuickFeedService) GetCourse(_ context.Context, in *connect.Request[qf.CourseRequest]) (*connect.Response[qf.Course], error) {
-	course, err := s.db.GetCourse(in.Msg.GetCourseID(), false)
+func (s *QuickFeedService) GetCourse(ctx context.Context, in *connect.Request[qf.CourseRequest]) (*connect.Response[qf.Course], error) {
+	status := courseStatus(ctx, in.Msg.GetCourseID())
```
JoachimTislov, line 106: status initializing is unnecessary

Can write it directly: 

```
course, err := s.db.GetCourseByStatus(in.Msg.GetCourseID(), courseStatus(ctx, in.Msg.GetCourseID()))
```

web/quickfeed_service.go
```diff
@@ -198,7 +199,12 @@ func (s *QuickFeedService) GetEnrollments(_ context.Context, in *connect.Request
...
-		enrollments, err = s.getEnrollmentsByCourse(in.Msg)
+		courseID := in.Msg.GetCourseID()
+		if isTeacher(ctx, courseID) {
+			enrollments, err = s.getEnrollmentsWithActivity(courseID)
+		} else {
+			enrollments, err = s.db.GetEnrollmentsByCourse(courseID, in.Msg.GetStatuses()...)
+		}
 		if err != nil {
 			s.logger.Errorf("GetEnrollments failed: course %d: %v", in.Msg.GetCourseID(), err)
 			return nil, connect.NewError(connect.CodeNotFound, errors.New("failed to get enrollments for course"))
```
JoachimTislov, line 210: courseID := in.Msg.GetCourseID()

Is not used inside the error handling
```
s.logger.Errorf("GetEnrollments failed: course %d: %v", in.Msg.GetCourseID(), err)
```

web/courses.go
```diff
@@ -323,22 +314,14 @@ func (s *QuickFeedService) getEnrollmentsWithActivity(courseID uint64) ([]*qf.En
...
-		enrollmentsWithActivity = append(enrollmentsWithActivity, enrollment)
-	}
-	pending, err := s.db.GetEnrollmentsByCourse(courseID, qf.Enrollment_PENDING)
-	if err != nil {
-		return nil, err
 	}
-	// append pending users
-	enrollmentsWithActivity = append(enrollmentsWithActivity, pending...)
-	return enrollmentsWithActivity, nil
+	return course.GetEnrollments(), nil
```
JoachimTislov, line 324: Update the getEnrollmentsWithActivity to getEnrollmentsWithActivityByCourse, minor change since enrollments are retrieved by course ID

web/quickfeed_service.go
```diff
@@ -187,7 +188,7 @@ func (s *QuickFeedService) UpdateEnrollments(ctx context.Context, in *connect.Re
 }
 
 // GetEnrollments returns all enrollments for the given course ID or user ID and enrollment status.
-func (s *QuickFeedService) GetEnrollments(_ context.Context, in *connect.Request[qf.EnrollmentRequest]) (*connect.Response[qf.Enrollments], error) {
+func (s *QuickFeedService) GetEnrollments(ctx context.Context, in *connect.Request[qf.EnrollmentRequest]) (*connect.Response[qf.Enrollments], error) {
```
JoachimTislov, line 191: Consider intializing following variables to reduce duplicate code:

```
msg := in.Msg
fetchMode := msg.GetFetchMode()
statuses := msg.GetStatuses()
```

database/gormdb_course.go
```diff
@@ -53,30 +54,50 @@ func (db *GormDB) CreateCourse(courseCreatorID uint64, course *qf.Course) error
...
 	var course qf.Course
+	if err := db.conn.First(&course, courseID).Error; err != nil {
+		return nil, err
+	}
+	return &course, nil
+}
 
-	if withEnrollments {
-		// we only want submission from users enrolled in the course
+func (db *GormDB) GetCourseByStatus(courseID uint64, status qf.Enrollment_UserStatus) (*qf.Course, error) {
```
JoachimTislov, line 66: Codecov wants tests for the new method - GetCourseByStatus

public/src/overmind/actions.tsx
```diff
@@ -135,18 +135,15 @@ export const updateAdmin = async ({ state, effects }: Context, user: User): Prom
...
-            case: "courseID",
-            value: value.courseID,
-        },
-        statuses: value.statuses,
+export const getCourseData = async ({ state, effects }: Context, { courseID }: { courseID: bigint }): Promise<void> => {
+    const response = await effects.api.client.getCourse({
+        courseID,
     })
     if (response.error) {
         return
```
JoachimTislov, line 143: Consider alerting user about the error ?

database/gormdb_course.go
JoachimTislov, line 66: Consider splitting up the Method to make it more readable, for example one for handling the status (switch statement) and one for handling the case where the status is equal to "enrollment teacher".

Why not include this logic in the switch statement:
```
if status == qf.Enrollment_TEACHER {
		// Set number of remaining slip days for each course enrollment
		for _, e := range course.Enrollments {
			e.SetSlipDays(&course)
		}
		for _, g := range course.Groups {
			// Set number of remaining slip days for each group enrollment
			for _, e := range g.Enrollments {
				e.SetSlipDays(&course)
			}
		}
	}
```

public/src/overmind/actions.tsx
```diff
@@ -329,7 +326,7 @@ export const approvePendingEnrollments = async ({ state, actions, effects }: Con
     const response = await effects.api.client.updateEnrollments({ enrollments })
     if (response.error) {
         // Fetch enrollments again if update failed in case the user was able to approve some enrollments
-        await actions.getEnrollmentsByCourse({ courseID: state.activeCourse, statuses: [Enrollment_UserStatus.PENDING] })
+        await actions.getCourseData({ courseID: state.activeCourse })
         return
```
JoachimTislov, line 330: I don't know if alerting the user about errors is common, but is the error handled in the backend, so the user doesn't need to know about it?

JoachimTislov (changes requested):
I think everything looks good apart from what I commented on, and just add tests for what codecov complains about, fix Jest Test as well as Deep source and it should be good. This closes #1165 when merged into main

database/gormdb_course.go
```diff
@@ -53,30 +54,50 @@ func (db *GormDB) CreateCourse(courseCreatorID uint64, course *qf.Course) error
...
+	m := db.conn
+	var course qf.Course
+	switch status {
```
meling, line 69: This method needs tests that cover the different status inputs. See the coverage complaints below.

database/gormdb_course.go
meling: I don't think logic above can be moved into the switch statement since it depends on the results from this query:
```go
if err := m.First(&course, courseID).Error; err != nil {
```

web/courses_test.go
```diff
@@ -171,7 +173,11 @@ func TestEnrollmentProcess(t *testing.T) {
 			}
 		}
 	}
-	if diff := cmp.Diff(pendingUserEnrollment, pendingCourseEnrollment, protocmp.Transform()); diff != "" {
+	if diff := cmp.Diff(pendingUserEnrollment, pendingCourseEnrollment, cmp.Options{
```
meling, line 176: You don't need the `cmp.Options{}` thing; this works fine too since `Diff` takes a variadic argument...
```go
	if diff := cmp.Diff(pendingUserEnrollment, pendingCourseEnrollment,
		protocmp.Transform(),
		protocmp.IgnoreFields(&qf.Enrollment{}, "course"),
	); diff != "" {
```

web/quickfeed_service.go
meling: I find it more clear to create a `status` variable first.

web/quickfeed_service.go
meling: Agree on this.

web/quickfeed_service.go
```diff
@@ -198,7 +199,12 @@ func (s *QuickFeedService) GetEnrollments(_ context.Context, in *connect.Request
 			return nil, connect.NewError(connect.CodeNotFound, errors.New("no enrollments found for user"))
 		}
 	case *qf.EnrollmentRequest_CourseID:
-		enrollments, err = s.getEnrollmentsByCourse(in.Msg)
+		courseID := in.Msg.GetCourseID()
```
meling, line 202: We should do `userID := in.Msg.GetUserID()` as well, but on Line 196.

meling (changes requested):
Some more nitpicks and stuff

database/gormdb_course.go
JosteinLindhom: We can create a method on `Course` that populates enrollment slipdays, e.g.:

```go
// PopulateSlipDays populates the slip days for all enrollments in the course.
func (course *Course) PopulateSlipDays() {
	for _, enrollment := range course.GetEnrollments() {
		enrollment.SetSlipDays(course)
	}
	for _, group := range course.GetGroups() {
		for _, enrollment := range group.GetEnrollments() {
			enrollment.SetSlipDays(course)
		}
	}
}
```

Which we can use:

```go
	...
	if status == qf.Enrollment_TEACHER {
		course.PopulateSlipDays()
	}
	return &course, nil
```

meling (approved):

PR #1162, opened 2024-12-26
Updated the parameter in delete repository to only require ID instead of the whole opt object

JoachimTislov:
The entries in the TestMockDeleteGroup can be reduced since the method delete group does not require opt anymore and therefore is only used when confirming the deletion of a group in the test itself

### DIFF
--- a/scm/github.go
+++ b/scm/github.go
@@ -239,7 +239,7 @@ func (s *GithubSCM) RejectEnrollment(ctx context.Context, opt *RejectEnrollmentO
 	if _, err := s.client.Organizations.RemoveMember(ctx, org.ScmOrganizationName, opt.User); err != nil {
 		return E(op, m, fmt.Errorf("failed to remove user: %w", err))
 	}
-	if err := s.deleteRepository(ctx, &RepositoryOptions{ID: opt.RepositoryID}); err != nil {
+	if err := s.deleteRepository(ctx, opt.RepositoryID); err != nil {
 		return E(op, m, err)
 	}
 	return nil
@@ -321,11 +321,11 @@ func (s *GithubSCM) UpdateGroupMembers(ctx context.Context, opt *GroupOptions) e
 }
 
 // DeleteGroup deletes a group's repository.
-func (s *GithubSCM) DeleteGroup(ctx context.Context, opt *RepositoryOptions) error {
+func (s *GithubSCM) DeleteGroup(ctx context.Context, id uint64) error {
 	const op Op = "DeleteGroup"
 
 	// options will be checked in deleteRepository
-	if err := s.deleteRepository(ctx, opt); err != nil {
+	if err := s.deleteRepository(ctx, id); err != nil {
 		return E(op, M("failed to delete group repository"), err)
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


scm/github.go
```diff
@@ -388,26 +388,27 @@ func (s *GithubSCM) createRepository(ctx context.Context, opt *CreateRepositoryO
...
-	if !opt.valid() {
-		return E(op, m, fmt.Errorf("missing fields: %+v", *opt))
+	if id == 0 {
+		return E(op, m, fmt.Errorf("missing ID"))
 	}
 
 	// if ID provided, get path and owner from github
-	if opt.ID > 0 {
-		repo, _, err := s.client.Repositories.GetByID(ctx, int64(opt.ID))
+	if id > 0 {
```
JosteinLindhom, line 399: Not sure we need to check that the `id` is greater than 0 here anymore with your changes. The case where the `id` is *not* greater than zero is handled above, therefore it is guaranteed to be greater than zero here.

scm/github.go
```diff
@@ -388,26 +388,27 @@ func (s *GithubSCM) createRepository(ctx context.Context, opt *CreateRepositoryO
...
+		repo, _, err := s.client.Repositories.GetByID(ctx, int64(id))
 		if err != nil {
-			return E(op, m, fmt.Errorf("failed to get repository %d: %w", opt.ID, err))
+			return E(op, m, fmt.Errorf("failed to get repository %d: %w", id, err))
 		}
-		opt.Repo = repo.GetName()
-		opt.Owner = repo.Owner.GetLogin()
-	}
+		var repo_name = repo.GetName()
+		var repo_owner = repo.Owner.GetLogin()
```
JosteinLindhom, line 405: We should avoid using underscores in names and use [`MixedCaps`](https://go.dev/doc/effective_go#mixed-caps) instead:

```go
// would be
repoName := repo.GetName()
// alternative
name := repo.GetName()
```

For variables that have explicit values we tend to use shorthand declarations, e.g. `owner := repo.Owner.GetLogin()`

We could also skip this step altogether and use the variables directly in `.Delete(...)`

```go
if _, err := s.client.Repositories.Delete(ctx, repo.GetName(), repo.Owner.GetLogin())
```

meling: Yes. The last alternative without using variables is my preference here; there is almost no cost to calling the Getter-methods within the `Delete` call, and calling them again if an error occurs should not happen often.

However, to be sure that that this code doesn't cause a panic, due to a nil pointer exception, we should use `repo.GetOwner().GetLogin()`:
```go
if _, err := s.client.Repositories.Delete(ctx, repo.GetName(), repo.GetOwner().GetLogin())
```
Using `GetOwner` and `GetLogin` is safe against calls on a `nil` object.

JoachimTislov:
> @JoachimTislov Ping! This requires minor fixes to get merged.

Yes, I forgot about this pull request, will fix it now

scm/github.go
JoachimTislov: Right

meling (approved):
Looks good to me!

meling:
@JoachimTislov I suspect that the codecov issue could be related to where you're creating the PR from (your own fork of quickfeed). But I'm not sure...

JoachimTislov:
> @JoachimTislov I suspect that the codecov issue could be related to where you're creating the PR from (your own fork of quickfeed). But I'm not sure...

Yes, you have to configure the secret key for code coverage on github, and grant the fork access to the token, but theres no point in doing that now since I won't be using the fork anymore

PR #1164, opened 2024-12-26
chore: Added testing for the parsing of the criteria.json file, and checks for if the course Id is set correctly

### LINKED ISSUE #729: Add test for criteria.json parsing
The `assignments.readTestsRepositoryContent` function parses the `criteria.json` file. However, this parsing is currently not tested. We should add a `lab4/criteria.json` and corresponding `lab4/assignment.yml` file to `assignments/testdata/tests` and add a test that the criteria file gets parsed correctly. Moreover, we should also check that the `courseID` gets initialized correctly based on this code:
```go
for _, bm := range benchmarks {
	bm.CourseID = courseID
	for _, c := range bm.Criteria {
		c.CourseID = courseID
	}
}
```

JoachimTislov:
* Resolves #729 

I also created a new courseID variable to reduce hard-coded IDs in the initialization of "wantAssignments"


assignments/testdata/tests/lab4/criteria.json
```diff
@@ -0,0 +1,10 @@
+[
+    {
+        "CourseID": 0,
```
JosteinLindhom, line 3: To best match how we use the `criteria.json` file "in the wild" you could omit the `CourseID` here (and below on line 6).

Example of how a teacher would create `criteria.json` (no course ID):

```json
[
    {
        "heading": "Assignment 1",
        "criteria": [
            {
                "description": "Links work"
            },
            {
                "description": "Images are links, opening in a new tab"
            }
        ]
    }
]
```

The minimal `criteria.json` file we could use in this test is

```json
[
    {
        "criteria": [
            {}
        ]
    }
]
```

meling: @JoachimTislov I suggest to add more `criteria.json` example files to test different legal versions. You may need to add more `labX` folders to `testdata/tests/`... since you cannot have multiple `criteria.json` files per lab.

Also, you may consider adding an invalid `criteria.json` file to check that bad files give an expected parse error, but I think it is enough to provide one example.

assignments/testdata/tests/lab4/criteria.json
```diff
@@ -0,0 +1,10 @@
...
+        "criteria": [
+            {
+                "CourseID": 0
+            }
+        ]
+    }
+]
```
meling, line 10: To avoid churn in commits, please use a newline at the end of files. Ask @JosteinLindhom how to fix this in VSCode. I think if you start VSCode with:
```sh
code quickfeed.code-workspace
```
It should be configured correctly, but not completely sure.

meling (changes requested):
See comments inline.

assignments/testdata/tests/lab4/criteria.json
JoachimTislov: Yes, the quickfeed code workspace adds a newline 👍

meling:
@JoachimTislov Ping! This requires minor fixes to get merged.

JoachimTislov:
> @JoachimTislov Ping! This requires minor fixes to get merged.

Yes, I did the requested changes, PR currently fails since we reached the rate limit for Codecov

assignments/walk_tests_repo_test.go
```diff
@@ -93,3 +164,8 @@ WORKDIR /quickfeed
 		t.Errorf("readTestsRepositoryContent() mismatch (-wantAssignments +gotAssignments):\n%s", diff)
 	}
 }
+
+// Check if the error is related to invalid JSON unmarshalling.
+func IsUnmarshalError(e error) bool {
```
meling, line 169: It should not be an exported function.

assignments/walk_tests_repo_test.go
```diff
@@ -80,10 +88,73 @@ WORKDIR /quickfeed
...
+			ScoreLimit:        80,
+			Deadline:          qtest.Timestamp(t, "2023-07-21T16:00:00"),
+			IsGroupLab:        false,
+			GradingBenchmarks: []*qf.GradingBenchmark{},
+		},
 	}
 
 	gotAssignments, gotDockerfile, err := readTestsRepositoryContent(testsFolder, 1)
 	if err != nil {
+		if IsUnmarshalError(err) {
```
meling, line 155: When will this happen? If this can happen, is it okay to not fail the test?

If there are `criteria.json` files that are expected to fail, then they should be tested separately (maybe in another test func) and checked that the error is the expected error.

meling (changes requested):

assignments/walk_tests_repo_test.go
JoachimTislov: I have to admit that this was a very cheap solution, considering the fact that there isn't an easy way to write proper testing for cases of an un-marshal error, because the "readTestsRepositoryContent" method only returns an error when it happens, no information about which lab it was. Therefore I added a check which allows for this to happen, specifically for lab 6 and 7, which is kind of pointless since you aren't really confirming the error...

I will try to find a proper way to test this.

assignments/walk_tests_repo_test.go
JoachimTislov: Oh yeah, forgot about the uppercase part in Go

assignments/walk_tests_repo_test.go
JoachimTislov: Nvm it only happes for lab 6 since the function returns the error and doesn't get to lab 7

assignments/walk_tests_repo_test.go
```diff
@@ -93,3 +139,28 @@ WORKDIR /quickfeed
...
+const negativeIntegerFolder = "testdata/invalidJsonTests/negativeInteger"
+
+func TestReadTestsRepositoryContentForInvalidCriteriaFiles(t *testing.T) {
+	checkLabWithInvalidCriteriaFile(t, invalidTypesFolder)
+	checkLabWithInvalidCriteriaFile(t, negativeIntegerFolder)
+}
+
+func checkLabWithInvalidCriteriaFile(t *testing.T, folder string) {
+	_, _, err := readTestsRepositoryContent(folder, 1)
+
```
meling, line 153: Nit: No need for blank line here

assignments/walk_tests_repo_test.go
```diff
@@ -93,3 +139,28 @@ WORKDIR /quickfeed
...
+	if err == nil {
+		t.Errorf("expected error")
+	}
+
```
meling, line 157: Nit: No need for blank line here

assignments/walk_tests_repo_test.go
```diff
@@ -93,3 +139,28 @@ WORKDIR /quickfeed
 		t.Errorf("readTestsRepositoryContent() mismatch (-wantAssignments +gotAssignments):\n%s", diff)
 	}
 }
+
+const invalidTypesFolder = "testdata/invalidJsonTests/invalidTypes"
+const negativeIntegerFolder = "testdata/invalidJsonTests/negativeInteger"
+
+func TestReadTestsRepositoryContentForInvalidCriteriaFiles(t *testing.T) {
+	checkLabWithInvalidCriteriaFile(t, invalidTypesFolder)
```
meling, line 147: This should be written as a subtest:

```go
tests := []struct{
  name string
  folder string
}{
    {name: "NameOfSubTest", folder: inavlidTypesFolder},
   ...
}
for _, tc := range tests {
  t.Run(tc.name, func(t *testing.T) {
...
  })
}
```

assignments/walk_tests_repo_test.go
```diff
@@ -93,3 +139,28 @@ WORKDIR /quickfeed
 		t.Errorf("readTestsRepositoryContent() mismatch (-wantAssignments +gotAssignments):\n%s", diff)
 	}
 }
+
+const invalidTypesFolder = "testdata/invalidJsonTests/invalidTypes"
```
meling, line 143: Should be define inside the test function where they apply unless they are needed in the global package name space.

meling (changes requested):

meling:
@JoachimTislov Ping! I requested a few more minor changes to this one.

JoachimTislov:
> @JoachimTislov Ping! I requested a few more minor changes to this one.

@meling Yes, i've pushed the requested changes

meling (approved):

PR #1168, opened 2025-01-05
fix: check that a submission is a group submission before fetching group repo Fixes #1167

### LINKED ISSUE #1167: bug: rebuilding individual submission for group submission does nothing
This code from `rebuild.go`

```go
if assignment.IsGroupLab {
	repo, err = s.getRepo(course, submission.GetGroupID(), qf.Repository_GROUP)
	s.logger.Debugf("Rebuilding submission %d for group(%d): %s, assignment: %+v, repo: %s",
		submission.GetID(), submission.GetGroupID(), name, assignment, repo.GetHTMLURL())
} else {
	repo, err = s.getRepo(course, submission.GetUserID(), qf.Repository_USER)
	s.logger.Debugf("Rebuilding submission %d for user(%d): %s, assignment: %+v, repo: %s",
		submission.GetID(), submission.GetUserID(), name, assignment, repo.GetHTMLURL())
}
```

treats all submissions for group assignments as if they are submitted by a group.

We allow students to submit code for group assignments to their individual repositories.
The code above will end up sending group ID `0` into the `getRepo` method in these cases.

JosteinLindhom:


meling (approved):
This looks good to me. 

I checked and couldn't find any other code (on master) that do something like:
```go
assignment.IsGroupLab && submission.GetGroupID() > 0
```
But maybe this pattern is used in one of the other PRs related to running group assignments from user repos. In which case, perhaps it is worth adding a helper function for this.

Issue #1174, opened 2025-01-07
chore: Backend, warning and errors


JoachimTislov:
- Package warning

  'No packages found for open file /home/yoa/quickfeed/*nameOfPackage(namespace)*.go.
  This file may be excluded due to its build tags; try adding `-tags=<build tag>` to your gopls "buildFlags" configuration
  See the documentation for more information on working with build tags:
  <https://github.com/golang/tools/blob/master/gopls/doc/settings.md#buildflags>.'
  
  In following files:
  
  - tools
  - kit/sh
    - cmd_race_test
    - race_test

- ci.local - compiler error

  'undefined: ci.Local [aix,ppc64]'
  
  In following files:
  
  - web
    - quickfeed_mock_client
    - rebuild_test
  - cmd/qcm/main.go
  - ci
    - local_unix_test
    - run_tests_clone_test

meling:
>   In following files:
>   
>   * tools

This will be fixed when Go 1.24 is released with support for specifying tools in `go.mod`

>   * kit/sh
>     * cmd_race_test
>     * race_test

If you start your VSCode from the command line like this:
```sh
code quickfeed.code-workspace
```
(or maybe you can double-click on this file...)

This is because this workspace file defines the following:
```json
    "buildFlags": ["-tags=race"]
```

We could consider moving back to saving settings in `.vscode/settings.json` now that gopls and the Go plugin supports the `go.work` file. That way, we would avoid forgetting to start VSCode with the workspace file.

> * ci.local - compiler error
>   'undefined: ci.Local [aix,ppc64]'
>   In following files:
>   
>   * web
>     
>     * quickfeed_mock_client
>     * rebuild_test
>   * cmd/qcm/main.go
>   * ci
>     
>     * local_unix_test
>     * run_tests_clone_test

This I have not seen before, but it surely is related to something we have messed up in the way we specify [build constraints](https://pkg.go.dev/cmd/go#hdr-Build_constraints) and file naming of these files:
```sh
tree ci
ci
├── local_unix.go
├── local_unix_test.go
├── local_windows.go
├── local_windows_test.go
```
Are you running on `aix,ppc64`??

JoachimTislov:
> > In following files:
> > 
> > * tools
> 
> This will be fixed when Go 1.24 is released with support for specifying tools in `go.mod`
> 
> > * kit/sh
> >   
> >   * cmd_race_test
> >   * race_test
> 
> If you start your VSCode from the command line like this:
> 
[...]

PR #1181, opened 2025-01-10
tests(ci): fix issue where test would not compile on Windows machines

JosteinLindhom:


JoachimTislov (approved):
Looks good, the compile error is gone

Closes: #1174, since the other warnings and errors have been resolved

meling:
Here is an even better idea. The tests `TestLocal` and `TestLocalWindows` should only have one test: `TestLocal` (they are testing the same thing.) We should move `TestLocal` to a new file, `local_test.go`.
This will run the same test whether or not you are on Unix or Windows. The difference is only the implementation files `local_unix.go` and `local_windows.go`.

Then, rename the `local_unix_test.go` and `local_windows_test.go` to `docker_unix_test.go` and `docker_windows_test.go`, leaving only the `checkOwner` function in the two files (and the build constraints). This is because it's the `docker_test.go` file that requires the `checkOwner` function, which is unrelated to the test in `local_*_test.go`

meling (approved):

PR #1192, opened 2025-01-23
Update database anonymizer script

JosteinLindhom:
The `anonymize` script is no longer compatible with our current database schema.
The changes in this PR updates the SQL queries to match our current schema.

The script now also randomizes the submission commit hashes and user refresh tokens.


meling (approved):
Looks good to me!

Issue #360, opened 2020-10-31
Create/set up docker container for development


meling:
From the `Remote - Containers` plugin in VSCode:

Whether you deploy to containers or not, containers make a great development environment because you can:

- Develop with a consistent, easily reproducible toolchain on the same operating system you deploy to.
- Quickly swap between different, isolated development environments and safely make updates without worrying about impacting your local machine.
- Make it easy for new team members / contributors to get up and running in a consistent development environment.
- Try out new technologies or clone a copy of a code base without impacting your local setup.

We should create such an environment so that we can more easily test new features...

meling:
Maybe we can use the [ngrok](https://ngrok.com) tool to simplify the oauth2 config.

JoachimTislov:
What is the status on this issue?

JosteinLindhom:
> What is the status on this issue?

We had a go at creating a docker container for development way back with moderate success.

Since then we've made it easier to set up the system, likely making it easier to create a useful docker container for dev.
(status = not done)

I remember looking at [this](https://github.com/air-verse/air) when working on this previously.

meling:
It would be great to get something like this working for us.

PR #1194, opened 2025-01-27
Docker container for development

JoachimTislov:
Closes: #360 

The docker image is created with air, and containers will be created with a volume

The changes in PR #1201 should work inside of the docker container, by simply adding the watch flag in the args array inside of the .air.toml configurations file.

Also removed localhost constraint, and added a binary launch profile which allows non root user to set capabilities for the binary and debug with it instead of running main.go directly with port issues.

Stumbled upon a [deepsource - CRT-D0011 issue](https://app.deepsource.com/gh/quickfeed/quickfeed/run/2a6bd676-de4f-4552-ac4c-062660995958/go/CRT-D0011?listcategory&listsort&listq&listindex=0), not sure why it flagged and disappeared later


web/auth/jwt.go
```diff
@@ -36,10 +36,17 @@ type TokenManager struct {
 }
 
 // NewTokenManager starts a new token manager. Will create a list with all tokens that need update.
-func NewTokenManager(db database.Database) (*TokenManager, error) {
+func NewTokenManager(db database.Database, dev *bool) (*TokenManager, error) {
+	secret := rand.String()
+	// Use a fixed secret for development.
+	// New secret is generated on each restart.
+	// Not sure if this is the best way to handle this.
```
JosteinLindhom, line 43: We could add a new function in the `env` package that returns `rand.String()` if no `QUICKFEED_AUTH_SECRET` (or some other name) is set.

Otherwise we can use the user defined secret.

This way we can avoid having users be logged out when the server restarts.
Changing the secret *and* restarting will invalidate all JWTs instead.

```go
// somewhere in the env package
func AuthSecret() string {
	authSecret := os.Getenv("QUICKFEED_AUTH_SECRET")
	if authSecret == "" {
		return rand.String()
	}
	return authSecret
}
```

Then you don't have to use `dev` as an argument in `NewTokenManager` either:

```go
func NewTokenManager(db database.Database) (*TokenManager, error) {
	manager := &TokenManager{
		db:     db,
		secret: rand.String(),
		secret: env.AuthSecret(),
	}
}
```

web/auth/jwt.go
```diff
@@ -36,10 +36,17 @@ type TokenManager struct {
 }
 
 // NewTokenManager starts a new token manager. Will create a list with all tokens that need update.
-func NewTokenManager(db database.Database) (*TokenManager, error) {
+func NewTokenManager(db database.Database, dev *bool) (*TokenManager, error) {
```
JosteinLindhom, line 39: `dev *bool` should be `dev bool`. This way you can avoid passing `nil` elsewhere:

```go
// web/users_test.go
// with dev *bool
tm, err := auth.NewTokenManager(db, nil)
// with dev bool
tm, err := auth.NewTokenManager(db, false)
```

In `main.go` you would instead send the value of `dev`, similar to how you do for `RegisterRouter`.

```go
tokenManager, err := auth.NewTokenManager(db, *dev)
```

However, if the comment below makes sense to use then the bool is not needed at all.

public/src/Helpers.ts
```diff
@@ -297,6 +297,14 @@ export const userLink = (user: User): string => {
     return `https://github.com/${user.Login}`
 }
 
+export const evaluateDev = (path: string, originalUrl?: string): string => {
+    if (process.env.NODE_ENV === "development") {
+        return `https://localhost:443${path}`
```
JosteinLindhom, line 302: We should not restrict development mode to `localhost`.

One example where this does not work is when running QuickFeed in development mode on a remote VM.

It also does not account for the domain set in `.env`, i.e. if I set my development server to run on DOMAIN="some valid IP or public domain".

public/src/overmind/effects.tsx
```diff
@@ -12,9 +13,10 @@ export class ApiClient {
      * Must be called before accessing the client.
      * @param errorHandler A function that is called when an error occurs.
      */
-    public init(errorHandler: (payload?: { method: string; error: ConnectError; } | undefined) => void) {
+    public init(errorHandler: (payload?: { method: string; error: ConnectError } | undefined) => void) {
         this.client = createResponseClient(QuickFeedService, createConnectTransport({
-            baseUrl: `https://${window.location.host}`
+            baseUrl: evaluateDev("/", window.location.host),
```
JosteinLindhom, line 18: Using `/` here will cause issues if *not* run in development mode.
The URL that will be used if I run on `https://example.com` and call `GetUser` will end up being `https://example.com/example.com/qf.QuickFeedService/GetUser`.

web/auth/jwt.go
JoachimTislov: Yeah, that would be alot cleaner. I had a similar solution ealier but with the dev parameter..

public/src/Helpers.ts
JoachimTislov: Yes, I thought about adding a duplicate field in the public .env file, but might there be a better solution? 

I can set a domain field in the .env file for now

main.go
```diff
@@ -59,17 +59,22 @@ func main() {
...
 	var srvFn web.ServerType
+	domain := env.Domain()
 	if *dev {
 		srvFn = web.NewDevelopmentServer
+		domain = fmt.Sprintf("https://%s", domain)
 	} else {
 		srvFn = web.NewProductionServer
 	}
-	log.Printf("Starting QuickFeed on %s%s", env.Domain(), *httpAddr)
+	log.Printf("Starting QuickFeed on %s%s", domain, *httpAddr)
```
meling, line 77: We should print the full URL also for production, so please remove the lines above, and do this:
```go
log.Printf("Starting QuickFeed on https://%s%s", env.Domain(), *httpAddr)
```
Maybe we don't need to print the `httpAddr` if it is `:443` since that's the default `https` port.

web/auth/jwt.go
meling: I like this approach, just minor a fix:
```go
func NewTokenManager(db database.Database) (*TokenManager, error) {
	manager := &TokenManager{
		db:     db,
		secret: env.AuthSecret(),
	}
}
```

web/auth/jwt.go
meling: See the alternative approach to set the secret below. That way you avoid changing the token manager signature.

web/auth/jwt.go
```diff
@@ -65,7 +72,7 @@ func (tm *TokenManager) NewAuthCookie(userID uint64) (*http.Cookie, error) {
 		HttpOnly: true,
 		Secure:   true,
 		Expires:  time.Now().Add(cookieExpirationTime),
-		SameSite: http.SameSiteStrictMode,
+		SameSite: http.SameSiteNoneMode,
```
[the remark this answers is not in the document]
JoachimTislov: Oh, sorry. Forgot about that one...

I will come with a solution

main.go
```diff
@@ -59,17 +59,22 @@ func main() {
 		log.Fatal(err)
 	}
 
-	if env.Domain() == "localhost" {
-		log.Fatal(`Domain "localhost" is unsupported; use "127.0.0.1" instead.`)
-	}
+	/*
```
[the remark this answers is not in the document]
JoachimTislov: Comment below?

web/auth/jwt.go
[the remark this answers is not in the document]
JoachimTislov: @meling I have added simple if statement checking if in dev, but can make it simpler and avoid the hassle of sending in false in every other reference of the NewAuthCoookie function. Wouldn't be as understandable and consistent (can somehow have this env variable not empty in prod) but cleaner:

```
func sameSite() string {
	if os.Getenv("QUICKFEED_AUTH_SECRET") == "" {
		return http.SameSiteStrictMode
	}
	return http.SameSiteNoneMode
}
```

public/src/overmind/effects.tsx
JoachimTislov: @JosteinLindhom I have updated the evaluteDev

JosteinLindhom:
> The secret for the token manager is static between reloads and persists as the string: "dev".

Refer to [in-line comments](https://github.com/quickfeed/quickfeed/pull/1194#discussion_r1931683328) about this. 
We should support user defined secrets such that we don't invalidate all sessions every time the server restarts.
This should be made a separate issue and be implemented in a separate PR.

> CORS - cross origin request is allowed for the hardcoded url: http://localhost:8082/, used in the webpack. Dependency: https://github.com/rs/cors, should be easy to implement this single feature without it, but the package comes with perks.

> Backend: OAuth2- login and logout, redirects to the webpack url if in development, the base url is hardcoded there as well.

> Frontend: created a method called evaluateDev to determine the Api client's base url and used in some href's related to authentication (logout and login)

> The api client's credentials (cookie) is now included over cross origin requests.

It seems that many of the changes in this PR is included in order to support communication between the QuickFeed server and webpack-dev-server.

I'm not entirely sure what the benefits of running the frontend separate from the backend are other than not having to manually refresh the browser to pick up changes. 

I don't think the benefit is worth the added complexity.

I've been considering removing the `webpack-dev-server` altogether as running `webpack --mode=development --watch` works just fine when serving the frontend via the QuickFeed server.

> Callback url is wrong if quickfeed is executed with the http.addr flag and the port differs from 443, the github app won't register the change and redirect to the initial one. I can create a separate issue.. for this one

Good catch! This should absolutely be made a separate issue.

> Air-verse for hot reloading after changes in the backend

This could be started in the Docker container along with webpack in `--watch` mode.

Come to think of it (just thinking out loud), maybe users could supply some flag / env when running the container configuring how QuickFeed should run:

1. Plain development mode, simply starts the QuickFeed server in dev mode
2. Hot reload development mode, with air and webpack running
3. Production mode

We should also consider persisting the database (and other relevant stuff) to the host system using volumes such that updates to the image / re-creating a container is non-destructive.

web/auth/jwt.go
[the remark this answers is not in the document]
meling: The better solution is to use the same site always. We do not want to have a separate endpoint for dev testing. Jostein's solution seems to work well, so I don't think this extra endpoint is necessary.

docker-compose.yml
```diff
@@ -0,0 +1,7 @@
+services:
```
meling, line 1: Do we need a docker-compose file? The dockerfile seems to cover everything that is needed.

web/auth/jwt.go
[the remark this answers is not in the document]
JoachimTislov: SameSiteNoneMode is required if you want to call the quickfeed server from a different domain. Reasoning being the time it saves you from having to reload the page between each change. Air-verse doesn't reload the page on changes

docker-compose.yml
JoachimTislov: Well not really, it only maps the docker port 443 to the host machines port 443..

JoachimTislov:
> I don't think the benefit is worth the added complexity.

Well I partially agree, I just know the annoyence of having to click refresh when working on the frontend will be frustating in the long run.

> 2. Hot reload development mode, with air and webpack running

Are you thinking a pf flag which allows you to simply run quickfeed -dev -hot to start both air and a webpack concurrently? The make target dev-hot-reload currently executes this.

or 

Have a hot reload flag which simplifies the if statements of which redirect url and if the cookie is going to be strict or none?

> We should also consider persisting the database (and other relevant stuff) to the host system using volumes such that updates to the image / re-creating a container is non-destructive.

Yes, didn't think of that. Would running air and webpack inside of the development container be worth the work to configure it. I haven't had a look into it..

Makefile
```diff
@@ -79,3 +80,27 @@ selenium:
...
 qcm:
 	@cd cmd/qcm; go install
+
+new-githubApp:
+	@if [ ! -f ".env" ]; then cp ".env-template" ".env"; fi
+	@if [ ! -f "./public/.env" ]; then cp "./public/.env-template" "./public/.env"; fi
+	@go install
+	@if [ -f /usr/local/bin/quickfeed ]; then sudo rm -f /usr/local/bin/quickfeed; fi
+	@if [ ! -f $(quickfeedPath) ]; then echo "Quickfeed was not installed" exit 1; fi
+	@sudo mv $(shell go env GOPATH)/bin/quickfeed /usr/local/bin/
```
JosteinLindhom, line 90: If Go is installed correctly `GOPATH` should be in the users `PATH`, making this step and the one on line 88 unnecessary.

Makefile
```diff
@@ -79,3 +80,27 @@ selenium:
 
 qcm:
 	@cd cmd/qcm; go install
+
+new-githubApp:
+	@if [ ! -f ".env" ]; then cp ".env-template" ".env"; fi
```
JosteinLindhom, line 85: If the `.env` file does not exist, the created `.env` file will contain placeholder values at this point.
The created GitHub app will therefore be incorrectly configured unless the user want to run using the default value for DOMAIN.

JoachimTislov: Yes, that is true. Either the user updates the template or its not possible for a user who wants to deploy to production to use the make target.

Makefile
JoachimTislov: Ok, have I missed a simple way of setting binaries to path? This was my kind of work around to use a binary everywhere

docker-compose.yml
JoachimTislov: I will keep it for now, we will configure volumes for the database and possibly air-verse, then I think we will need the docker-compose file

JoachimTislov:
I will finish this when #1201 is merged into main and in the meantime look into docker volumes

public/src/components/navbar-buttons/LogoutButton.tsx
JoachimTislov, line 1: This fixes a deepsource issue

Makefile
```diff
@@ -79,3 +80,27 @@ selenium:
...
+	@if [ ! -f "./public/.env" ]; then cp "./public/.env-template" "./public/.env"; fi
+	@go install
+	@if [ -f /usr/local/bin/quickfeed ]; then sudo rm -f /usr/local/bin/quickfeed; fi
+	@if [ ! -f $(quickfeedPath) ]; then echo "Quickfeed was not installed" exit 1; fi
+	@sudo mv $(shell go env GOPATH)/bin/quickfeed /usr/local/bin/
+	@sudo setcap 'cap_net_bind_service=+ep' /usr/local/bin/quickfeed
```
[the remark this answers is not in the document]
JosteinLindhom: I see this target has been removed, however just adding an explanation:

On Linux ports 1-1024 are privileged, requiring sudo rights to bind to them. The `setcap` command can grants this permission to a binary such that it can be run as non-root while still binding to a privileged port.

On Windows and macOS, this is handled differently and the `setcap` command does not exists.
So in case someone wants to run QuickFeed in dev mode or otherwise on macOS or Windows, the referenced step would crash, preventing any subsequent steps to run.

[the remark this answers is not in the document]
JoachimTislov: Oh ok

meling:
This has too many useless commits to fix up things, so rather than me squashing the commits, I suggest that you open a new PR from master and bring each of the files from this PR into the new one, using a few commits of appropriate size and with appropriate commit messages. Please also include usage instructions in the PR description.

JoachimTislov:
> This has too many useless commits to fix up things, so rather than me squashing the commits, I suggest that you open a new PR from master and bring each of the files from this PR into the new one, using a few commits of appropriate size and with appropriate commit messages. Please also include usage instructions in the PR description.

Ok, I understand.

JoachimTislov:
Close in favour of #1220

PR #1200, opened 2025-01-29
Add/Update documentation and create a program for generating graphs, visualizing quickfeed's structure

### LINKED ISSUE #1171: chore: Write more documentation and update existing documentation
The code base already has quite amount of documentation, but I would recommend adding a main documentation file which allows anyone who's new to the project get familiar with the code.

Main documentation file:
- Will give an overview over all of the other documentation
- The README.md file in root folder should have a link to this file
- Location of the file could be root folder, but makes sense to have it in the current "doc" subfolder

General improvements:
- Update old documentation with outdated instructions/information, for example the devtools target was changed to tools in the makefile.
- Add information for each subfolder within the root quickfeed folder, explaining what operations and functionality the folder contains, and its role in the quickfeed application.
- Create a list over different abbreviations in the code base, like scm - source code management
- Write more comments in the code, explaining the logic within methods.

JoachimTislov:
Closes #1171 when im finished.

Graph:

Currently redesigning the structure of the graph-generation program to be more flexible, interactive and essentially save time. Allowing for easy updates and creation of visual maps over specified parts of the system, not just everything.

I am also missing a tool which can retrieve references in the react app, and would also like to somehow connect the backend and frontend by using protobuf's features.

Documentation:

- Created a file for overview over everything documentation related
- Moved some information around to define a better structure.


meling:
Did a quick pass on this. (Not the code)... I prefer smaller PRs.
I suggest that you close this and reopen a new PR with only documentation changes.

If your documentation markdown files refers to the package graph, then it should be included as a viewable image (PNG/SVG) either embedded in the markdown doc or via a link. The .dot file could be included too, but keep it together with the image file. Most users maybe don't want to view the graph as a .dot file.

The graph generator should be added as a separate PR, or in an external repository. Note that there are existing tools that does similar things, so make sure your effort isn't wasted, if there is another tool that could be used instead.

JoachimTislov:
The graph generation isn't any complicated code, and I hope theres existing libraries for this... I just didn't find any in my search. As of the direction of the code, I think setting up a repository would be the best decision.. and then it can be used as depency if it ends up being reliable and useful. I will try my best.

I will open a new PR with only documentation.

PR #1201, opened 2025-01-30
feat: add file watcher for live-reloading in development environment

JosteinLindhom:
This PR adds live reloading of the browser if changes are detected in the given folder (`/public/dist` in this case).

If QuickFeed starts in -dev mode, we add an endpoint `/watch` that the frontend can connect to. If a file change is detected, the server sends a message to the frontend client, which triggers a reload.

To activate this functionality the server has to run in development mode, and the frontend must be compiled in development mode.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/alta/protopatch v0.5.3
 	github.com/beatlabs/github-auth v0.0.0-20240615135342-292f72d79b19
 	github.com/docker/docker v27.1.1+incompatible
+	github.com/fsnotify/fsnotify v1.8.0
 	github.com/go-git/go-git/v5 v5.13.0
 	github.com/golang-jwt/jwt v3.2.2+incompatible
 	github.com/google/go-cmp v0.6.0
@@ -44,7 +45,6 @@ require (
 	github.com/emirpasic/gods v1.18.1 // indirect
 	github.com/fatih/structtag v1.2.0 // indirect
 	github.com/felixge/httpsnoop v1.0.4 // indirect
-	github.com/fsnotify/fsnotify v1.8.0
 	github.com/go-git/gcfg v1.5.1-0.20230307220236-3a3c6141e376 // indirect
 	github.com/go-git/go-billy/v5 v5.6.0 // indirect
 	github.com/go-logr/logr v1.4.2 // indirect
--- a/go.mod
+++ b/go.mod
@@ -44,6 +44,7 @@ require (
 	github.com/emirpasic/gods v1.18.1 // indirect
 	github.com/fatih/structtag v1.2.0 // indirect
 	github.com/felixge/httpsnoop v1.0.4 // indirect
+	github.com/fsnotify/fsnotify v1.8.0
 	github.com/go-git/gcfg v1.5.1-0.20230307220236-3a3c6141e376 // indirect
 	github.com/go-git/go-billy/v5 v5.6.0 // indirect
 	github.com/go-logr/logr v1.4.2 // indirect
--- a/internal/reload/watcher.go
+++ b/internal/reload/watcher.go
@@ -1,0 +1,113 @@
+package reload
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"sync"
+
+	"github.com/fsnotify/fsnotify"
+)
+
+type Watcher struct {
+	clients map[chan string]bool
+	mu      sync.Mutex
+}
+
+// NewWatcher creates a new watcher for the given path.
+// The watcher listens for file changes and broadcasts them to all connected clients.
+// Note: Usually you only ever have one client as this is intended for live-reloading
+// a web page in a development environment.
+func NewWatcher(path string) *Watcher {
+	watcher := &Watcher{
+		clients: make(map[chan string]bool),
+	}
+	go watcher.watch(path)
+	return watcher
+}
+
+func (w *Watcher) watch(path string) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer watcher.Close()
+	// Start watching the given path.
+	err = watcher.Add(path)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Start listening for events.
+	for {
+		select {
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return
+			}
+			// We only care about writes and creates.
+			if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) {
+				// Broadcast event to all clients.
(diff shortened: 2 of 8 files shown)


main.go
```diff
@@ -121,6 +122,12 @@ func main() {
 	router := qfService.RegisterRouter(tokenManager, authConfig, *public)
 	handler := h2c.NewHandler(router, &http2.Server{})
 
+	if *dev {
+		// Wrap handler with file watcher
+		// for live-reloading in development mode.
+		handler = web.DevHandler(*watch, handler)
+	}
```
JoachimTislov, line 129: You can simplify it to:

```go
if *dev && *watch {
 
 handler = web.DevHandler(handler)

}

Comment (JosteinLindhom) [reply main.go]: True!

File: go.mod [hunk 2]
```diff
@@ -44,6 +44,7 @@ require (
 	github.com/emirpasic/gods v1.18.1 // indirect
 	github.com/fatih/structtag v1.2.0 // indirect
 	github.com/felixge/httpsnoop v1.0.4 // indirect
+	github.com/fsnotify/fsnotify v1.8.0
```
meling, line 47: Should not be in the `indirect` section.

internal/reload/watcher.go
```diff
@@ -0,0 +1,128 @@
...
+func NewWatcher(path string) (*Watcher, error) {
+	watcher := &Watcher{
+		clients: make(map[chan string]bool),
+	}
+	w, err := fsnotify.NewWatcher()
+	if err != nil {
+		log.Printf("Failed to create watcher: %v", err)
+		return nil, err
+	}
+	defer w.Close()
```
JosteinLindhom, line 32: Just a heads up to future Jostein:

```go
thing, close := newThing()
defer close() // why we don't code late at night
go thing.needsToBeNotClosed()
return thing
```

Is not a smart way to keep things in the goroutine working.

(This comment was sponsored by the 7 gray strands of hair generated by line 32)

internal/reload/watcher.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+
+func (watcher *Watcher) Handler(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "text/event-stream")
+	w.Header().Set("Cache-Control", "no-cache")
+	w.Header().Set("Connection", "keep-alive")
+
+	// Create a new channel for this client
+	client := make(chan string)
```
meling, line 103: should the client channel have more capacity? this channel requires synchronization between the sending and receiving goroutine. I suspect that perhaps at least 1 element capacity would be beneficial, but not completely sure.

internal/reload/watcher.go
```diff
@@ -0,0 +1,129 @@
...
+	watcher.addClient(client)
+	// Listen for changes and send updates
+	for {
+		select {
+		case msg, ok := <-client:
+			if !ok {
```
meling, line 109: If I recall correctly, this `!ok` thing happens when the channel is closed. Currently, I think it is only closed by `removeClient`, which is called below when the client disconnects and this for-loop returns... Hence, I don't think this check is necessary.

internal/reload/watcher.go
```diff
@@ -0,0 +1,129 @@
...
+			fmt.Fprintf(w, "data: %s\n\n", msg)
+			w.(http.Flusher).Flush()
+		case <-r.Context().Done(): // Client disconnected
+			watcher.removeClient(client)
+			return
+		}
+	}
+}
+
+func (*Watcher) webpack() {
```
meling, line 121: Do we want this to be a method? It doesn't need to be, but I'm willing to be convinced that it is useful to keep it together with the `Watcher` type.

internal/reload/watcher.go
```diff
@@ -0,0 +1,129 @@
...
+// Broadcast a message to all clients
+func (w *Watcher) broadcastMessage(msg string) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	for ch := range w.clients {
+		select {
+		case ch <- msg:
+			continue
+		default:
+			log.Println("Failed to send message to client")
```
meling, line 92: Does it make sense to log the message that was dropped.

internal/reload/watcher.go
```diff
@@ -0,0 +1,129 @@
...
+
+		case err := <-watcher.Errors:
+			if err != nil {
+				log.Println("error watching files:", err)
+			}
+		}
+	}
+}
+
+func (w *Watcher) addClient(ch chan string) {
```
meling, line 69: doc comment.

internal/reload/watcher.go
```diff
@@ -0,0 +1,129 @@
...
+				return
+			}
+			// We only care about writes and creates.
+			if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) {
+				// Broadcast event to all clients.
+				w.broadcastMessage(event.Name)
+			}
+
+		case err := <-watcher.Errors:
+			if err != nil {
```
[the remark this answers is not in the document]
meling: On second thought, do we expect such errors to happen while watching ... but it is okay for the watcher to continue. If so, we should log errors instead of returning the error. In either case, I don't think we need to check `if err != nil`... just log or return. I assume that the `Errors` channel is closed only if the watcher is closed, in which case, the `err` will be `nil` and we can return... But not sure.

internal/reload/watcher_test.go
```diff
@@ -0,0 +1,91 @@
...
+		for {
+			line, err := reader.ReadString('\n')
+			if err != nil {
+				break
+			}
+			// ignore empty lines; not all received messages contain data
+			if line != "" {
+				// send the message to the event channel
+				eventChan <- line
+				break
```
meling, line 74: Is this intended? It will read only one line and exit the for-loop/goroutine.

meling (changes requested):

internal/reload/watcher_test.go
JosteinLindhom: We're only interested in checking if we receive a non-empty message.
Since we only check for a single event below there was no point in sending multiple events on the channel here.

The main thing tested here is that the file modification on line 47 is picked up and sent as expected to our client.
I trigger a write 3 times, with some delay between writes to account for startup.

The received event will be identical for all three events ("data: $filename").

internal/reload/watcher.go
```diff
@@ -107,7 +102,7 @@ func (watcher *Watcher) Handler(w http.ResponseWriter, r *http.Request) {
 	for {
 		select {
 		case msg := <-client:
-			fmt.Fprintf(w, "data: %s\n\n", msg)
```
JosteinLindhom: Apparently this extra newline is required to indicate end of message.

meling (approved):

Issue #1198, opened 2025-01-28
feat: Support user defined secrets


JoachimTislov:
Discussed in issue: #1194 

Related:

```go
func AuthSecret() string {
	authSecret := os.Getenv("QUICKFEED_AUTH_SECRET")
	if authSecret == "" {
		return rand.String()
	}
	return authSecret
}
```

> We should support user defined secrets such that we don't invalidate all sessions every time the server restarts. 

Uncertain if the simple function above is all you wanted to implement to achieve this feature?  @JosteinLindhom

meling:
I believe it is all you need; and calling it in the right place of course.

meling:
We probably want to add it to the `.env` file as well.

PR #1205, opened 2025-01-31
Add support for user defined secrets

JoachimTislov:
Closes: #1198


internal/env/authSecret.go
```diff
@@ -0,0 +1,18 @@
+package env
```
JosteinLindhom, line 1: Rename this file to `auth_secret.go` to keep it consistent with how `.go` files are named elsewhere.

.env-template
```diff
@@ -6,6 +6,10 @@ QUICKFEED_APP_ID=""
 QUICKFEED_APP_KEY=$QUICKFEED/internal/config/github/quickfeed.pem
 QUICKFEED_CLIENT_ID=""
 QUICKFEED_CLIENT_SECRET=""
+
+# Empty for random secret, or set your own custom secret
+QUICKFEED_AUTH_SECRET=""
```
JosteinLindhom, line 11: This should be added to the docs.

We should explain the consequences of setting / not setting a secret here.
- Not setting a secret will invalidate all JWT tokens when the server is restarted.
- If the secret is revealed, or easily guessed, no JWT can be trusted. i.e, with a secret set to `secret`, anyone who knows this can modify and sign their own JWT.

internal/env/authSecret.go
JoachimTislov: Oh didn't notice those

.env-template
JoachimTislov: I added a field in deploy.md, can refer to this in the .env-template, wasn't sure if you wanted the documentation directly in the file or in docs.

JosteinLindhom:
If it makes sense we could also make some tool that automatically generates and sets a secret, such that users do not have to do this themselves.

For example:

```go
// cmd/secret/main.go
package main

import (
	"fmt"

	"github.com/quickfeed/quickfeed/internal/env"
	"github.com/quickfeed/quickfeed/internal/rand"
)

// This tool is used to generate a random secret for signing JWT tokens.
func main() {
	fmt.Println("Generating random secret for signing JWT tokens...")
	if err := env.Save(env.RootEnv(".env"), map[string]string{
		"QUICKFEED_AUTH_SECRET": rand.String(),
	}); err != nil {
		panic(err)
	}
	fmt.Println("Done.")
}
```

And a make target:

```make
// Makefile
secret:
    go run cmd/secret/main.go
```

Running `make secret` will then create or replace the key/value pair `QUICKFEED_AUTH_SECRET=<some string>`

```sh
$ cat .env | grep AUTH_SECRET
QUICKFEED_AUTH_SECRET=""
$ make secret
go run cmd/secret/main.go
Generating random secret for signing JWT tokens...
Done.
$ cat .env | grep AUTH_SECRET
QUICKFEED_AUTH_SECRET=a7bd9e1d93c5ada92d0ae8c9febeb67110eaa7449f9a6e22e26a28ee77e8eb61

File: internal/env/auth_secret.go [hunk 3]
```diff
@@ -0,0 +1,18 @@
+package env
+
+import (
+	"os"
+
+	"github.com/quickfeed/quickfeed/internal/rand"
+)
+
+// AuthSecret returns the secret used to sign JWT tokens.
```
meling, line 9:
```go
// AuthSecret returns the JWT signing secret obtained from the QUICKFEED_AUTH_SECRET
// environment variable, or a random secret if unset.
```
I'm not sure we want to return a random secret if it is unset, since we should require that `QUICKFEED_AUTH_SECRET` is set when we start the server or that we set it via the `-secret` flag (it gets saved and loaded in `main.go`.)

However, I don't like the idea of panicking if it is unset either. @JosteinLindhom what do you think?

meling (approved):
I'm marking this as approved; it looks pretty good to me now. I've moved some code to `main.go` and rewrote the `deploy.md` doc. Please take a look @JoachimTislov and @JosteinLindhom. I also have one small open question.

JoachimTislov:
> I'm marking this as approved; it looks pretty good to me now. I've moved some code to `main.go` and rewrote the `deploy.md` doc. Please take a look @JoachimTislov and @JosteinLindhom. I also have one small open question.

The only thing is when executing quickfeed with air, it will write the same secret to the env file each time the server restarts, but otherwise it looks good. Air-verse is configured in this [PR](https://github.com/quickfeed/quickfeed/pull/1194)

meling:
> The only thing is when executing quickfeed with air, it will write the same secret to the env file each time the server restarts, but otherwise it looks good. Air-verse is configured in this pr #1194

It isn't necessary to run quickfeed with the `-secret` flag when running with air. You can hardcode a local-only (for watching) secret in the `.env` file, or run with the `-secret` flag the first time.

> Edit: Also, I have experienced issues when running binaries/programs with `sudo` which alter files. Not allowed to either access the file or delete it without elevated privileges.

This doesn't sound right. Running with `sudo` is what gives you elevated privileges. If anything, this must be a WSL/Windows issue unrelated to the change in this PR, no? We already write to the `.env` file when creating the GitHub app with the `-new` flag.

JoachimTislov:
> This doesn't sound right. Running with `sudo` is what gives you elevated privileges. If anything, this must be a WSL/Windows issue unrelated to the change in this PR, no? We already write to the `.env` file when creating the GitHub app with the `-new` flag.

Yes not related, but will trigger the issue on WSL.

The issue happens in the specific case where a user would run quickfeed with `sudo quickfeed -secret *string*` to bypass setting capabilities for the port permission.

Issue #1202, opened 2025-01-30
chore: Resolve five deepsource issues in the backend


JoachimTislov:
- SCT-A000: Possible hardcoded secrets detected in source code
- RVV-B0012: Unused parameter in function
- GO-W1027: Empty slice literal used to declare a variable
- GO-W1032: Use t.Setenv and friends instead of os.Setenv for test file(s)
- GO-W1029: Usage of both value and pointer receivers

PR #1206, opened 2025-02-03
Deepsource, backend and secret issues, 16 in total.

JoachimTislov:
Partially Resolves: #1202

RVV-B0012 was commited with SCT-A000, my bad.

GO-W1027 involved updating the `GetTasks` to return nil when tasks array is empty, and remove unnecessary initialization in `SynchronizeAssignmentTasks`

Codes for issues:
- SCT-A000: Possible hardcoded secrets detected in source code
- RVV-B0012: Unused parameter in function
- GO-W1027: Empty slice literal used to declare a variable
- GO-W1032: Use t.Setenv and friends instead of os.Setenv for test file(s)
- GO-W1029: Usage of both value and pointer receivers

### DIFF
--- a/ci/run_tests_clone_test.go
+++ b/ci/run_tests_clone_test.go
@@ -11,7 +11,7 @@ import (
 )
 
 func TestCloneAndCopyRunTests(t *testing.T) {
-	t.Setenv("QUICKFEED_REPOSITORY_PATH", "QUICKFEED_REPOSITORY_PATH")
+	t.Setenv("QUICKFEED_REPOSITORY_PATH", "$HOME/tmp/courses")
 	qfTestOrg := scm.GetTestOrganization(t)
 	sc, qfUserName := scm.GetTestSCM(t)
 
--- a/ci/run_tests_clone_test.go
+++ b/ci/run_tests_clone_test.go
@@ -2,7 +2,6 @@ package ci
 
 import (
 	"context"
-	"os"
 	"strings"
 	"testing"
 
@@ -11,11 +10,8 @@ import (
 	"github.com/quickfeed/quickfeed/scm"
 )
 
-func init() {
-	os.Setenv("QUICKFEED_REPOSITORY_PATH", "$HOME/tmp/courses")
-}
-
 func TestCloneAndCopyRunTests(t *testing.T) {
+	t.Setenv("QUICKFEED_REPOSITORY_PATH", "QUICKFEED_REPOSITORY_PATH")
 	qfTestOrg := scm.GetTestOrganization(t)
 	sc, qfUserName := scm.GetTestSCM(t)
 
--- a/qf/submission.go
+++ b/qf/submission.go
@@ -115,7 +115,7 @@ func (s *Submission) BeforeCreate(tx *gorm.DB) error {
 	if s.GetGroupID() > 0 {
 		// If the submission is for a group, create a new grade for each user in the group.
 		// Get all the user IDs in the group
-		userIDs := []uint64{}
+		var userIDs []uint64
 		tx.Model(&Enrollment{}).Where("group_id = ?", s.GetGroupID()).Pluck("user_id", &userIDs)
 
 		if len(userIDs) == 0 {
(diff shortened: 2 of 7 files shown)


JosteinLindhom:
It's kind of hard to figure out what each change in this PR does, or why it's done.

It would be helpful if either:
- One commit includes changes related to a single Deepsource issue, with the commit indicating which issue it fixes (e.g. `deepsource(JS-0336): Avoid using promises in places not designed to handle them`
- One PR focused on a smaller number (or even one) issue.

What do you think @meling ?

Tackling a lot (or even all, as indicated in this PR) of the Deepsource issues at once I fear will result in a gigantic PR.

Also, note that some of the changes introduced so far in this PR so far alters / breaks the UI, one example:
image

Master:
image

JoachimTislov:
> Also, note that some of the changes introduced so far in this PR so far alters / breaks the UI

Sorry about this. Solved a specific issue where deepsource doesn't recommend "onclick" on elements which aren't designed for being clicked by converting to a button element. I will fix this first, I don't know why I did those changes apart from wanting to completing the issue deepsource presented.

The better solution is to either ignore with comment or add onkeydown or onkeyup. I will address this later.

> It would be helpful if either:

>  - One commit includes changes related to a single Deepsource issue, with the commit indicating which issue it fixes (e.g. deepsource(JS-0336): Avoid using promises in places not designed to handle them
> - One PR focused on a smaller number (or even one) issue.
What do you think @meling ?

> Tackling a lot (or even all, as indicated in this PR) of the Deepsource issues at once I fear will result in a gigantic PR.

I agree about splitting it up, but containing it in one PR would be possible. I will write commits for the specific issues, with code and number of instances for given issue.

meling:
I think we discussed it last week @JoachimTislov ... I think we shouldn't waste time fixing deep source issues in the frontend until we have migrated away from overmind to a new state manager. Otherwise, we may be fixing things that aren't going to be relevant with a new state manager.

Moreover, I agree with @JosteinLindhom that we should break up the deep source fixes into separate smaller PRs that can be merged more easily and documented with adequate commit messages.

JoachimTislov:
> I think we discussed it last week @JoachimTislov ... I think we shouldn't waste time fixing deep source issues in the frontend until we have migrated away from overmind to a new state manager. Otherwise, we may be fixing things that aren't going to be relevant with a new state manager.

No, we did not. Theres multiple issues which won't be resolved by implementing a new state manager, and they aren't complex. I think solving these before implementing the new manager is applicable

> Moreover, I agree with @JosteinLindhom that we should break up the deep source fixes into separate smaller PRs that can be merged more easily and documented with adequate commit messages.

Ok, I will split them into PRs

qf/submission.go
```diff
@@ -115,7 +115,7 @@ func (s *Submission) BeforeCreate(tx *gorm.DB) error {
 	if s.GetGroupID() > 0 {
 		// If the submission is for a group, create a new grade for each user in the group.
 		// Get all the user IDs in the group
-		userIDs := []uint64{}
+		var userIDs []uint64
```
meling, line 118: Please create an issue to add a test for `Submission.BeforeCreate`. I don't think we have a test for this method.

Should be done in a separate PR.

meling (changes requested):
Note that `RunData` vs `*RunData` was intentional since most methods do not and should not update `RunData`... But I think it is fine to make the change to keep things consistent and follow the recommendation.

ci/run_tests_clone_test.go
```diff
@@ -11,11 +10,8 @@ import (
 	"github.com/quickfeed/quickfeed/scm"
 )
 
-func init() {
-	os.Setenv("QUICKFEED_REPOSITORY_PATH", "$HOME/tmp/courses")
-}
-
 func TestCloneAndCopyRunTests(t *testing.T) {
+	t.Setenv("QUICKFEED_REPOSITORY_PATH", "QUICKFEED_REPOSITORY_PATH")
```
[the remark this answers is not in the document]
JoachimTislov: Oh, not sure how that happend..

qf/submission.go
JoachimTislov: Will do.

JoachimTislov:
> Note that `RunData` vs `*RunData` was intentional since most methods do not and should not update `RunData`... But I think it is fine to make the change to keep things consistent and follow the recommendation.

Oh so thats the different between those two.. aah makes sense

meling (approved):
Looks good to me!

Issue #1208, opened 2025-02-06
Add test for Submission BeforeCreate


JoachimTislov:
In file submission.go, line 102

JosteinLindhom:
> [@meling](https://github.com/meling) I think we can remove BeforeCreate, it has zero references

It's a hook used by GormDB (<https://gorm.io/docs/hooks.html>), not by us directly.
The logic within is triggered just before a submission is created in the database.

It cannot be removed unless the logic is moved somewhere else.

JoachimTislov:
> It's a hook used by GormDB (https://gorm.io/docs/hooks.html), not by us directly. The logic within is triggered just before a submission is created in the database.
> 
> It cannot be removed unless the logic is moved somewhere else.

Ok, I understand

meling:
> > [@meling](https://github.com/meling) I think we can remove BeforeCreate, it has zero references
> 
> It's a hook used by GormDB (https://gorm.io/docs/hooks.html), not by us directly. The logic within is triggered just before a submission is created in the database.

We should add a comment about this in BeforeCreate's doc comment.

PR #1209, opened 2025-02-07
Add test for CreateSubmission and resolve bug of having to deliver a new submission two times.

JoachimTislov:
Closes: #1208 

I had to create a new helper for the required gorm.DB type. The gorm.DB instance used in the database folder is not exported, and could therefore not use the TestDB helper.

I think I have come across a minor mistake in the group part of the function. The grades in submission is overwritten and then its later attempting to retrieve grades for the group. Not sure if this is intentional.

The go test fails because of that specific case.

Theres also the case where both user and group id is larger than zero and the user's grades will be overwritten by the group. I would assume this case does not happen, but adding a if statement to handle it wouldn't hurt.


JosteinLindhom:
> I think I have come across a minor mistake in the group part of the function. The grades in submission is overwritten and then its later attempting to retrieve grades for the group. Not sure if this is intentional.

I don't think this is intentional. 

> The go test fails because of that specific case.
> 
> Theres also the case where both user and group id is larger than zero and the user's grades will be overwritten by the group. I would assume this case does not happen, but adding a if statement to handle it wouldn't hurt.

Good observation. 

The `BeforeCreate` hook is only invoked in `CreateSubmission`, specifically on this line:

https://github.com/quickfeed/quickfeed/blob/6bc80cd135c4ff27398fb4de545377a6a0ab375c/database/gormdb_submission.go#L63

If the `db.check(submission)` on line 24 check fails, it will not create the submission.
The `check` function checks if both are `> 0`, and returns an error if they are.

**Having thought about this for a bit:**

- `CreateSubmission` is only used in `RecordResult`
- `BeforeCreate` specifically is only invoked if a previous submission does not already exist in the database, i.e. it's a new submission
- The new submission does *not* have any grades associated with it, as these are created in `BeforeCreate`

As a consequence, `s.GetStatusByUser` will *always* return NONE.
This will be an issue if the new submission has a score that is greater than or equal to the associated assignment score limit _and_ the assignment is autoapproved. Otherwise, submissions correctly get status NONE.

This function creates the new submission:

https://github.com/quickfeed/quickfeed/blob/6bc80cd135c4ff27398fb4de545377a6a0ab375c/ci/record_results.go#L83-L100

If `previous` is `nil`, which it will be the first time a student / group pushes for a given assignment, `r.Assignment.SubmissionStatus(previous, score)` will also return `nil` (meaning  `submission.Grades == nil`)

The returned new submission is then used here:

https://github.com/quickfeed/quickfeed/blob/6bc80cd135c4ff27398fb4de545377a6a0ab375c/ci/record_results.go#L33-L36

which will invoke the `BeforeCreate` hook that will create all `NONE` grades for user(s) associated with the submission. Both Submission and Grade is then inserted into the database

JoachimTislov:
Ok, good work. I will look into this further

JosteinLindhom:
It would likely be sufficient to fetch the associated assignment in `BeforeCreate` and do something like this *after* initializing the grades:

```go
s.Grades = assignment.SubmissionStatus(s, s.GetScore())
```

qf/submission_test.go
```diff
@@ -208,3 +212,45 @@ func TestCountApprovedSubmissions(t *testing.T) {
...
+		}},
+	}
+
+	for key, test := range tests {
+		t.Run(key, func(t *testing.T) {
+			err := test.submission.BeforeCreate(test.db)
+			if (err == nil) != (test.wantErr == nil) || err != nil && test.wantErr != nil && err.Error() != test.wantErr.Error() {
+				t.Errorf("Expected err: %v, got: %v\n", err, test.wantErr)
+			}
+			if !reflect.DeepEqual(test.wantResult, test.submission.Grades) {
```
meling, line 251: Replace with `cmp.Equal` or `cmp.Diff`

qf/submission_test.go
```diff
@@ -208,3 +212,45 @@ func TestCountApprovedSubmissions(t *testing.T) {
...
+			{UserID: 1, SubmissionID: submission3.GetID(), Status: qf.Submission_NONE},
+			{UserID: 2, SubmissionID: submission3.GetID(), Status: qf.Submission_REJECTED},
+			{UserID: 3, SubmissionID: submission3.GetID(), Status: qf.Submission_REVISION},
+		}},
+	}
+
+	for key, test := range tests {
+		t.Run(key, func(t *testing.T) {
+			err := test.submission.BeforeCreate(test.db)
+			if (err == nil) != (test.wantErr == nil) || err != nil && test.wantErr != nil && err.Error() != test.wantErr.Error() {
```
meling, line 248: Too long if expression

internal/qtest/test_helper.go
```diff
@@ -10,14 +10,19 @@ import (
 	"connectrpc.com/connect"
 	"github.com/quickfeed/quickfeed/database"
 	"github.com/quickfeed/quickfeed/qf"
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+	"gorm.io/gorm/schema"
 )
 
+const testDb = "test.db"
```
meling, line 18: rename: `testDB`

database/gormdb_submission_internal_test.go
```diff
@@ -0,0 +1,80 @@
...
+		if err := gormDB.Close(); err != nil {
+			t.Error(err)
+		}
+		if err := os.Remove(f.Name()); err != nil {
+			t.Error(err)
+		}
+	}
+}
+
+func TestSetGrades(t *testing.T) {
```
meling, line 41: We have a lot of tests of the `CreateSubmission` method. I wonder if this test is actually necessary, or if the code paths that this test exercises in `setGrade` is already covered.

If it is not fully covered, I would much rather that we create a test that calls `CreateSubmission` instead of testing the `setGrade` method on its own. I know it was probably my suggestion to add tests for this method, but in retrospect, now that we have moved it into the database package, it seems less of a thing that we need to test separately. Also, best practices for testing is to test exported methods/interfaces, and not internal methods like `setGrade`... Hence, if we can avoid this test, there is less code to maintain and we can also drop the `testGormDB` helper.

database/gormdb_submission.go
```diff
@@ -59,6 +59,12 @@ func (db *GormDB) CreateSubmission(submission *qf.Submission) error {
 				sc.SubmissionID = submission.ID
 			}
 		}
+		if submission.ID == 0 {
+			// Initialize grades for the new submission
+			if err := setGrades(tx, submission); err != nil {
+				return err // will rollback transaction
```
meling, line 65: Maybe we could have a test that exercises this error path too.

database/gormdb_submission_internal_test.go
JoachimTislov: Yes, I agree. The amount of code for testing `setGrades` is not worth it. The only part which wasn't covered by the create submission tests except the bug, were the if statement checking if the group is empty which in theory will never happen since a group can't be created as empty. This restricts testing for the case as well.

database/gormdb_submission.go
JoachimTislov: Set grades will in theory never return `ErrRecordNotFound`, because of this check in [check](https://github.com/quickfeed/quickfeed/blob/master/database/gormdb_submission.go#71)

```go
var assignment int64
if err := db.conn.Model(&qf.Assignment{}).Where(&qf.Assignment{
   ID: submission.AssignmentID,
}).Count(&assignment).Error; err != nil {
   return fmt.Errorf("assignment %d not found: %w", submission.AssignmentID, err)
}

// Exactly one assignment and user/group must exist together.
if assignment+idCount != 2 {
   return fmt.Errorf("inconsistent database state: %w", gorm.ErrRecordNotFound)
}
```

I am not certain if the other possible errors should be tested.

JoachimTislov: Update, I changed from `t.Fatal(err)` to `t.Error(err)` since an error should not occur when creating a submission, which will detect those cases.

JoachimTislov: Nevermind codecov disagrees

meling (approved):
Looks good to me! @JosteinLindhom do you have anything more to add?

database/gormdb_submission.go
```diff
@@ -67,6 +72,41 @@ func (db *GormDB) CreateSubmission(submission *qf.Submission) error {
 	})
 }
 
+// setGrades adds grades for any user or group related to the submission
+// which are then saved to the database upon creation of the submission.
+func setGrades(tx *gorm.DB, s *qf.Submission) error {
+	if s.GetUserID() > 0 {
```
meling, line 78: To be bulletproof, we could use a switch statement here instead, since that prevents the case where the `Submission` has both `UserID > 0` and `GroupID > 0`... They should not both be `> 0`. And why remove the zero check on both:
```go
if s.GetUserID() == 0 && s.GetGroupID() == 0 {
	return errors.New("submission must have either user or group ID")
}
```
Is this not necessary?

database/gormdb_submission_test.go
```diff
@@ -41,6 +41,64 @@ func setupCourseAssignment(t *testing.T, db database.Database) (*qf.User, *qf.Co
...
+
+	want := &qf.Submission{
+		ID:           submission.ID,
+		AssignmentID: assignment.ID,
+		UserID:       user.ID,
+		Score:        1,
+		Grades:       []*qf.Grade{{UserID: user.ID, SubmissionID: submission.ID, Status: qf.Submission_APPROVED}},
+	}
+
+	if diff := cmp.Diff(submission, want, protocmp.Transform()); diff != "" {
```
meling, line 75: I prefer to use `got` to make it clear that it is the result of the operation we are testing.

database/gormdb_submission_test.go
```diff
@@ -41,6 +41,64 @@ func setupCourseAssignment(t *testing.T, db database.Database) (*qf.User, *qf.Co
...
+		t.Errorf("Expected same submission, but got (-sub +want):\n%s", diff)
```
meling, line 76: Tell which function is being tested:
`"CreateSubmission(): (-got +want):\n%s"`

database/gormdb_submission.go
```diff
@@ -67,6 +72,41 @@ func (db *GormDB) CreateSubmission(submission *qf.Submission) error {
...
+			}
+		}
+	}
+	// Find the assignment associated with the submission
+	var assignment qf.Assignment
+	if err := tx.First(&assignment, s.GetAssignmentID()).Error; err != nil {
+		return err
+	}
+	// Set the submission status based on the assignment's auto-approve settings
+	s.Grades = assignment.SubmissionStatus(s, s.GetScore())
```
meling, line 106: This will overwrite whatever is set for `s.Grades` on lines 81 and 95. So this cannot be correct; otherwise the initialization above is useless, no?

meling (changes requested):

database/gormdb_submission.go
JoachimTislov: This check is done in [check](https://github.com/quickfeed/quickfeed/blob/master/database/gormdb_submission.go#L71) which is called before `setGrade`

database/gormdb_submission.go
JoachimTislov: Partly true, it only overides the grades to approved if the statement `a.GetAutoApprove() && score >= a.GetScoreLimit()` is satisfied.

```go
func (a *Assignment) SubmissionStatus(latest *Submission, score uint32) []*Grade {
	if a.GetAutoApprove() && score >= a.GetScoreLimit() {
		latest.SetGradeAll(Submission_APPROVED)
	}
	// keep existing status if already approved/revision/rejected
	return latest.GetGrades()
}
```

I don't like how its written, its not that understandable. I can add more comments explaining why that logic is there.

JosteinLindhom: Since the incoming submission is a new submission it does not have any grades yet, i.e. the `s.Grades` is empty.

If the submission is a user submission we create a new grade for the user. 
`s.GetStatusByUser(s.GetUserID())` will always return `NONE` on line 83, and can be omitted since `NONE` == `0` *or* replaced with `qf.Submission_NONE`.
The same goes if it's a group submission.

These grades need to exist for `assignment.SubmissionStatus(s, s.GetScore())` to return correct grades based on `s` as it will try to `SetGradeAll`.

database/gormdb_submission.go
```diff
@@ -67,6 +72,41 @@ func (db *GormDB) CreateSubmission(submission *qf.Submission) error {
 	})
 }
 
+// setGrades adds grades for any user or group related to the submission
```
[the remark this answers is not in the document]
JoachimTislov: > ```go
> // InitUserIDs initializes the submission with the provided user IDs.
> func (s *Submission) InitUserIDs(userIDs []uint64) {
> 	s.Grades = make([]*Grade, len(userIDs))
> 	for i, userID := range userIDs {
> 		s.Grades[i] = &Grade{
> 			UserID:       userID,
> 			SubmissionID: s.GetID(), // TODO(meling): @JosteinLindhom is this necessary?
> 		}
> 	}
> }
> ```

I think creating a method for this isn't necessary. It will only be used by setGrades.

JoachimTislov:
Keep in mind, `TestCreateApproveList` checks if students have x amount of approved submissions by creating submission with preset status. This is also done in `newTestRunSubmission` and `newManualReviewSubmission`, by getting previous grades; `Grades:       previous.GetGrades(),`.

The `TestCreateApproveList` therefore fails when you move `s.Grades = make([]*Grade, len(userIDs))` outside of the group if statement, since this avoids resetting the grades of students submissions.

I have conluded to only initiliaze grades of they are nil, otherwise you would overrite them..

JoachimTislov:
By previous comment, should every group submissions grades be overwritten and checked again if they should be approved or not. 

Im not sure if its more desirable to reset and check if the new submission should have grades which are approved, but currently wouldn't that make sense, because the grades from the previous submissions are intialized in a new submission.

JoachimTislov:
The tests fail because of an outdated docker test, this is resolved in separate PR #1239: [commit](https://github.com/quickfeed/quickfeed/pull/1239/commits/8413a609b0a975e59f9bcfad7879f5834ae98600)

meling (approved):

Issue #1211, opened 2025-02-07
Env variable `QUICKFEED_APP_KEY` is updated to absolute path in manifest


JoachimTislov:
The predefined value of `QUICKFEED_APP_KEY` is equal to `$QUICKFEED/internal/config/github/quickfeed.pem`, but is in my case update to `/home/yoa/quickfeed/internal/config/github/quickfeed.pem`. 

This causes issues inside of the docker container, because the path is invalid.

Solution is to remove the appKey in [envToUpdate](https://github.com/quickfeed/quickfeed/blob/master/web/manifest/manifest.go#L158) inside of the `coversion` method in manifest.go

Updates to this:

```go
envToUpdate := map[string]string{
	appID:         strconv.FormatInt(config.GetID(), 10),
	clientID:      config.GetClientID(),
	clientSecret:  config.GetClientSecret(),
	webhookSecret: config.GetWebhookSecret(),
}
```

PR #1212, opened 2025-02-07
Remove modification of env variable `QUICKFEED_APP_KEY`

JoachimTislov:
Closes: #1211

### DIFF
--- a/web/manifest/manifest.go
+++ b/web/manifest/manifest.go
@@ -20,7 +20,6 @@ import (
 
 const (
 	appID         = "QUICKFEED_APP_ID"
-	appKey        = "QUICKFEED_APP_KEY"
 	clientID      = "QUICKFEED_CLIENT_ID"
 	clientSecret  = "QUICKFEED_CLIENT_SECRET"  // skipcq: SCT-A000
 	webhookSecret = "QUICKFEED_WEBHOOK_SECRET" // skipcq: SCT-A000
@@ -158,7 +157,6 @@ func (m *Manifest) conversion() http.HandlerFunc {
 		// Save the application configuration to the envFile
 		envToUpdate := map[string]string{
 			appID:         strconv.FormatInt(config.GetID(), 10),
-			appKey:        appKeyFile,
 			clientID:      config.GetClientID(),
 			clientSecret:  config.GetClientSecret(),
 			webhookSecret: config.GetWebhookSecret(),


web/manifest/manifest.go
```diff
@@ -158,7 +157,6 @@ func (m *Manifest) conversion() http.HandlerFunc {
 		// Save the application configuration to the envFile
 		envToUpdate := map[string]string{
 			appID:         strconv.FormatInt(config.GetID(), 10),
-			appKey:        appKeyFile,
```
meling, line 161: The `appKeyFile` is obtained via `env.AppKey()` and used to create the directories and save the PEM file. Issue #1211 claims that the path in `appKeyFile` is `$QUICKFEED/internal/config/github/quickfeed.pem`. Is that right?

If so, does it work within the docker container to save the PEM file? I didn't think that `os.WriteFile` would work with environment variables as part of its path... What happens to the PEM file in the docker container? This could also be a problem outside of a container, but since we most often don't run with a new .env file without the relevant expansion, we may not have noticed.

JoachimTislov: > The appKeyFile is obtained via env.AppKey() and used to create the directories and save the PEM file. Issue https://github.com/quickfeed/quickfeed/issues/1211 claims that the path in appKeyFile is $QUICKFEED/internal/config/github/quickfeed.pem. Is that right?

Yes

> If so, does it work within the docker container to save the PEM file? I didn't think that os.WriteFile would work with environment variables as part of its path... What happens to the PEM file in the docker container? 

Yes, I would assume so, but this is not configured in the [docker PR](#1194). The PEM file is currently not being updated in the docker container, because a github app is required to develop inside of the docker container. Creating an app in the docker container can be configured by having an if statement which checks for the given error `missing application ID for provider github` and run `quickfeed -dev -new`, but this would complicate the script file.

>This could also be a problem outside of a container, but since we most often don't run with a new .env file without the relevant expansion, we may not have noticed.

I am not sure if I fully understand your concern, but the path `$QUICKFEED/internal/config/github/quickfeed.pem` is relative to the env variable `$QUICKFEED's` definition, and overriding the variable to the actual path with for example `/home/meling/quickfeed/` will be problematic when you use it in a docker container because its wrong. Both home and meling directories does not exist in the isolated environment. The working directory is set to `/quickfeed`, a volume with a direct mount is configured to reflect changes done in the repository. `$QUICKFEED` will be an empty string and the complete path will be `/quickfeed/internal/config/github/quickfeed.pem`.

I think we resolved most of your concerns ealier today, I just wanted to clarify and answer your comment.

JoachimTislov:
The issue in the docker container arises in `NewSCMConfig` called from `main.go - main`; `scmConfig, err := scm.NewSCMConfig()`, because the app key or rather path to the key is retrieved there.

```go
appKey := env.AppKey()
createAppKey, err := key.FromFile(appKey)
if err != nil {
	return nil, fmt.Errorf("error reading key from file: %w", err)
}
appConfig, err := app.NewConfig(appID, createAppKey)
if err != nil {
	return nil, fmt.Errorf("error creating GitHub application client: %w", err)
}
```

JosteinLindhom (approved):

meling:
Superseded by #1485

Issue #1210, opened 2025-02-07
Creating a new development app with the secret flag is not possible


JoachimTislov:
```sh
% quickfeed -dev -new -secret
2025/02/07 22:43:29 Generating new random secret for signing JWT tokens...
2025/02/07 22:43:29 Loading environment variables from /home/yoa/quickfeed/quickfeed/.env
2025/02/07 22:43:29 Starting QuickFeed on https://localhost:443
2025/02/07 22:43:29 /home/yoa/quickfeed/quickfeed/.env.bak already exists; check its content before removing and try again
```

The issue lies inside of `Save`, it renames the old `.env` file to `.env.bak` before updating the .env file:

```go
func Save(filename string, env map[string]string) error {
	// Load the existing file's content before renaming it.
	content := load(filename)
	bakFilename := filename + ".bak"
	if exists(bakFilename) {
		if err := os.Remove(bakFilename); err != nil {
			return err
		}
	}
	if err := os.Rename(filename, bakFilename); err != nil {
		return err
	}
	// Update the file with new environment variables.
	return update(filename, content, env)
}
```

A solution would be to bypass the creation of the `.env.bak` file, by adding a parameter of wanting to create a backup of the `.env`, like this:

```go
func Save(filename string, env map[string]string, performBackup bool) error {
	// Load the existing file's content before renaming it.
	content := load(filename)
	if performBackup {
		bakFilename := filename + ".bak"
		if exists(bakFilename) {
			if err := os.Remove(bakFilename); err != nil {
				return err
			}
		}
		if err := os.Rename(filename, bakFilename); err != nil {
			return err
		}
	}
	// Update the file with new environment variables.
	return update(filename, content, env)
}
```

meling:
This is working as intended; see the output message:
```
2025/02/07 22:43:29 /home/yoa/quickfeed/quickfeed/.env.bak already exists; check its content before removing and try again
```
This is a safety precaution. It is meant to prevent accidentally overwriting of the `.env` file. You must manually remove the `.env.bak` file if you intended to overwrite the `.env` file.

It is not the intention that quickfeed should be started with the `-new` and `-secret` flags every time; they are meant to be used once and the `.env` file should be reused thereafter.

JoachimTislov:
> This is a safety precaution. It is meant to prevent accidentally overwriting of the `.env` file. You must manually remove the `.env.bak` file if you intended to overwrite the `.env` file.

[...]

PR #1213, opened 2025-02-08
Make `new` and `secret` flags compatible

JoachimTislov:
The case of wanting to both create a new github app and generate a secret was restricted because the modification of the env variable `QUICKFEED_AUTH_SECRET` with the save function creates the backup file: `.env.bak`, happens before creating the app.

The restriction of creating an app when the `.env.bak` file exists is intentional, as Hein mentioned in the related issue. But is in this case problematic since when a user wants to create a new app with a new secret its not possible because the secret generation occurs before the instance of creating the app.

The secret is required to be generated before the [NewTokenManager](https://github.com/quickfeed/quickfeed/blob/master/main.go#L117), because it is used when creating the [Token manager](https://github.com/quickfeed/quickfeed/blob/master/web/auth/jwt.go#L41)

The auth secret is required when a user doesn't use the `secret` flag

Resolves: #1210

### DIFF
--- a/internal/env/auth_secret.go
+++ b/internal/env/auth_secret.go
@@ -1,9 +1,24 @@
 package env
 
-import "os"
+import (
+	"os"
+
+	"github.com/quickfeed/quickfeed/internal/rand"
+)
+
+const authSecret = "QUICKFEED_AUTH_SECRET" // skipcq: SCT-A000
 
 // AuthSecret returns the JWT signing secret obtained from
 // the QUICKFEED_AUTH_SECRET environment variable.
 func AuthSecret() string {
-	return os.Getenv("QUICKFEED_AUTH_SECRET")
+	return os.Getenv(authSecret)
+}
+
+// NewAuthSecret generates a new random JWT signing secret,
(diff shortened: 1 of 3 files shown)


JosteinLindhom:
This issue seems to be caused by calling `env.Prepared(envFile)` in manifest.go on line 40.
`env.Prepared` will return an error if a backup already exists.

Moving the secret `Save` below the app creation stuff looks like it should work fine, but you need to load the env file again.

My understanding is as follows:
1. We need to `env.Load` the `.env` file early to make sure we have the correct domain for our GitHub app
2. We need to `env.Load` the `.env` file after app conversion, which is handled by `StartAppCreationFlow`
3. We need to `env.Load` the `.env` file after setting a new secret

1 always needs to happen. 
2 needs to happen if `-new` is set, and depends on 1.
3 needs to happen if `-secret` is set, but does not depend on the others as it indicates that a new secret should be set regardless of what 1 might populate `QUICKFEED_AUTH_SECRET` with.

meling (approved):

Issue #1215, opened 2025-02-08
suggestion: Add support for github app creation without `.env` files


JoachimTislov:
Background: 
- The implementation of a [docker container for development](#1194) sparked my interest in reducing the amount work it takes for a developer to start developing with an active quickfeed application locally.  

- I made a simple make tarket for generating the files and executing the `quickfeed -dev -new`, refer to [discussion](https://github.com/quickfeed/quickfeed/pull/1194#discussion_r1933440975). I later removed this target since it lacked the support for users who wants to create an app for production.
  - The issue occurs when copying the root `.env-template` without updating both env variables `DOMAIN` and `QUICKFEED_WHITELIST` which are required when constructing an app for production. 

- Recently we integrated the [secret generation](#1205) into quickfeed by adding the `secret` flag. This reminded me of the possibility of creating a new flag which can support both users. 
  - The `DOMAIN` flag can be utilized to allow users to specify the domain and prevents them from having to create both `env` files.
 
Implementation - **A maintainable and flexible solution**
- Handle the case when either `.env` files doesn't exist by creating them. 
- The env variable `DOMAIN` will by default be updated if its `example.com`, and `QUICKFEED_WHITELIST` will only be updated if the `dev` flag is not specified and the `DOMAIN` is not localhost. If the `DOMAN` flag is localhost and the env variable `DOMAIN` is localhost/example.com will we require the user to define a domain with the flag or update the `DOMAIN` manually.
- The `DOMAIN` flag will by default be localhost
- Adding domain validation would be reasonable, to prevent invalid domain.
   - Validation will relate to what mode quickfeed is executed in, dev equal local and prod equal public. 
   - Basic validation is also needed to prevent false updates
 
Update:

Sort of related to this is the test environment deployment, refer to [Preparing the enviroment for testing](https://github.com/quickfeed/quickfeed/blob/master/doc/deploy.md#preparing-the-environment-for-testing). Adding a flag for this would be helpful

JoachimTislov:
Can also add a `force` flag which ignores certain constraints. For example, creating a new app is restricted if the `.env.bak` exist or if app ID is defined. Using this flag can bypass these errors and allow the user to forcefully create an app.

meling:
[...]

PR #1219, opened 2025-02-17
Joachim tislov/issue1215

JoachimTislov:
Resolves: #1215


JoachimTislov:
I can add tests, if this is desirable... @meling

JoachimTislov:
This PR is now update to date with the corresponding issue's latest discussion.

Some of the features: (will update later)

- Secret generation is now executed and the program is existed afterwards.
- The secret is now generated if its empty
- The app creation flow is started if the App ID is empty
- The user is asked for confirmation when creating an app on a local domain, the app information exists in corresponding env file, given the environment and for both backup files. Im considering adding a way of either coping the backfiles or skip overiting given backup files. This opens up for a more flexible intercation, rather than requiring the user to this manually.

PR #1220, opened 2025-02-21
Docker container for development

### LINKED ISSUE #360: Create/set up docker container for development

JoachimTislov:
This pr continues #1194, and contains the following:

- Docker environment for development
  - A `docker-compose.yml` file was created to ease the start up experience, the command `docker-compose up` should create both the image and a container as long as a github app has been created, which is specified in the documentation.
  - The docker image is built on top of `1.23:alpine` and installs the required depencies to run Quickfeed. Air is installed separately because its not a direct dependency of Quickfeed and is only used in development.
  - A volume with bind mount has been configured to reflect the changes made in the repository. Air will not work without this configuration. I chose a volume instead of a bind mount to persist the configuration, this is not an issues with `docker-compose` since you can configure a bind mount there. Its useful when running Quickfeed with `docker daemon`, the docker desktop does not provide a volume setting for some reason..
  - [Air-verse](https://github.com/air-verse/air) is used for hot reloading the backend
    - A script was made to improve visibility over commands which are executed after each time a file is saved, this would alternatively be a long white string in the `.air.toml` file in the `cmd` field.
    - The public folder along with a few other folders are excluded to prevent false updates. Air will only watch `go` files in the backend. Air executes quickfeed with both `dev` and `watch` flag to run in development and hot reload the frontend, which was configured in this pr: #1201 
  - The related documentation is written in the docker-dev.md file located in the `doc` subfolder
- A simple launch profile to execute the quickfeed binary in debug mode
- Removed the localhost constraint, this was an issue on mac but has been fixed

This warning is addressed in the documentation: 

`WARN[0000] The "QUICKFEED" variable is not set. Defaulting to a blank string.`

The reason for this warning is the .env file which is located in the same directory (root). The solution is to either 
- Move the docker files to a different directory
- Enclose the values of env variables which use the `$QUICKFEED` variable in a nested string
- Execute this command `echo 'export QUICKFEED=' >> $HOME/.bashrc && source $HOME/.bashrc`
- Run docker command with the definition: `QUICKFEED=""`. 

Nested string explanation:

This `QUICKFEED_CERT_FILE=$QUICKFEED/internal/config/certs/fullchain.pem` should be updated to this `QUICKFEED_CERT_FILE="$QUICKFEED/internal/config/certs/fullchain.pem"`

This can be done with `fmt.Sprintf("\"%s\"", *variable*)`

[Helpful resource for dealing with the warning](https://forums.docker.com/t/warn-0000-the-he-variable-is-not-set-defaulting-to-a-blank-string/137212)

Closes: #360


JoachimTislov:
After building the image and creating a container, you simply need to start it and call the server by entering the url in the browser, thereby activating the watcher, which will refresh the browser when updating the Frontend files. 

Air-verse is quite nice as it will recompile QuickFeed after each change in Go/backend code and execute the new binary. I've tested this by commenting out the `GetUser` endpoint and reverting the change. The server crashes, rebuilds itself and starts up the server.

Though the setup of the docker container might be challenging, afterwards you no longer need to manage your own QuickFeed instance, everything will be handled by the container :smiley:

Issue #1195, opened 2025-01-28
chore/bug:  Update the redirect url for the github app when the http.addr flag is used


JoachimTislov:
Github app redirect url will be outdated if the http.addr is set to different port. Everything is set when creating the app and redirect url defaults to the port 443

```go
func GetCallbackURL(baseURL string) string {
	return getURL(baseURL, Callback)
}
```
Creates the intial callback url, and updating would require a similar implementation like creating the app in `manifest.go`

meling:
I don't understand the problem, could you elaborate?

Are you saying that passing `-http.addr` with another port like `:9000`, then the GitHub app won't work because the callback URL always uses `:443`?

I think the suggested function looks fine, but I haven't looked at the details.

JoachimTislov:
> Are you saying that passing `-http.addr` with another port like `:9000`, then the GitHub app won't work because the callback URL always uses `:443`?

Yes

> I think the suggested function looks fine, but I haven't looked at the details.

There a slight issue with it because the baseUrl does not take into fact that the port is 443, in development its most likely just 127.0.0.1 or localhost, so the port is also needed if you want to redirect to a different port.

A solution would be updating the base url to this (in development i presume)
```go
BaseUrl = fmt.Sprintf("%s%s", domain, portNumber)
```

meling:
This sounds easy to fix. Please open a PR.

JoachimTislov:
> This sounds easy to fix. Please open a PR.

Easy how? The fix includes adding a way of dynamically updating the redirect url of the github app when ran in development with the http.addr flag.

Requires extracting the current domain in the env file, combininig it with the provided port and send a post request to the github app which needs to be done in the browser (same as when creating the app), not possible through an API because the user does not provide a key. 

The last part will be difficult I think.. unless theres no demand of user input after succesfully updating the redirect url. 

@JosteinLindhom Thoughts?

JoachimTislov:
Theres multiple ways of solving this certain issue:

1. Create a new app with different redirect url, since Github does not support updating the redirect url through their API
    - This requires modifying `GetCallbackUrl` to include the port number
    - This solution ensures a stable callback, but is time consuming
2. Configure external or internal proxy to either forward or reverse the connection    
    - Handling the case where the `http.addr` flag is used and it differs from port `:443`, by settting up a  proxy. The redirect url when a github app is created will always be on port :443, since its the default port for https.
[...]

PR #1222, opened 2025-02-21
Resolve http.addr bug and replace http.addr flag with env variable PORT

JoachimTislov:
Closes: #1195 

The baseUrl for the redirect url on `master branch` does not include the http address. I made a simple method which combines the domain and the http address. Default callback url will thereby be: `https://localhost:443/auth/callback/` which is the same as `https://localhost/auth/callback/`

`https://uis.itest.run/` runs on the default https port and the port is therefore hidden. If you add the port `443` and enter the url: `https://uis.itest.run:443/`, will the port number be removed. Attempting to connect on a different port will fail unless Quickfeed is active there as well.

A user of Quickfeed can now specify their own port to run the Quickfeed server on, the default port is 443, and you can possibly ignore the port when its the default https port, but doing so will most likely cause issues when creating the server, as the http address would then be `:`.

The port will now persist over multiple instances rather then resetting to `443` when the `http.addr` flag is not used.

### DIFF
--- a/doc/deploy.md
+++ b/doc/deploy.md
@@ -6,6 +6,7 @@
   - [Setup](#setup)
     - [Install Tools for Deployment](#install-tools-for-deployment)
     - [Install Tools for Development](#install-tools-for-development)
+    - [The PORT environment variable](#the-port-environment-variable)
     - [Preparing the Environment for Production](#preparing-the-environment-for-production)
     - [Preparing the Environment for Testing](#preparing-the-environment-for-testing)
     - [First-time Installation](#first-time-installation)
@@ -63,6 +64,14 @@ To install:
 
 The `devtools` make target will download and install various Protobuf compiler plugins and the grpcweb Protobuf compiler.
 
+### The PORT environment variable
+
+The `PORT` environment variable can be changed to any valid port, and we strongly advise running Quickfeed a secure port in production.
+It can be set to an unprivileged port and prevent permission issues when [running on a privileged port](#running-server-on-a-privileged-port).
+This is helpful for debugging, as it allows you to run Quickfeed directly by setting program to `"${workspaceFolder}/main.go"`, in the launch profile.
+Alternatively, you can grant the Quickfeed binary access to the privileged port and run it as an executable.
+The launch profile program should then be set to `"${env:GOPATH}/bin/quickfeed"`.
+
 ### Preparing the Environment for Production
 
 QuickFeed expects the `.env` file to contain certain environment variables.
@@ -80,6 +89,8 @@ QUICKFEED_WEBHOOK_SECRET=""
 
 # QuickFeed server domain or ip
 DOMAIN="example.com"
+# Quickfeed server port
+PORT="443"
 
 # Comma-separated list of domains to allow certificates for.
 # IP addresses and "localhost" are *not* valid.
@@ -103,6 +114,8 @@ QUICKFEED_CERT_FILE=$QUICKFEED/internal/config/certs/fullchain.pem
 
 # QuickFeed server domain or ip
 DOMAIN="127.0.0.1"
+# Quickfeed server port
+PORT="443"
 ```
 
 The [QuickFeed App installation process](#first-time-installation) will guide you through the rest of the setup,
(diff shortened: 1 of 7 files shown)


JosteinLindhom:
We should update the doc to reflect the updated `.env-template`
https://github.com/quickfeed/quickfeed/blob/master/doc/deploy.md#preparing-the-environment-for-production

Otherwise it looks good.

web/manifest/manifest.go
```diff
@@ -50,7 +50,7 @@ func ReadyForAppCreation(envFile string, chkFns ...func() error) error {
 }
 
 func CreateNewQuickFeedApp(srvFn web.ServerType, httpAddr, envFile string) error {
```
meling, line 52: Should this line in `StartAppCreationFlow` below:
```go
log.Printf("Go to https://%s/manifest to install the QuickFeed GitHub App.\n", env.Domain())
```
also use `DomainWithPort()`?

meling: Also, this function has a test in `web/manifest/manifest_test.go` that calls it like this:
```go
if err := manifest.CreateNewQuickFeedApp(web.NewDevelopmentServer, ":443", envFile); err != nil {
```
Should this test be adjusted?

main.go
```diff
@@ -70,13 +69,13 @@ func main() {
...
 	}
-	log.Printf("Starting QuickFeed on %s%s", env.Domain(), *httpAddr)
+	log.Printf("Starting QuickFeed on %s", env.DomainWithPort())
 
 	if *newApp {
 		if err := manifest.ReadyForAppCreation(envFile, checkDomain); err != nil {
 			log.Fatal(err)
 		}
-		if err := manifest.CreateNewQuickFeedApp(srvFn, *httpAddr, envFile); err != nil {
+		if err := manifest.CreateNewQuickFeedApp(srvFn, env.HttpAddr(), envFile); err != nil {
```
meling, line 78: I suggest that we simplify the `web.ServerType` like this:
```go
type ServerType func(handler http.Handler) (*Server, error)
```
And instead call `env.HttpAddr()` in the different handlers?

That would simplify stuff and avoid certain mistakes.

meling (changes requested):

main.go
```diff
@@ -122,7 +121,7 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	authConfig := auth.NewGitHubConfig(env.Domain(), scmConfig)
+	authConfig := auth.NewGitHubConfig(env.DomainWithPort(), scmConfig)
```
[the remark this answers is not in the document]
JoachimTislov: > Not sure which of these should pass/fail, but that's easy to configure in a table-driven test.

Everything will pass. Im not sure how they would fail. Everything related to the `RedirectURL` is deterministic and there isn't anything that could go wrong. A string is sent into `NewGitthubConfig` as `baseURL` and the `RedirectURL` becomes - `https://*string*/auth/callback`.

Its are only checking `if string == string`, right? Or am I understanding this incorrectly?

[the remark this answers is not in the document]
meling: I meant to actually start the `auth/callback` endpoint with different URLs, as we do it here in `main.go` ... and `webserver.go`, see these lines in `RegisterRouter`:
```go
...
router.HandleFunc(auth.Auth, auth.OAuth2Login(s.logger, authConfig, callbackSecret))
router.HandleFunc(auth.Callback, auth.OAuth2Callback(s.logger, s.db, tm, authConfig, callbackSecret))
```
Does this make sense to you @JosteinLindhom? Or is the `auth/callback` only relevant when GitHub calls us back during the auth process and thus it would only work when running in production?

doc/deploy.md
```diff
@@ -91,6 +93,8 @@ You only need to edit the `$DOMAIN` environment variable to point to your public
 The [QuickFeed App installation process](#first-time-installation) will guide you through the rest of the setup,
 setting the environment variables in your `.env` file and saving the `quickfeed.pem` file.
 
+The `PORT` environment variable can be changed to any valid port and could be used to avoid the server from [running on a privileged port](#running-server-on-a-privileged-port). This helpful for when you want to run Quickfeed in debug mode, since you can't input your sudo password into the debug console. A possible workaround to run Quickfeed on privileged port in debug mode is to set capabilities of the  binary you wish to debug with and then start debugging, but keep in mind switching from a privileged port will allow you to run Quickfeed directly without creating a binary - `go run $QUICKFEED/main.go`.
```
meling, line 96: One sentence per line, please.
There are also language issues (missing words).
I don't think I would frame it as a "possible workaround"; it is an alternative approach.
The last sentence is also too long.
Suggest to ask GPT to check the language and write it more clearly.

Finally, note that there is another section with the flags that are still referring to the now removed `http.addr` flag and a section titled `Running Server on a Privileged Port` that also needs to be updated accordingly.

meling (changes requested):
Docs needs fixing

main.go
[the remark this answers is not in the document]
JoachimTislov: > auth/callback only relevant when GitHub calls us back during the auth process

I think this correct.

meling (approved):
Looks good to me!

Issue #818, opened 2022-10-06
UpdateFromTestsRepo uses a single mutex for all updates


meling:
This may delay processing for courses that update their `tests` repo concurrently. We should consider adding a mutex per course, to allow concurrency per course.

JoachimTislov:
> This may delay processing for courses that update their `tests` repo concurrently. We should consider adding a mutex per course, to allow concurrency per course.

Yes this can delay processing, the `updateMutex` variable is to my understanding global and stored on the heap, which will only allow one update to happen at time.

The function `UpdateFromTestsRepo` is executed in these two cases:
- A teacher pushes an update to the tests repository, the Github Webhook then triggers a creation of a `go routine`, and updates the semaphore limit, which is initialized as `const maxConcurrentTestRuns = 5`
    
  ```go
  go func() {
  	wh.sem <- struct{}{} // acquire semaphore
  	wh.handlePush(e)
  	<-wh.sem // release semaphore
  	// remove commitID from duplicate map (to avoid memory leak)
  	wh.dup.Remove(commitID)
  }()
  ```
- A teacher can also request an update in Quickfeed's web interface;
![Image](https://github.com/user-attachments/assets/07383147-94c9-49e4-b9d2-111913d4d78c)
This will be call the [`UpdateAssignments`](https://github.com/quickfeed/quickfeed/blob/master/web/quickfeed_service.go#496) API, which will run `UpdateFromTestsRepo` in the main thread.

If teachers sends multiple requests to Quickfeed by either the API or through the Github webhook, will it result in a queue and the response for teacher using the API will be delayed. The `push` for tests repositories will eventually be handled.

That said, theres a limited amount of courses and the case of multiple teachers wanting to update their tests is rare, but configuring a `mutex` per course is reasonable, and will improve response time.

PR #1223, opened 2025-03-01
Add mutex for each course rather then using a common update mutex

JoachimTislov:
This implementation is quite complex. The courseMutexMap is stored in memory, on the heap and updated with the mapMu mutex since it will be accessed concurrently. The unlock method is questionable because it does not handle the case of the mutex not existing, but this will be considered as a developer mistake. Lock should always be called before unlock.

Closes: #818

### DIFF
--- a/qf/course.go
+++ b/qf/course.go
@@ -35,15 +35,16 @@ func (course *Course) UpdateDockerfile(dockerfile string) bool {
 }
 
 var (
-	// Mutex for each course
+	// courseMutexMap contains a single mutex for each course.
 	courseMutexMap = make(map[uint64]*sync.Mutex)
-	// Mutex for the course mutex map
+	// mapMutex protects the courseMutexMap.
 	mapMutex = sync.Mutex{}
 )
 
-// Lock indexes the course mutex map with the course ID and locks the mutex.
-// The mutex is initialized if it does not exist.
-// This method is called when concurrently accessing the course.
+// Lock locks the course to prevent concurrent updates to the course.
+// It returns a corresponding unlock function which must be called when the update is done.
+// Specifically, this method is used to prevent concurrent database updates
+// derived from the test repository. See [assignments.UpdateFromTestsRepo].
 func (course *Course) Lock() func() {
 	mapMutex.Lock()
 	if _, ok := courseMutexMap[course.GetID()]; !ok {
@@ -78,7 +79,7 @@ func (course *Course) CloneDir() string {
 }
 
 func (course *Course) TeacherEnrollments() []*Enrollment {
-	enrolledTeachers := []*Enrollment{}
+	var enrolledTeachers []*Enrollment
 	for _, enrollment := range course.Enrollments {
 		if enrollment.IsTeacher() {
 			enrolledTeachers = append(enrolledTeachers, enrollment)
--- a/qf/course.go
+++ b/qf/course.go
@@ -34,35 +34,26 @@ func (course *Course) UpdateDockerfile(dockerfile string) bool {
 	return updated
 }
 
-// Mutex for each course
 var (
-	courseMuMap = make(map[uint64]*sync.Mutex)
-	mapMu       = sync.Mutex{}
+	// Mutex for each course
+	courseMutexMap = make(map[uint64]*sync.Mutex)
+	// Mutex for the course mutex map
+	mapMutex = sync.Mutex{}
 )
 
 // Lock indexes the course mutex map with the course ID and locks the mutex.
 // The mutex is initialized if it does not exist.
 // This method is called when concurrently accessing the course.
-func (course *Course) Lock() {
-	mapMu.Lock()
-	if _, ok := courseMuMap[course.ID]; !ok {
-		courseMuMap[course.ID] = &sync.Mutex{}
+func (course *Course) Lock() func() {
+	mapMutex.Lock()
+	if _, ok := courseMutexMap[course.GetID()]; !ok {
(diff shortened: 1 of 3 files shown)


meling:
See https://go.dev/blog/synctest

qf/course.go
```diff
@@ -33,6 +34,37 @@ func (course *Course) UpdateDockerfile(dockerfile string) bool {
 	return updated
 }
 
+// Mutex for each course
+var (
+	courseMuMap = make(map[uint64]*sync.Mutex)
```
meling, line 39: rename to `courseMutexMap`

JoachimTislov: Will do

qf/course.go
```diff
@@ -33,6 +34,37 @@ func (course *Course) UpdateDockerfile(dockerfile string) bool {
...
+	mapMu       = sync.Mutex{}
+)
+
+// Lock indexes the course mutex map with the course ID and locks the mutex.
+// The mutex is initialized if it does not exist.
+// This method is called when concurrently accessing the course.
+func (course *Course) Lock() {
```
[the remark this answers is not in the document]
JoachimTislov: Yes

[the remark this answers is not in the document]
JoachimTislov: @meling Write a test with the package or without?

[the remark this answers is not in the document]
meling: Without at first; let’s see what it ends up looking like.

qf/course_test.go
```diff
@@ -44,6 +47,64 @@ func TestUpdateDockerfile(t *testing.T) {
...
 }
 
+func TestLock(t *testing.T) {
+	var wg sync.WaitGroup
+
+	course := &qf.Course{ID: 1, CourseCreatorID: 0}
+	want := uint64(10)
+	rang := 5
+
+	for i := range rang {
```
meling, line 57: I found it a bit confusing at first that you are creating 5 goroutines, yet the `want` variable is 10. I think it will be more intuitive and easier to understand if you do this:
```go
for range want {
	wg.Add(1)
	go func() {
		defer wg.Done()

		unlock := course.Lock()
		defer unlock()

		course.CourseCreatorID++
	}()
}
```

qf/course_test.go
```diff
@@ -44,6 +47,64 @@ func TestUpdateDockerfile(t *testing.T) {
...
+	ch := make(chan struct{})
+	go func() {
+		wg.Wait()
+		ch <- struct{}{}
+	}()
+
+	select {
+	case <-ch:
+		t.Errorf("Lock failed to prevent concurrent access")
+	case <-time.After(1 * time.Second):
```
meling, line 98: 1 second is too long. 10 ms should suffice, I think. If it is flaky in CI, we can adjust it higher. And if/when the synctest experiment is accepted it shouldn't be an issue.

qf/course_test.go
```diff
@@ -44,6 +47,64 @@ func TestUpdateDockerfile(t *testing.T) {
...
+
+	// Ensure the lock works as expected
+	// Run two goroutines, expects only one update, because the lock is not released.
+
+	want = uint64(0)
+	for i := range 2 {
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			_ = course.Lock()
```
meling, line 84: Note that this test can't be run more than once.
```sh
go test -v -run TestLock -count 2
=== RUN   TestLock
--- PASS: TestLock (1.00s)
=== RUN   TestLock
^Csignal: interrupt
FAIL	github.com/quickfeed/quickfeed/qf	5.745s
```
I tried doing:
```go
	unlock = course.Lock()
	t.Cleanup(unlock)
```
here, but it didn't help. I'm not sure why.

Note that the second run of the test will find a course mutex in the map for the course (since it is a global variable)... which shouldn't be a problem if the course is unlocked at the end of this test. Don't have time to debug this.

meling (changes requested):
There is a deadlock issue if you run the test more than once.

qf/course_test.go
JoachimTislov: > (since it is a global variable)...

Yes, but I thought that didn't matter. I thought every test started with default variables.

I will solve it.

qf/course.go
```diff
@@ -33,6 +34,28 @@ func (course *Course) UpdateDockerfile(dockerfile string) bool {
...
 }
 
+var (
+	// Mutex for each course
+	courseMutexMap = make(map[uint64]*sync.Mutex)
+	// Mutex for the course mutex map
+	mapMutex = sync.Mutex{}
+)
+
+// Lock indexes the course mutex map with the course ID and locks the mutex.
```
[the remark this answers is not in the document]
JoachimTislov: Got it

qf/course_test.go
JoachimTislov: Understandable, I couldn't really think of simple way of doing it in the mist of writing a go routine test which I've never done before.

qf/course_test.go
JoachimTislov: Agree

qf/course_test.go
JoachimTislov: Oh you are running with the same global variable when executing: `go test -v -run TestLock -count 2`

JoachimTislov:
> There is a deadlock issue if you run the test more than once.

Beware, this is only an issue when using the same global variable. The test is working as intended. I want to confirm that a deadlock happens and realese the lock a bit later.

meling (approved):

Issue #1180, opened 2025-01-08
perf(test-runner/go): pre-fetch tests dependencies and initialize assignment scores 


JosteinLindhom:
There are some potential improvements we can do to decrease the time spent running tests. 

We could pre-fetch test dependencies in Docker images, e.g. in Go we could try to cache the `kit` module when building the image

We could pre-run tests when changes are made to either the `tests` or `assignments` repository to generate template score objects.
These object could be mounted into testing containers and injected with the required secret to initialize scores.
This way we can avoid running tests in the assignments repo to initialize scores on every test run.

This issue description needs to be reformulated to make more sense ^

meling:
While fresh in my memory, let me add that we could pre-run some variant of the following command for every update/push to the `tests` repository. Let me name the script `update-tests.sh`:
```sh
SCORE_INIT=1 go test -v ./... 2>&1 | grep TestName
```
We can extract the test names and score information from this and save it in the database. Then, for every CI run, we don't have to worry about student code failing tests early due to panic and calculating incorrect total scores since we know the `MaxScore` and `Weight` from the saved information.

I thought a bit about where we could store this information, and it seems the `Assignment` type (`types.proto`) might be the best option:
```proto
repeated score.Score ScoreInfo = 14;  // list of tests with score information (MaxScore and Weight)
```
I'm unsure about the `ScoreInfo` name; it's the placeholder name for now.

I'm also not sure if this (being part of `Assignment`) would get sent to the frontend unnecessarily, but we could **clean** the field in the outgoing interceptor. 

The full version of `update-tests.sh` for Go courses could look like the initialization step of our current `run.sh` script:
```sh
printf "*** Initializing Tests for %s ***\n" "$CURRENT"

# Move to folder with assignment handout code for the current assignment to test.
cd "$ASSIGNMENTS/$CURRENT"
# Remove assignment handout tests, if any, to avoid interference, but keep quickfeed tests.
find . \( -name '*_test.go' -and -not -name '*_qf_test.go' \) -exec rm -rf {} \;

# Copy tests into the base assignments folder for initializing test scores
cp -r "$TESTS"/* "$ASSIGNMENTS"/

# $TESTS does not contain go.mod and go.sum: make sure to get the kit/score package
go get -t github.com/quickfeed/quickfeed/kit/score
go mod tidy
# Initialize test scores
SCORE_INIT=1 go test -v ./... 2>&1 | grep TestName
```
[...]

PR #1225, opened 2025-03-03
Inject default score templates for an assignment into the docker container, by pre-running tests on the assignments repository, on a push event

JoachimTislov:
Resolves: #1180 

This pr initialize assignment scores and is compatible with the current design, the `update-tests.sh` isn't executed if it does not exist in the `tests` repository. As for the pre-fetching of the kit module; 

This line in the docker file is not possible, because kit is a module, not a standalone package
```dockerfile
RUN go install github.com/quickfeed/quickfeed/kit@latest
```

I am therefore looking into how all of the depencies can be downloaded (pre-fetched), for example with an image like this:

```dockerfile
FROM golang:1.23-alpine

# Install bash and git (and build-base to get gcc)
# (this is required when building FROM: golang:alpine)
RUN apk update && apk add --no-cache git bash build-base docker openrc
RUN wget -O- -nv https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.41.1

WORKDIR /quickfeed

COPY go.mod go.sum ./

RUN go mod download
```
Currently struggling with building the image in the correct context, for it to detect/locate the go.mod and go.sum files.

This pr commit: "chore: Resolve issues" found 20 DeepSoucre issues. Great!


meling:
I only looked at this in passing, and I think this is touching too many parts in bad ways. I don't like it. We need to find another way. I will comment in the #1180. Leaving it open a bit longer, but I intend to close this.

Issue #1107, opened 2024-06-25
chore: do we need to use slug.Make(opt.Name)


meling:
We currently only use `slug.Make(opt.Name)` when looking up an organization (from the `scm`); at this point it might already be too late, since the organization has been created on GitHub and thus, whatever the user enters when looking up the course via our frontend, it would already have been sluggified and the user should probably be aware... 

Previously we also used `slug.Make` when looking up a repository, but this was not used when creating repositories and there could an argument made for doing the slugging before adding the repository name to the database as well.

@JosteinLindhom thoughts?

meling:
We do have similar functionality implemented for limiting what students can do for group names. The question is whether or not we could replace our group name logic with `slug.Make()` with lower case disabled.

JoachimTislov:
To my understanding does slug.Make convert norwegian charaters into english ones (æ -> ae, ø -> o, å -> aa). So doing this for repository and group names would be reasonable, but I don't see the issue and reasoning behind converting norwegian to english characters, does it cause issues? I don't think a string would affect anything... im not sure if I understand the problem here

Group name solution, not with this: `lower case disabled.`
file: scm_helpers.go:36
```go 
func createRepo(ctx context.Context, sc scm.SCM, course *qf.Course, group *qf.Group) (*qf.Repository, error) {
        groupName := slug.Make(group.GetName())
	opt := &scm.GroupOptions{
		Organization: course.ScmOrganizationName,
		GroupName:    groupName ,
		Users:        group.UserNames(),
	}
```

Repository name solution:
github.go:380
```go
repositoryName := slug.Make(opt.Repo)
repo, _, err = s.client.Repositories.Create(ctx, opt.Owner, &github.Repository{
	Name:    github.String(repositoryName),
	Private: github.Bool(opt.Private),
})
```

meling:
We have two uses of `slug.Make`:

1. In `scm/github_mock_opts.go`:
```go
var toOrg = func(course *qf.Course) github.Organization {
	return github.Organization{
		ID:    github.Int64(int64(course.ScmOrganizationID)),
		Login: github.String(slug.Make(course.ScmOrganizationName)),
	}
}
```
2. In `scm/github.go`:
```go
func (s *GithubSCM) GetOrganization(ctx context.Context, opt *OrganizationOptions) (*qf.Organization, error) {
...
	} else {
		name := slug.Make(opt.Name)
		githubOrg, _, err = s.client.Organizations.Get(ctx, name)
```
[...]

PR #1228, opened 2025-03-07
Slugify group name without lower case and student repo with lower case

JoachimTislov:
Resolves: #1107

### DIFF
--- a/scm/github.go
+++ b/scm/github.go
@@ -266,15 +266,18 @@ func (s *GithubSCM) CreateGroup(ctx context.Context, opt *GroupOptions) (*Reposi
 	if !opt.valid() {
 		return nil, E(op, m, fmt.Errorf("missing fields: %+v", *opt))
 	}
+	slug.Lowercase = false
+	groupName := slug.Make(opt.GroupName)
+	slug.Lowercase = true
 	if _, err := s.GetOrganization(ctx, &OrganizationOptions{Name: opt.Organization}); err != nil {
 		// organization must exist
 		return nil, E(op, m, err)
 	}
-	if _, err := s.getRepository(ctx, &RepositoryOptions{Owner: opt.Organization, Repo: opt.GroupName}); err == nil {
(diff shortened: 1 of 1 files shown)


scm/github.go
```diff
@@ -413,7 +416,7 @@ func (s *GithubSCM) createStudentRepo(ctx context.Context, organization string,
 	// if repo is found, it is safe to reuse it
 	repo, err := s.createRepository(ctx, &CreateRepositoryOptions{
 		Owner:   organization,
-		Repo:    qf.StudentRepoName(user),
+		Repo:    slug.Make(qf.StudentRepoName(user)),
```
meling, line 419: This is not needed.
```go
func StudentRepoName(userName string) string {
	return userName + StudentRepoSuffix
}
```
Where:
```go
	StudentRepoSuffix = "-labs"
```
The `userName` is already a GitHub user name, and so it cannot be an invalid repository name.

scm/github.go
```diff
@@ -266,15 +266,18 @@ func (s *GithubSCM) CreateGroup(ctx context.Context, opt *GroupOptions) (*Reposi
 	if !opt.valid() {
 		return nil, E(op, m, fmt.Errorf("missing fields: %+v", *opt))
 	}
+	slug.Lowercase = false
+	groupName := slug.Make(opt.GroupName)
```
meling, line 270: I don't want to change the group name entered by students in the frontend. We should prevent students from creating a group name that isn't a valid repository name. We should do it in the frontend.

meling (commented):
See my comments. None of these changes are desirable. Please close it; I will comment on the issue with suggestions.

meling:
Closing as not necessary. See [comment on #1107](https://github.com/quickfeed/quickfeed/issues/1107#issuecomment-2706545953).

JoachimTislov:
> See my comments. None of these changes are desirable. Please close it; I will comment on the issue with suggestions.

I understand.

scm/github.go
JoachimTislov: Oh right, I forgot.

Issue #1178, opened 2025-01-07
chore: Move settings back to .vscode/settings.json


meling:
As I mentioned in [issue #1174 (comment)](https://github.com/quickfeed/quickfeed/issues/1174#issuecomment-2576290029): 

We should move settings back to `.vscode/settings.json` now that `gopls` and the Go plugin support the `go.work` file. That way, we would avoid forgetting to start VSCode with the workspace file.

The success criteria: is VSCode able to work correctly (compile, lint, etc.) within both the main module and the `kit` module?

PR #1229, opened 2025-03-07
Move settings back to .vscode/settings.json

JoachimTislov:
Resolves: #1178 

> The success criteria: is VSCode able to work correctly (compile, lint, etc.) within both the main module and the kit module?

Yeah, it works for me.

Following was removed because its not a valid workspace configuration... Im not certain why.

```json
"cSpell.enableFiletypes": [
	"plaintext"
],

"eslint.alwaysShowStatus": true,
```


meling (approved):
Nice! Works for me too!

Issue #948, opened 2022-12-08
Chore: Remove UpdateSubmissionRequest message


meling:
We can remove the `UpdateSubmissionRequest` message and send a `Submission` message, as we do for `UpdateCourse` and friends.

meling:
A small snag: `Submission` does not have `CourseID`

meling:
The current messages could also be converted to just one message as below:
```proto
message UpdateSubmissionRequest {
    uint64 submissionID      = 1;
    uint64 courseID          = 2;
    uint32 score             = 3;
    bool released            = 4;
    Submission.Status status = 5;
}

message UpdateSubmissionsRequest {
    uint64 courseID     = 1;
    uint64 assignmentID = 2;
    uint32 scoreLimit   = 3;
    bool release        = 4;
    bool approve        = 5;
}
```
Like this:
```proto
// UpdateSubmissionRequest is used to update manually reviewed submissions.
message UpdateSubmissionRequest {
    uint64 CourseID          = 1;
    uint64 AssignmentID      = 2;  // if non-zero, update all submissions
    uint64 SubmissionID      = 3;  // if non-zero, update single specific submission
    uint32 Score             = 4;  // only used for single submission
    bool Release             = 5;  // indicate whether or not to release submission(s) to students
    Submission.Status Status = 6;
}
```

It is unclear if we need to use the `scoreLimit` field when submitting submissions for all students. To me, it makes more sense to let the server decide whether or not a submission's (recorded) score is higher than the assignment's score limit.

That said, these custom rules for the fields may be too convoluted and we should stick with something simpler. Or find a better option.

PR #1230, opened 2025-03-07
Merge UpdateSubmissionsRequest and UpdateSubmissionsRequest into a single message

JoachimTislov:
Closes: #948


qf/requests.proto
```diff
@@ -55,20 +55,15 @@ message SubmissionRequest {
...
-
-message UpdateSubmissionsRequest {
     uint64 courseID     = 1;
-    uint64 assignmentID = 2;
-    uint32 scoreLimit   = 3;
-    bool release        = 4;
-    bool approve        = 5;
+    uint64 assignmentID = 2; // if non-zero, update all submissions
+    uint64 submissionID = 3; // if non-zero, update single specific submission
+    uint32 score        = 4; // only used for single submission
```
JosteinLindhom, line 63: This could previously be used to approve all submissions with a score >= `score`, i.e. we could send a request with `assignmentID = 10, score = 90, approve = true` and it would approve all submissions for assignment 10 where score was >= 90.

qf/requests.proto
```diff
@@ -55,20 +55,15 @@ message SubmissionRequest {
...
-    repeated qf.Grade grades = 5;
-}
-
-message UpdateSubmissionsRequest {
     uint64 courseID     = 1;
-    uint64 assignmentID = 2;
-    uint32 scoreLimit   = 3;
-    bool release        = 4;
-    bool approve        = 5;
+    uint64 assignmentID = 2; // if non-zero, update all submissions
```
meling, line 61: Shouldn’t it be invalid to send a msg with assignmentID>0 and submissionID>0? That is, perhaps we should make the two fields oneof to prevent it, or what do you think? Would this solve your issue @JosteinLindhom ? That is, sending just `assignmentID = 10, score = 90` could be interpreted as approve all submissions for assignment 10 where score was >= 90. We should document it carefully though.

qf/requests.proto
meling, line 63: Is score better (more correct) than scoreLimit? Does it represent an actual score for a student’s student submission or is it the assignment’s score limit? Or is it both?

qf/validation.go
```diff
@@ -67,7 +67,7 @@ func (req *SubmissionRequest) IsValid() bool {
 
 // IsValid ensures that both submission and course IDs are set
 func (req *UpdateSubmissionRequest) IsValid() bool {
-	return req.GetCourseID() > 0 && req.GetSubmissionID() > 0
+	return req.GetCourseID() > 0 && (req.GetSubmissionID() > 0 || req.GetAssignmentID() > 0)
```
meling, line 70: See discussion above. Also, whatever the conclusion about the above, we should add a test for this.

buf.gen.ui.yaml
```diff
@@ -2,7 +2,7 @@
 # For details, see https://docs.buf.build/configuration/v1/buf-gen-yaml
 version: v1
 plugins:
-  - plugin: buf.build/bufbuild/es
+  - plugin: buf.build/bufbuild/es:v1.10.0
```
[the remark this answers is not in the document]
JoachimTislov: Yes, @meling presented this solution way back, Im not sure if theres a better way to resolve it. 

The issue is when not specifying that version and all the code generated has a bunch of compiler errors.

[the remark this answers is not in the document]
meling: See [draft PR #1147](https://github.com/quickfeed/quickfeed/pull/1147)... which is necessary if we don't pin to v1.10.0. I suggest that we pin it here, and fix #1147 soon...

qf/requests.proto
JoachimTislov: `UpdateSubmissionRequest's` current score field would be both.. score and `minimumScore`, as stated in the frontend when creating the request; 

`scoreLimit: state.review.minimumScore`

The updateSubmission updates the score to a given value, where as update submissions updates all the submissions given a minium score.

JoachimTislov: > assignmentID = 10, score = 90, approve = true

approve is always true by;

```go
func (s *QuickFeedService) updateSubmissions(request *qf.UpdateSubmissionsRequest) error {
	query := &qf.Submission{
		AssignmentID: request.GetAssignmentID(),
		Score:        request.GetScoreLimit(),
		Released:     request.GetRelease(),
	}

---> return s.db.UpdateSubmissions(query, true)
}

// Should be
return s.db.UpdateSubmissions(query, request.GetApprove())
// To not ignore it
```


This logic;

```go
status := qf.Submission_APPROVED
if !approve {
	status = qf.Submission_REJECTED
}
```

Appears to be redundant and not needed if approve is always true.


Thereby after stating previous points, will sending the status as `qf.Submission_APPROVED` be the same. Meaning this is still possible: `assignmentID = 10, score = 90, approve = true and it would approve all submissions for assignment 10 where score was >= 90.`

You can even specify if submissions above a certain score should be certain status.

Did I understand this correctly? Im pretty fried after trying to understand access control, including the test for it...

qf/requests.proto
JoachimTislov: This XOR statement works well: `req.GetSubmissionID() > 0 != (req.GetAssignmentID() > 0)`

public/src/overmind/namespaces/review/actions.ts
```diff
@@ -130,14 +130,22 @@ export const setMinimumScore = ({ state }: Context, minimumScore: number): void
     state.review.minimumScore = minimumScore
 }
 
+const determineReleaseString = (release: boolean, approve: boolean): string => {
```
meling, line 133: This PR must be updated, since there is now this more readable code:
```ts
    const releaseString = () => {
        if (release && approve) return "release and approve"
        if (release) return "release"
        if (approve) return "approve"
        return ""
    }
```

JoachimTislov: @meling Done

Issue #1155, opened 2024-11-29
chore: Remove LastActivityDate from CountApprovedSubmissions and rename to UpdateTotalApproved


meling:
Assuming our discussed simplification of the `LastActivityDate` can be done; I think we can rename this method to `UpdateTotalApproved` and we should add a simple doc comment saying something like:

```go
// UpdateTotalApproved updates the total approved assignments for the current enrollment 
// based on the provided submissions.
```

_Originally posted by @meling in https://github.com/quickfeed/quickfeed/pull/1152#discussion_r1862268376_

JoachimTislov:
> Assuming our discussed simplification of the LastActivityDate can be done

Im not sure were this was discussed, but updating the LastActitivyDate seems to be redudant

PR #1232, opened 2025-03-09
chore: Remove LastActivityDate and rename CountApprovedSubmissions to UpdateTotalApproved and 

JoachimTislov:
Closes: #1155 

- Rename CountApprovedSubmissions to UpdateTotalApproved
- Remove unnecessary total approved counter
- Simplify and merge if statements

Im not sure if reseting the `m.TotalApproved` to `0` in the beginning of the method is necessary. I couldn't quite understand if this method is only called with the value as zero or not. Previously it was overwritten by a separate counter variable, which I think is redundant, the increments can be done directly.

I switched out the struct map with a bool map to merge the if statements. Im not sure if this type of mapping is buggy, but the tests pass and there aren't any compling or runtime errors.

### DIFF
--- a/qf/enrollment.go
+++ b/qf/enrollment.go
@@ -3,8 +3,6 @@ package qf
 import (
 	"fmt"
 	"time"
-
-	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
 )
 
 // gracePeriod is the grace period for submissions after the deadline.
@@ -139,29 +137,12 @@ func (m *Enrollments) UserIDs() []uint64 {
 	return userIDs
 }
 
-func (m *Enrollment) CountApprovedSubmissions(submissions []*Submission) {
-	var totalApproved uint64
-	var submissionDate time.Time
-	duplicateAssignments := make(map[uint64]struct{})
-
+func (m *Enrollment) UpdateTotalApproved(submissions []*Submission) {
+	duplicateAssignments := make(map[uint64]bool)
 	for _, s := range submissions {
-		// Ignore duplicate approved assignments
-		if _, ok := duplicateAssignments[s.GetAssignmentID()]; ok {
-			continue
-		}
-		if s.IsApproved(m.GetUserID()) {
-			duplicateAssignments[s.GetAssignmentID()] = struct{}{}
-			totalApproved++
-
-			// Update submissionDate if needed
-			if m.GetLastActivityDate() == nil {
-				submissionDate = s.NewestSubmissionDate(submissionDate)
-			}
+		if s.IsApproved(m.GetUserID()) && !duplicateAssignments[s.GetAssignmentID()] {
+			duplicateAssignments[s.GetAssignmentID()] = true
+			m.TotalApproved++
 		}
 	}
-
-	if m.LastActivityDate == nil && !submissionDate.IsZero() {
-		m.LastActivityDate = timestamppb.New(submissionDate)
-	}
-	m.TotalApproved = totalApproved
 }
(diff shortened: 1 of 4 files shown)


qf/enrollment.go
```diff
@@ -139,29 +137,12 @@ func (m *Enrollments) UserIDs() []uint64 {
...
-		}
-		if s.IsApproved(m.GetUserID()) {
-			duplicateAssignments[s.GetAssignmentID()] = struct{}{}
-			totalApproved++
-
-			// Update submissionDate if needed
-			if m.GetLastActivityDate() == nil {
-				submissionDate = s.NewestSubmissionDate(submissionDate)
-			}
+		if s.IsApproved(m.GetUserID()) && !duplicateAssignments[s.GetAssignmentID()] {
```
meling, line 143: This makes it a bit harder to understand this code. It is more obvious to keep the two if statements separate, along with the comment: `// Ignore duplicate approved assignments`. Please revert the logic to its previous approach. 

Note that using the `struct{}{}` approach requires less allocation. It isn't terribly important here, but it is idiomatic Go (from ChatGPT):

• Using an empty `struct{}` to indicate presence, saves memory since it avoids storing a bool.
• This is idiomatic Go when a map is used as a set.
```go
	var totalApproved uint64
	duplicateAssignments := make(map[uint64]struct{})
	for _, s := range submissions {
		// Ignore duplicate approved assignments
		if _, ok := duplicateAssignments[s.GetAssignmentID()]; ok {
			continue
		}
		if s.IsApproved(m.GetUserID()) {
			duplicateAssignments[s.GetAssignmentID()] = struct{}{}
			totalApproved++
		}
	}
	m.TotalApproved = totalApproved
```

meling (changes requested):
Prefer readability/glancability over fewer lines of code.

qf/enrollment.go
```diff
@@ -139,29 +137,12 @@ func (m *Enrollments) UserIDs() []uint64 {
...
+			duplicateAssignments[s.GetAssignmentID()] = true
+			m.TotalApproved++
```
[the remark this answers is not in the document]
JoachimTislov: As metioned in the PR description:

> Im not sure if reseting the m.TotalApproved to 0 in the beginning of the method is necessary. I couldn't quite understand if this method is only called with the value as zero or not

I think adding a test for this would reasonable, to prevent the case: "surprise ourselves in the future..."

qf/enrollment.go
```diff
@@ -139,29 +137,12 @@ func (m *Enrollments) UserIDs() []uint64 {
 	return userIDs
 }
 
-func (m *Enrollment) CountApprovedSubmissions(submissions []*Submission) {
-	var totalApproved uint64
-	var submissionDate time.Time
-	duplicateAssignments := make(map[uint64]struct{})
-
+func (m *Enrollment) UpdateTotalApproved(submissions []*Submission) {
```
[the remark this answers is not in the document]
JoachimTislov: Oh, I apparently forgot..

qf/enrollment.go
JoachimTislov: Im not sure if I agree with this. I think any decent developer would find `s.IsApproved` along with `!duplicateAssignments` quite readable and understandable, you could even add a comment like this  `// Increment total approved if its approved and not a duplicate` above to explain it even better. This if statement would of course not be possible without it being the `map[uint64]bool` type

But I understand your concern, and can revert the changes.

JoachimTislov:
> Prefer readability/glancability over fewer lines of code.

Understood

qf/enrollment.go
JoachimTislov: I think we should refer to a new developer who isn't familiar with the code, when refering to the amount of time it would take to understand the logic.

meling (approved):

PR #1239, opened 2025-03-13
Resolve DeepSource - RVV-B0001. Refactor and update tests, and write new and more tests cases for quickfeed_service.go

JoachimTislov:
Closes: #1240 
Closes: #1267

Partially related to #1259 

Codecov:

image


meling (changes requested):
There are several untested code paths in `quickfeed_service.go` that seems like they should be tested. We should open a new issue to improve test coverage for this part.

I've not had time to review this properly. Will try to do it tomorrow morning.

web/rebuild.go
```diff
@@ -13,7 +13,7 @@ import (
 const maxContainers = 10
 
 // rebuildSubmission rebuilds the given assignment and submission.
```
meling, line 15: Should update doc comment as well.

web/rebuild.go
```diff
@@ -69,7 +69,7 @@ func (s *QuickFeedService) rebuildSubmission(request *qf.RebuildRequest) error {
 	return nil
 }
 
-func (s *QuickFeedService) rebuildSubmissions(request *qf.RebuildRequest) error {
+func (s *QuickFeedService) internalRebuildSubmissions(request *qf.RebuildRequest) error {
```
meling, line 72: We should rename this to `internalRebuildAll` or `...AllSubmissions`

web/quickfeed_service.go
```diff
@@ -336,11 +363,16 @@ func (s *QuickFeedService) GetSubmission(_ context.Context, in *connect.Request[
 // GetSubmissions returns the submissions matching the query encoded in the action request.
 func (s *QuickFeedService) GetSubmissions(ctx context.Context, in *connect.Request[qf.SubmissionRequest]) (*connect.Response[qf.Submissions], error) {
 	s.logger.Debugf("GetSubmissions: %v", in.Msg)
-	submissions, err := s.getSubmissions(in.Msg)
+	query := &qf.Submission{
+		UserID:  in.Msg.GetUserID(),
+		GroupID: in.Msg.GetGroupID(),
+	}
+	subs, err := s.db.GetLastSubmissions(in.Msg.GetCourseID(), query)
```
meling, line 370: Could we make the `db.GetLastSubmissions` return a `qf.Submissions` directly, so that we could avoid the `submissions := ...` line below (Line 375).

web/quickfeed_service.go
```diff
@@ -367,7 +399,18 @@ func (s *QuickFeedService) GetSubmissionsByCourse(_ context.Context, in *connect
 
 // UpdateSubmission is called to approve the given submission or to undo approval.
 func (s *QuickFeedService) UpdateSubmission(_ context.Context, in *connect.Request[qf.UpdateSubmissionRequest]) (*connect.Response[qf.Void], error) {
-	err := s.updateSubmission(in.Msg.GetSubmissionID(), in.Msg.GetGrades(), in.Msg.GetReleased(), in.Msg.GetScore())
+	submission, err := s.db.GetSubmission(&qf.Submission{ID: in.Msg.GetSubmissionID()})
+	if err != nil {
+		return nil, connect.NewError(connect.CodeNotFound, errors.New("failed to get submission"))
```
meling, line 404: We should log the actual error on the server here, like we do below (Line 415):
```go
s.logger.Errorf("UpdateSubmission failed: %v", err)
```
We don't want to return the actual error to the user, but we want to have a log entry if it does happen... Can help debugging a problem in the future.

web/quickfeed_service.go
```diff
@@ -367,7 +399,18 @@ func (s *QuickFeedService) GetSubmissionsByCourse(_ context.Context, in *connect
...
+	}
+	for _, grade := range in.Msg.GetGrades() {
```
meling, line 406: Would it make sense to move this logic to a new method `SetGradeAndScore(qf.UpdateSubmissionRequest)` on the `qf.Submission` type?

web/quickfeed_service.go
```diff
@@ -397,12 +440,16 @@ func (s *QuickFeedService) RebuildSubmissions(_ context.Context, in *connect.Req
 
 // CreateBenchmark adds a new grading benchmark for an assignment.
 func (s *QuickFeedService) CreateBenchmark(_ context.Context, in *connect.Request[qf.GradingBenchmark]) (*connect.Response[qf.GradingBenchmark], error) {
-	bm, err := s.createBenchmark(in.Msg)
-	if err != nil {
+	if _, err := s.db.GetAssignment(&qf.Assignment{
+		ID: in.Msg.AssignmentID,
+	}); err != nil {
+		return nil, err
```
meling, line 446:
```
s.logger.Errorf("CreateBenchmark failed for %+v: %v", in, err)
return nil, connect.NewError(connect.CodeInvalidArgument, errors.New("failed to create benchmark"))
```

web/quickfeed_service.go
```diff
@@ -397,12 +440,16 @@ func (s *QuickFeedService) RebuildSubmissions(_ context.Context, in *connect.Req
 
 // CreateBenchmark adds a new grading benchmark for an assignment.
 func (s *QuickFeedService) CreateBenchmark(_ context.Context, in *connect.Request[qf.GradingBenchmark]) (*connect.Response[qf.GradingBenchmark], error) {
-	bm, err := s.createBenchmark(in.Msg)
-	if err != nil {
+	if _, err := s.db.GetAssignment(&qf.Assignment{
```
meling, line 443: Can this `GetAssignment` be done first in `CreateBenchmark`?... Or are we using `CreateBenchmark` somewhere else, where we do not need to get the assignment first?

meling (changes requested):
Many of these changes requires that we add logging of errors on the server side and return a `connect.NewError` instead of the actual error.

Also, it now has merge conflicts that must be fixed first.

Also, check if some of the "business logic" that only involve the `s.db` can be migrated to the relevant database method instead of doing multiple database calls from these `internal*` methods or `quickfeed_service.go` methods.

web/quickfeed_service.go
JoachimTislov: Yes, but it will affect alot more lines the tests... 

All the lines would go from submissions... -> submissions.Submissions...

JoachimTislov: The annoyance can be resolved by naming the *qf.Submissions type to "s" and have s.Submissions every in the test

meling: ok; let's keep it for now. we can fix the tests and all later if we want to do it.

web/quickfeed_service.go
JoachimTislov: `SetGradeAndScore` does not reflect the `released` update. This can either be specified in doc comment or the name could be `Update`, `ToggleApproval` or `SetReleasedAndMarks`..

assignments/walk_tests_repo_test.go
```diff
@@ -40,10 +40,7 @@ func TestWalkTestsRepository(t *testing.T) {
 }
 
 func TestReadTestsRepositoryContent(t *testing.T) {
-	wantDockerfile := `FROM golang:1.19-alpine
-RUN apk update && apk add --no-cache git bash build-base
-WORKDIR /quickfeed
-`
+	wantDockerfile := "FROM golang:1.24-alpine\nRUN apk update && apk add --no-cache git=~2.47 bash=~5.2.37 build-base=~0.5\nWORKDIR /quickfeed\n"
```
meling, line 43: This change is now in master; so you should merge master into this PR.

database/gormdb_assignment.go
```diff
@@ -217,6 +217,11 @@ func (db *GormDB) GetCourseSubmissions(courseID uint64, submissionType qf.Submis
 
 // CreateBenchmark creates a new grading benchmark
 func (db *GormDB) CreateBenchmark(query *qf.GradingBenchmark) error {
+	if _, err := db.GetAssignment(&qf.Assignment{
+		ID: query.GetAssignmentID(),
+	}); err != nil {
+		return err
+	}
```
meling, line 224: Need test for error

qf/submission.go
```diff
@@ -42,6 +42,17 @@ func (s *Submission) GetStatusByUser(userID uint64) Submission_Status {
 	return Submission_NONE
 }
 
+// SetReleasedAndMarks sets the submission's released status, grades and score
+func (s *Submission) SetReleasedAndMarks(request *UpdateSubmissionRequest) {
```
meling, line 46: We don't have a concept `Marks` so this is confusing.
```go
// SetGradesAndRelease sets the submission's grade, score and released status.
```

Also, we need a test for this.

meling (changes requested):

web/quickfeed_service.go
JoachimTislov: Ping @meling

meling (changes requested):
This still does not have tests for the issues raised by codecov. These are important parts of the code that we should have good tests for. We should fix tests as part of this PR.

web/quickfeed_service.go
meling: Do you have a question about this? Or are you looking for guidance on what to write in the error messages? If so, write what you think is appropriate and I will suggest improvements in code review.

meling: My main point is that we should always log the root cause error, even if we also send it to the frontend. We always want to log errors. They happen rarely, so they don’t cause any major concerns in the logs. But when they happen they may be essential for debugging a problem.

JoachimTislov: I think you misunderstood my confusion by;

> "We don't want to return the actual error to the user"

Can you clarify if this contradicts with the current errors returned in the api end-points or if there's something im missing here?

web/groups.go
```diff
@@ -189,7 +169,7 @@ func (s *QuickFeedService) getGroupUsers(request *qf.Group) ([]*qf.User, error)
 
 	users, err := s.db.GetUsers(userIds...)
 	if err != nil {
-		return nil, err
+		return nil, connect.NewError(connect.CodeInternal, errors.New("failed to get users"))
```
meling, line 172: Applies for all cases: We should return just `err` from helper methods like this one so that the original error can be logged at the call site in `quickfeed_service.go` only, where we should do the `connect.NewError` creation.

web/users.go
```diff
@@ -2,13 +2,13 @@ package web
 
 import "github.com/quickfeed/quickfeed/qf"
 
-// updateUser updates the user profile according to the user data in
+// editUserProfile updates the user profile according to the user data in
 // the request object. If curUser is admin, and the request may also
 // promote the user to admin.
-func (s *QuickFeedService) updateUser(curUser *qf.User, request *qf.User) (*qf.User, error) {
+func (s *QuickFeedService) editUserProfile(curUser *qf.User, request *qf.User) error {
```
meling, line 8:
```diff
- func (s *QuickFeedService) editUserProfile(curUser *qf.User, request *qf.User) error
+ func (s *QuickFeedService) editUserProfile(curUser, request *qf.User) error
```

web/quickfeed_service.go
```diff
@@ -85,9 +85,23 @@ func (s *QuickFeedService) UpdateCourse(ctx context.Context, in *connect.Request
 	scmClient, err := s.getSCM(ctx, in.Msg.GetScmOrganizationName())
 	if err != nil {
 		s.logger.Errorf("UpdateCourse failed: could not create scm client for organization %s: %v", in.Msg.GetScmOrganizationName(), err)
-		return nil, connect.NewError(connect.CodeNotFound, err)
+		return nil, connect.NewError(connect.CodeNotFound, fmt.Errorf("unable to connect to the gitHub organization: %s", in.Msg.GetScmOrganizationName()))
```
meling, line 88: In the interest of consistency across all service methods, replace these return types with just this:
```go
return nil, connect.NewError(connect.CodeNotFound, errors.New("unable to connect to the GitHub organization for the course"))
```
The logs will have more information if this should ever happen, but right now, we don't have the course organization name available in `UpdateEnrollments`, `UpdateGroup`, and `DeleteGroup`. We could make this common error message a global variable that we can reuse instead of having slightly different error messages for each of these.
```go
var scmConnectErr = connect.NewError(connect.CodeNotFound, errors.New("unable to connect to the GitHub organization for the course"))

return nil, scmConnectErr
```

meling (changes requested):
Phew! This was a major effort to review. We need to keep things smaller.

database/gormdb_submission.go
```diff
@@ -274,23 +282,81 @@ func (db *GormDB) GetReview(query *qf.Review) (*qf.Review, error) {
 
 // CreateReview creates a new submission review.
 func (db *GormDB) CreateReview(query *qf.Review) error {
+	submission, err := db.GetSubmission(&qf.Submission{ID: query.GetSubmissionID()})
+	if err != nil {
+		return fmt.Errorf("failed to get submission: %v", err)
+	}
+	assignment, err := db.GetAssignment(&qf.Assignment{ID: submission.GetAssignmentID()})
+	if err != nil {
+		return fmt.Errorf("failed to get assignment: %v", err)
```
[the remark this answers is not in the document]
JoachimTislov: Its quite tricky to test error paths. There are restrictions for what you can do with the database CRUD functions. For example in this case will you not be able to test the `GetAssignment` error path as its not possible to create a submission `without an associated assignment`. 

In other cases is the same ID used for database queries and you can therefore only test one error path. I don't know if theres a good solution to this. Maybe you can somehow mock the different parts or something similar..

JoachimTislov:
> Phew! This was a major effort to review. We need to keep things smaller.

I agree! — it can be pretty tricky to keep track of all the changes I've made, especially when trying to hold multiple pieces of logic in mind at once. It starts to feel inefficient. I prefer to keep the changes in the same pull request if the changes will touch the same files or if it's related to the same feature and the functions depend on each other.

Are there any best practices for managing the size of PRs?

Issue #40, opened 2017-07-28
Make test data for database that gets inserted on agserver startup


meling:
This is for testing only, so that we all see the same data when testing the system.

I prefer it to be done in Go, e.g. reading from sql file.

nicro950:
can anyone confirm that the test data is good enough

meling:
Where is it?

nicro950:
/mics/ag.db.sql

meling:
We now have a database of several courses and large sets of students. We should anonymize the students, replace links, and upload a test database for others to use for development. We also need to make sure that the `access_token` field is removed.

meling:
Can we add a test db to the repo? Not sure, but I think we might have done this, and it is just a matter of setting it up properly… and documentation

JosteinLindhom:
We have a test db, with some documentation [here](https://github.com/quickfeed/quickfeed/tree/master/doc/dev)

meling:
Two questions:

1. Is the documentation sufficient and still accurate? It looks a bit outdated and perhaps underspecified. The doc points to qf406; not sure if there are any private repos in it and it uses course-info instead of info. Anyway, I'm wondering if we should switch to use a database with courses qf101 and qf102, and define some policy for making changes to those. E.g., qf101 is currently used by the pull request bachelor assignment, but we could perhaps set up qf102 to be stable with some testable things. That is, that we can in fact run specific tests using scm GitHub APIs.
2. Is the database sufficiently populated with necessary data?

meling:
Replaced by #1638, which covers this along with the SCM side that #1341 asked for.

The specific ask here, a populated database that everyone shares, is delivered by #1639 (the fixture set) and #1642 (seeding a local instance from it). #1643 then retires both `doc/dev/test.db` and the `make dev-db` workflow that downloads a real database and runs the anonymizer over it.

Worth recording before this closes: `doc/dev/test.db` is not merely old, it no longer matches the schema. It still has a `remote_identities` table that is not in `AutoMigrate`, and it is missing `grades`, `notes`, `assignment_feedbacks`, `feedback_receipts`, and `test_infos`. The instructions in `doc/dev/testdb.md` to `UPDATE remote_identities` cannot work.

meling:
Closing in favor of #1638.

PR #1242, opened 2025-03-14
Anonymizer script now supports fetching user info via Github API

JosteinLindhom:


JoachimTislov (commented):
Looks good, I think integrating the `python3 cmd/anonymize/main.py --database /path/to/db --admin JosteinLindhom` into a reasonably named makefile target would ease the expierence of generating a usable database for development, and would close #40 

Refer to [Passing arguments to "make run"](https://stackoverflow.com/questions/2214575/passing-arguments-to-make-run). Do something like this in the makefile:

```makefile
dev-db:
     python3 cmd/anonymize/main.py --database /path/to/db --admin $(admin)
// and have an if statement checking if the admin argument is not an empty string.
```

Which you can call like this: `make dev-db admin=JosteinLindhom`

JoachimTislov:
The path to the db could be updated to `doc/dev/test.db` or similar and the new database you generated can be placed there.


Im not sure if we should do this in a separate PR...

meling:
Let’s fix #40 as part of this PR since there is no issue associated with it.

meling (approved):
@JosteinLindhom I've added a Makefile target as suggested by Joachim. Please check my changes. I haven't tested this myself, since I didn't want to download the database ;-)

PR #1243, opened 2025-03-14
Syncronize group name validation with Github's repository name validation

JoachimTislov:
Closes: #1107 

Refer to [what-is-the-format-for-valid-git-repo-names](https://stackoverflow.com/questions/21262997/what-is-the-format-for-valid-git-repo-names)

Im not sure about periods in group names but its allowed in github repo names.

Added restriction of a group to be empty inside of validation, this was only checked in the frontend.


web/groups.go
```diff
@@ -197,14 +198,21 @@ func (s *QuickFeedService) getGroupUsers(request *qf.Group) ([]*qf.User, error)
...
 }
 
-// only allow letters, numbers, dash and underscore.
-var regexpNonAuthorizedChars = regexp.MustCompile("[^a-zA-Z0-9-_]")
+// only allow letters, numbers, underscore, dash and period.
+var regexpNonAuthorizedChars = regexp.MustCompile("[^a-zA-Z0-9_.-]")
 
 // checkGroupName returns an error if the group name is invalid; otherwise nil is returned.
 func (s *QuickFeedService) checkGroupName(courseID uint64, groupName string) error {
+	if groupName == "" {
```
meling, line 206: Missing test for this

web/groups.go
```diff
@@ -197,14 +198,21 @@ func (s *QuickFeedService) getGroupUsers(request *qf.Group) ([]*qf.User, error)
 	return users, nil
 }
 
-// only allow letters, numbers, dash and underscore.
-var regexpNonAuthorizedChars = regexp.MustCompile("[^a-zA-Z0-9-_]")
+// only allow letters, numbers, underscore, dash and period.
+var regexpNonAuthorizedChars = regexp.MustCompile("[^a-zA-Z0-9_.-]")
```
meling, line 202: Even if GitHub allows period in the repo name, we don’t need to allow it. I prefer to disallow it.

web/groups.go
```diff
@@ -197,14 +198,21 @@ func (s *QuickFeedService) getGroupUsers(request *qf.Group) ([]*qf.User, error)
...
 
 // checkGroupName returns an error if the group name is invalid; otherwise nil is returned.
 func (s *QuickFeedService) checkGroupName(courseID uint64, groupName string) error {
+	if groupName == "" {
+		return ErrGroupNameEmpty
+	}
 	if len(groupName) > maxGroupNameLength {
 		return ErrGroupNameTooLong
 	}
+	// check if group name contains space(s)
```
meling, line 212: Missing test for this. Also,  don’t think it is needed as a separate check since the reg exp does not allow it.

public/src/Helpers.ts
```diff
@@ -340,7 +340,7 @@ export const getSubmissionCellColor = (submission: Submission, owner:  Enrollmen
 }
 
 // pattern for group name validation. Only letters, numbers, underscores and dashes are allowed.
-const pattern = /^[a-zA-Z0-9_-]+$/
+const pattern = /^[a-zA-Z0-9_.-]+$/
```
meling, line 343: See my comment below about period

meling (changes requested):

meling:
This looks like it can be merged once the conflicts have been fixed.

meling:
@JoachimTislov ^^^

meling (approved):

PR #1262, opened 2025-03-21
Replace SubmissionStatus with SetGradesIfApproved

### LINKED ISSUE #1261: chore: replace SubmissionStatus with new method on Submission instead
This code is confusing. It does two things and the main thing it does is not expressed in the method name. It is also confusing that this is a method on `Assignment` since it is updating the provided `Submission`.
```go
// SubmissionStatus returns the existing grade submission status, or an approved submission status
// if the score of the latest submission is sufficient to autoapprove the assignment.
func (a *Assignment) SubmissionStatus(latest *Submission, score uint32) []*Grade {
	if a.GetAutoApprove() && score >= a.GetScoreLimit() {
		latest.SetGradeAll(Submission_APPROVED)
	}
	// keep existing status if already approved/revision/rejected
	return latest.GetGrades()
}
```
This should be replaced by a single method:
```go
// SetGradesIfApproved marks the submission approved for all group members
// or a single user if the assignment is autoapprove and
// the score is greater or equal to the assignment's score limit.
func (s *Submission) SetGradesIfApproved(a *Assignment, score uint32) {
	if a.GetAutoApprove() && score >= a.GetScoreLimit() {
		s.SetGradeAll(Submission_APPROVED)
	}
}
```
We don't need to return the submission's `[]Grade` since the user already has the `s` reference and can call `s.GetGrades()` after calling `SetGradesIfApproved`.

meling:
Fixes #1261

- **chore: Implement auto-approval grading logic on submission**
- **chore: Move TestSetGradesIfApproved to submission_test.go**


JosteinLindhom:
Looks good!

JoachimTislov:
^^ Cleaner 👍

Issue #1281, opened 2025-03-27
NavBarActiveCourse is shown in navbar when user is at Dashboard. Its redudant


JoachimTislov:
![Image](https://github.com/user-attachments/assets/0518c519-0d7c-4a25-99fb-28087bc56f2a)

Simply adding this to the if statement in NavBarActiveCourse.tsx, resolves the issue;

```typescript
const location = useLocation()
if (!isLoggedIn || !activeCourse || !enrollment --> || location.pathname === "/") {
    return null
}
```

PR #1293, opened 2025-03-27
Remove unnecessary visibility of NavbarActiveCourse when user is at the Dashboard

JoachimTislov:
Closes: #1281


meling (approved):
I don’t have a strong opinion whether or not it is necessary to hide the dashboard text. @JosteinLindhom

JosteinLindhom:
> I don’t have a strong opinion whether or not it is necessary to hide the dashboard text. @JosteinLindhom

The main issue I think if that the teacher / student toggle is visible as well, and toggling it will affect the previously selected course.
This might not be very intuitive, and the change in this PR only displays it if actively viewing a course.

Issue #1282, opened 2025-03-27
Sidebar moves over main page. Doesn't take the main page into account when reloading.


JoachimTislov:
Teacher page after reload:

![Image](https://github.com/user-attachments/assets/ad5df883-0c7f-4cd6-a47c-64c6b058dabf)

Switching out `useEffect` with `useLayoutEffect` resolves this issue.

Like this:

```typescript
  useLayoutEffect(() => {
        if (!state.showFavorites) {
            actions.toggleFavorites()
        }
        actions.setActiveCourse(courseID)
        actions.getCourseData({ courseID })
    }, [courseID])
```

JosteinLindhom:
I experienced this issue while reviewing #1301. 
However, when checking our production server (lagging behind master) this issue seems to not be reproducible there.
Not sure which previous change broke this, but it is broken in master as you point out here.

PR #1294, opened 2025-03-27
Prevent the sidebar from moving in front of the page.

JoachimTislov:
useLayoutEffect takes the sidebar and main page into consideration when loading the page and prevents it.

Closes: #1282

### DIFF
--- a/public/src/pages/CoursePage.tsx
+++ b/public/src/pages/CoursePage.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from "react"
+import React, { useLayoutEffect } from "react"
 import { Redirect } from "react-router"
 import { getCourseID, isEnrolled, isTeacher } from "../Helpers"
 import { useActions, useAppState } from "../overmind"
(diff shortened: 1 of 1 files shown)


public/src/pages/CoursePage.tsx
```diff
@@ -14,13 +14,15 @@ const CoursePage = () => {
...
 
-    useEffect(() => {
+    useLayoutEffect(() => {
         if (!state.showFavorites) {
             actions.toggleFavorites()
         }
         actions.setActiveCourse(courseID)
         actions.getCourseData({ courseID })
-    }, [courseID])
+        // Having state.showFavorites in the dependency array locks the sidebar as open.
```
meling, line 23: This first comment line should be moved closer to where used (above the if statement). That said, @JosteinLindhom please check if it makes sense to you. I don’t know what the dependency array refers to.

JoachimTislov: > This first comment line should be moved closer to where used (above the if statement).

The comment is related to the array beneath;

```typscript
   // Having state.showFavorites in the dependency array locks the sidebar as open.
   // eslint-disable-next-line react-hooks/exhaustive-deps
  }, --> [actions, courseID])
```

It can't be inside there, because then it always reverses the change from `state.showFavorites` being updated to false to be true again. That is done above and is related but the comment is meant for the array.

A better placement would probably be beside the array;

```typescript
}, [actions, courseID]) // Having state.showFavorites in the dependency array locks the sidebar as open.
```

Issue #1285, opened 2025-03-27
NavbarLabs - Duplicate keys in react elements 


JoachimTislov:
![Image](https://github.com/user-attachments/assets/edb69685-5578-4c79-ac82-4360fc2f262d)

The fix is to use the submission ID instead of assignment ID

PR #1296, opened 2025-03-27
Use submission ID instead of assignment ID to prevent duplicate keys for elements (react issue)

JoachimTislov:
Closes: #1285


meling (approved):
Looks good to me! 
@JosteinLindhom i prefer to have your input on these frontend issues as well; even though this seems easy enough.

Issue #1280, opened 2025-03-27
Two Ready Buttons when Review Assignment is "In Progress"


JoachimTislov:
![Image](https://github.com/user-attachments/assets/ac959f40-837a-4aa7-8d3d-a7faa0fa730c)

The header bar could also be orange and the `In progress` can be moved into the header;

![Image](https://github.com/user-attachments/assets/6aa30351-1ebe-4900-a95b-a1f6c2df5da0)

JoachimTislov:
Closing because its intended.

Issue #1279, opened 2025-03-27
Ready should be placed beside released/release button


JoachimTislov:
Either of the places is better;

![Image](https://github.com/user-attachments/assets/4f80202c-b641-4e9b-8355-2d0ad018f472)

The `lab3` text can also be larger

And im not sure if `Ready` can be `Not Ready`, but if thats the case, then its even an better reason to be beside `released`

JosteinLindhom:
If no review exist for a submission, there will be a button that says `Add Review` in place of `Ready`.

If a submission is not ready, the button will say `In Progress`.
There can be multiple reviews for a submission. In that case there will be multiple buttons. Clicking one of the buttons will then change which review to look at.

I don't think we currently use more than 1 review per submission in any course.

meling:
Could you guys draw a flow chart to more clearly explain what the different things means and how they relate to each other.

PR #1274, opened 2025-03-26
Frontend: Performance improvements and resolve DeepSource issues

JoachimTislov:
Be aware. The last commit is only temporary. It was time consuming and difficult doing this the right way and therefore ended up pushing my changes in order to not loose them. I don't trust my capabilities on git locally.

Closes: #1269
Related to [Frontend performance improvements](https://github.com/quickfeed/quickfeed/issues/1045)

PR #1300, opened 2025-03-27
Restructure the review assignments - lab view

JoachimTislov:
Closes: #1280 
Resolves: #1279

I personally don't like how the assignment name and mark ready button isn't aligned.. 

Im thinking maybe increasing the text and capatalizing the first letter would be sufficient.

Before:

- Ready and released

image

- In progress and not released:

image

After: 

- In progress and not released

image

- Ready and released:

image

### DIFF
--- a/public/src/components/ManageSubmissionStatus.tsx
+++ b/public/src/components/ManageSubmissionStatus.tsx
@@ -76,7 +76,7 @@ const ManageSubmissionStatus = ({ courseID, reviewers }: { courseID: string, rev
                         text={viewIndividualGrades ? "All Grades" : "Individual Grades"}
                         color={Color.GRAY}
                         type={ButtonType.OUTLINE}
-                        className="col mr-2"
+                        className="col mx-1"
                         onClick={async () => setViewIndividualGrades(!viewIndividualGrades)}
                     />
                 )}
@@ -85,7 +85,7 @@ const ManageSubmissionStatus = ({ courseID, reviewers }: { courseID: string, rev
                         text={rebuilding ? "Rebuilding..." : "Rebuild"}
                         color={Color.BLUE}
                         type={ButtonType.OUTLINE}
-                        className="col mr-2"
+                        className="col mx-1"
                         onClick={handleRebuild}
                     />
                 )}
@@ -97,21 +97,21 @@ const ManageSubmissionStatus = ({ courseID, reviewers }: { courseID: string, rev
                         text="Approve"
                         color={Color.GREEN}
                         type={getButtonType(Submission_Status.APPROVED)}
-                        className="col mr-2"
+                        className="col mx-1"
                         onClick={() => handleSetStatus(Submission_Status.APPROVED)}
                     />
                     <DynamicButton
                         text="Revision"
                         color={Color.YELLOW}
                         type={getButtonType(Submission_Status.REVISION)}
-                        className="col mr-2"
+                        className="col mx-1"
                         onClick={() => handleSetStatus(Submission_Status.REVISION)}
                     />
                     <DynamicButton
                         text="Reject"
                         color={Color.RED}
                         type={getButtonType(Submission_Status.REJECTED)}
-                        className="col mr-2"
+                        className="col mx-1"
                         onClick={() => handleSetStatus(Submission_Status.REJECTED)}
                     />
                 </div>
@@ -127,7 +127,7 @@ const ManageSubmissionStatus = ({ courseID, reviewers }: { courseID: string, rev
                                         text="Approve"
                                         color={Color.GREEN}
                                         type={getGradeButtonType(grade, Submission_Status.APPROVED)}
-                                        className="mr-2"
+                                        className="mx-1"
                                         onClick={() => handleSetGrade(grade, Submission_Status.APPROVED)}
                                     />
                                 </td>
@@ -136,7 +136,7 @@ const ManageSubmissionStatus = ({ courseID, reviewers }: { courseID: string, rev
                                         text="Revision"
                                         color={Color.YELLOW}
                                         type={getGradeButtonType(grade, Submission_Status.REVISION)}
-                                        className="mr-2"
+                                        className="mx-1"
                                         onClick={() => handleSetGrade(grade, Submission_Status.REVISION)}
                                     />
                                 </td>
@@ -145,7 +145,7 @@ const ManageSubmissionStatus = ({ courseID, reviewers }: { courseID: string, rev
                                         text="Reject"
                                         color={Color.RED}
                                         type={getGradeButtonType(grade, Submission_Status.REJECTED)}
-                                        className="mr-2"
+                                        className="mx-1"
                                         onClick={() => handleSetGrade(grade, Submission_Status.REJECTED)}
                                     />
                                 </td>
--- a/public/src/components/ReviewResult.tsx
+++ b/public/src/components/ReviewResult.tsx
@@ -3,7 +3,6 @@ import { Review } from "../../proto/qf/types_pb"
 import { hasBenchmarks } from "../Helpers"
 import Benchmark from "./manual-grading/Benchmark"
 import Criteria from "./manual-grading/Criterion"
-import MarkReadyButton from "./manual-grading/MarkReadyButton"
 import SummaryFeedback from "./manual-grading/SummaryFeedback"
 
 
@@ -40,13 +39,6 @@ const ReviewResult = ({ review }: { review?: Review }) => {
             </tbody>
             <tfoot>
                 <SummaryFeedback review={review} />
-                {!review.ready
-                    ?
-                    <tr>
-                        <MarkReadyButton review={review} />
-                    </tr>
-                    : null
-                }
             </tfoot>
         </table>
     )
--- a/public/src/components/manual-grading/MarkReadyButton.tsx
+++ b/public/src/components/manual-grading/MarkReadyButton.tsx
@@ -1,14 +1,11 @@
 import React from "react"
-import { Review } from "../../../proto/qf/types_pb"
 import { Color } from "../../Helpers"
-import { useActions, useAppState } from "../../overmind"
+import { useActions } from "../../overmind"
 import Button, { ButtonType } from "../admin/Button"
 
 
-const MarkReadyButton = ({ review }: { review: Review }) => {
-    const allCriteriaGraded = useAppState((state) => state.review.graded === state.review.criteriaTotal)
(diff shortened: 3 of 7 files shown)


meling:
I’m a bit confused by all the complexity here. But would it not be better to keep all clickable buttons in the labX title bar, but placed to the right side. The status fields below shouldn’t contain any buttons. If not already supported, we should probably have a keyboard shortcut to mark ready.

JoachimTislov:
> I’m a bit confused by all the complexity here. But would it not be better to keep all clickable buttons in the labX title bar, but placed to the right side. The status fields below shouldn’t contain any buttons. If not already supported, we should probably have a keyboard shortcut to mark ready.

It would be reasonable to group everything together, but considering the amount of buttons, it might be to many in one place. The "Ready" or "In progress" button beside "lab3" is disabled.

> we should probably have a keyboard shortcut to mark ready.

I can look into it.

JosteinLindhom:
I would prefer that the button is placed next to the release lab button I think.

Currently it looks like this for me:
bilde

I think this might look better:

bilde

While you're at it, feel free to make stuff centered vertically in the header. Currently it's off by a bit.

JosteinLindhom:
> The "Ready" or "In progress" button beside "lab3" is disabled.

It's not disabled, although clicking it will visually do anything.
When a user selects a submission we try to select the first review associated with it.

The `Ready` / `In Progress` button, when clicked, selects the review it's associated with.

We do not currently have any courses with more than one review per submission however.
If we did, it could look like this (with the changes in this PR): 

bilde

Where clicking on either of the buttons would switch between reviews.

meling:
I like @JosteinLindhom suggested title bar with two buttons. But I would go further and move the Mark in progress button up also. That said, I really don’t understand why we need so many buttons. Especially since we have three buttons at the bottom as well… couldn’t we collapse these things into a simpler more understandable interface?

meling:
We could model this as a state machine with transitions due to button clicks.

JoachimTislov:
> It's not disabled, although clicking it will not visually do anything.
When a user selects a submission we try to select the first review associated with it.

Right, but what styling is "disabled";

image


@JosteinLindhom, could you test recent commit resolves your concern in #1279

JosteinLindhom:
> @JosteinLindhom, could you test if the recent commit resolves your concern/point in #1279

It seems to work now.

The `Release Lab` and `Revert Release` seem to be the wrong way around.
A released lab shows `Release Lab` while unreleased labs show `Revert Release`.

JoachimTislov:
> The Release Lab and Revert Release seem to be the wrong way around.

Yes, good catch. I can actually remember myself doing this silly mistake..

JoachimTislov:
> I really don’t understand why we need so many buttons. Especially since we have three buttons at the bottom as well… couldn’t we collapse these things into a simpler more understandable interface?

Collapsing the interface would worsen the user experience. Its easier to view all of the buttons and be able to choose between them, instead of needing to first click the drop down and then choose an outcome. The less clicking the better.

> But I would go further and move the Mark in progress button up also.

This is now up there;

image

JoachimTislov:
DeepSource doesn't like this: `function () { actions.review.createReview() }`as well. Good to know. I am currently fixing all of the JS-0417 occurrences in this PR #1274, so I won't do it here.

meling:
> > I really don’t understand why we need so many buttons. Especially since we have three buttons at the bottom as well… couldn’t we collapse these things into a simpler more understandable interface?
> 
> Collapsing the interface would worsen the user experience. It's easier to view all of the buttons and be able to choose between them, instead of needing to first click the drop down and then choose an outcome. The less clicking the better.

I'm **not arguing for two-stage clicking** via a dropdown or something. I just don't understand the process here. What is the difference between Mark in progress, Ready, Release... Can't these buttons be combined and show different things at different stages of the process...

I mean should it be possible to click Release or Ready when Mark in progress has not yet been clicked? Similarly, when In progress, should it be possible to click Release?

I think a flow chart or state machine diagram could better show the possible steps here so that we can make the process understandable to the user. 

Moreover, does it make sense to Approve, Revision, Reject separately from these other buttons. 

> > But I would go further and move the Mark in progress button up also.
> 
> This is now up there;

There is a slight space difference between the buttons in the last image...

JosteinLindhom:
The way it works is like this:

- Submission exists, but no review yet
- Teacher clicking the submission will only see an `Add Review` button
    - Clicking this button will create a new review for the selected submission
- The created review will have `ready = false` and submission has `submission.released = false`.

At this point these buttons are visible:
- `Release`
- `Mark Ready`
    - This one is disabled until all criteria are graded

After grading all criteria, the reviewer may mark the review as ready (`review.ready = true`).
If a review is marked as ready, the `Mark in progress` button appears. This simply sets `review.ready = false`.

The `Approve`, `Revision` and `Reject` buttons are not visible until a review is marked as ready.

If a submission has a review, and it's not marked ready it will show as this in the table (shadowed borders):
bilde

The `Release` buttons sets the `submission.released = true`. If not released students will not see the results of the review.

I agree that it does not make much sense to show the `Release` button if the review is not ready.

JosteinLindhom:
We added the `Mark Ready` button at the end of the review (bottom) to make reviewing smoother for TAs.
Some assignments have had a long list of criteria, where you would have to scroll down.
Having the `Mark Ready` button at the natural conclusion of the review process (the bottom) proved helpful.

This was during a time where TAs would review multiple submissions in rapid succession, however we seem to have moved to in-person reviews in recent times where the removal of the button at the bottom might not cause that much friction.

JoachimTislov:
The latest update should be a more user friendly solution, but may still be alot of buttons... 

Release is not shown unless the review is ready.

Demonstration;

PR-1300-reviewForm

JoachimTislov:
I forgot to mention that the previous demonstration has the ManageSubmissionsStatus and Add review button visible at all times. ManageSubmissionsStatus was by mistake and add review was there to visualize multiple reviews for one lab.

Heres the visuals for the latest commit;

PR-1300-reviewForm-2

meling:
Why is the mark in progress button differently sized? Why is it placed on the left, when the release button is on the right? Why is the mark in progress shown after release is clicked? Why can a single user click many different #x ready buttons? 

I think it is still very confusing and has too many button clicks. Is there a flow chart or state diagram explaining this (as I’ve asked for before)?

Glattshaved:
I have some improvement ideas, i think we need to focus more on simplicity and intuitive design here

- The header with the lab-name on the right side menu should be the only thing visible in the flexbox for a simpler look.
- Simplified name convention on the buttons, use more typical language like "Edit Review", "Create Review", "Submit Review"
- Make the whole right side interface attached while scrolling if you want this functionality
- "#1 Ready/In Progress" should rather be tabs to choose between the reviews the user are working on, not buttons. Too many buttons now.
- Better compartmentalization and informative text for each section. ( Lab title, lab info, review info etc)
- Should add tooltips for buttons with simple explanations on what functionality they have.
- Better coloring for intuitive design as well, looks a bit messy here also because compartmentalization is bad
image

- Also this section where you do the actual review should be more connected to the Review buttons

image

meling:
@Glattshaved Could you try to provide one or more mockups with just the buttons and info fields (no logic yet), to showcase your suggestion?

Glattshaved:
Here is a mockup of what i'm thinking

Additional thoughts:

-Edit button for the Review section can have functionality to set "In Progress" if needed. And "Ready" once it's Saved.

Mockup Review Assignments

meling:
@Glattshaved This looks promising... I need to walk through this more carefully on Friday or maybe tomorrow evening, but one nitpick I would fix is to remove the `:` in the headings. And maybe we can come up with a better title for the `General Info`... Are the green `Approved` and `Ready` "buttons" actually buttons? Perhaps if they are not buttons, then maybe we shouldn't make them look like buttons. Maybe some other way to label them green... Need to think about it some more.

JoachimTislov:
> Is there a flow chart or state diagram explaining this (as I’ve asked for before)?

My skills are quite fresh, but this can somewhat resemble the current logic and process;

image

JoachimTislov:
The graphing software I used was pretty old, heres figma;

Pdf: [Results.Review.pdf](https://github.com/user-attachments/files/19582108/Results.Review.pdf)

 Picture:

image

JoachimTislov:
> Why is the mark in progress button differently sized? Why is it placed on the left, when the release button is on the right? Why is the mark in progress shown after release is clicked? Why can a single user click many different #x ready buttons?

@meling 

Im not sure if you are asking me, but if so; I can't answer for someone else's logic and design choices..

As the title of this PR is "Restructure the review assignments - lab view", not redefine logic and redesign the whole view. I have simply grouped every button together in the banner.

I think we should consult these concern in a different issue..

JosteinLindhom:
> Why is the mark in progress button differently sized?

The `MarkReadyButton` has (for reasons I cannot recall) different visual styles based on the `ready` status:

```ts
const MarkReadyButton = ({ review }: { review: Review }) => {
    const ready = review.ready
    ...

    return (
        <Button
            text={ready ? "Mark In progress" : "Mark Ready"}
            color={ready ? Color.YELLOW : Color.GREEN}
            type={ready ? ButtonType.BADGE : ButtonType.BUTTON}
            ...
        />
    )
}
```

Removing the condition in `type={ready ? ...}` should make the button consistent regardless of ready status.

> Why can a single user click many different #x ready buttons?

The `#x [Ready | In Progress]` shows the status of review `#x`. Clicking this button will switch to review `#x`.
If a submission has a single review, a single button will show. Multiple reviews for a submission will show as:

```
#1 Ready | #2 In Progress
```

Clicking these buttons will switch between the reviews.
Perhaps these buttons should not say `Ready` or `In Progress`, rather they should show the name of the reviewer of respective review that will be selected on click.
The colors could of the button could still be related to its ready status however.

JosteinLindhom:
I see that my previous comment about using the name in the `#x [Ready | In Progress]` was actually implemented a few moments ago!

JoachimTislov:
I think adding a bit more information for the reviewer to see is good, as this allows the reviever to confirm what lab they've selected and don't accidentally add a review to the wrong lab;

image

JoachimTislov:
> I see that my previous comment about using the name in the `#x [Ready | In Progress]` was actually implemented a few moments ago!

Great minds think alike

JoachimTislov:
I am done with this PR. @meling @JosteinLindhom 

I think this is better than whats on main, and I've done what I wished to do.

@Glattshaved Feel free to work more on this matter if its not good enough... 

PR-1300-reviewForm-4

JoachimTislov:
Closing this as not an applicable contribution

Issue #1289, opened 2025-03-27
Suggestion: Consider placing the "back button" relative to the main div on the page


JoachimTislov:
The "back button" is at a fixed spot all of the time and isn't placed relative to the main div on the page. This isn't a problem most of the cases, but for me with a larger screen, gets the following case:

![Image](https://github.com/user-attachments/assets/93d423c0-4574-40f2-9f6a-c9a361c8166d)

If the div is centered, then is should be beside it.

meling:
I think the table should be left-aligned, next to the back-arrow.

PR #1301, opened 2025-03-28
Adjust members page to be displayed over the whole page

JoachimTislov:
Resolves: #1289 

This is a different solution from the planned one in the related issue, but I kinda like it and it was a very minimal change. Only downside is maybe the width on large screens, but you atleast won't make people with a small screen frustrated.

Before:

image

After:

image

Can optionally reduce it to 75% width and aligned to the left;

image

Only need to add: "div className="w-75" inside of <> and in the end add a close tag for the div: `</div>`

### DIFF
--- a/public/src/components/Members.tsx
+++ b/public/src/components/Members.tsx
@@ -93,7 +93,7 @@ const Members = () => {
     })
 
     return (
-        <div className='container'>
+        <>
             <div className="row no-gutters pb-2">
                 <div className="col-md-6">
                     <Search />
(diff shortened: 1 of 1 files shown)


JosteinLindhom:
Looks good. 

The sidebar hides part of the table when doing a full page load on `/members`, but that is likely related to #1294.

meling:
Ok. I would suggest to add a margin on the right comparable to the width introduce by the arrow margin on the left.

JoachimTislov:
> Ok. I would suggest to add a margin on the right comparable to the width introduce by the arrow margin on the left.

The margin is there, I just didn't include it in the picture... my bad. Heres a better one;

image

meling (approved):
@Glattshaved Any comments?

Issue #1286, opened 2025-03-27
Suggestion: Consider adding an arrow for "back to dashboard"


JoachimTislov:
I think this is only relevant to the student's lab and teacher page;

![Image](https://github.com/user-attachments/assets/21ca2ebd-f456-4ad2-b153-7b87f337c2c2)

![Image](https://github.com/user-attachments/assets/ce42ecb1-c955-4d34-a6d7-5142cc5be906)

JoachimTislov:
See #1305 for details about me closing this issue.

PR #1305, opened 2025-03-29
Redirect button - Set root as Dashboard for Teacher and Student Page

JoachimTislov:
Closes: #1286 

Before pictures are on the related issue.

This PR:

- Student:

image

- Teacher:

image


JoachimTislov:
This change breaks the routing. It bypasses directly to the dashboard from for example Results or a given lab..

I will fix it soon.

JoachimTislov:
This is now solved..

JoachimTislov:
In retrospect, this doesn't actually help that much and only adds a third link to the dashboard, which currently only favorite courses. They are already visible on the sidebar...

Issue #1278, opened 2025-03-27
Clicking released in assignments review doesn't persist after reload


JoachimTislov:
After click:

![Image](https://github.com/user-attachments/assets/b80a2ea9-c654-48b2-9099-97f494a89d61)

After reload:

![Image](https://github.com/user-attachments/assets/ffdcf690-5e68-4f1a-b176-afa2edd49817)

The update appears to only be temporary and isn't reflected on the server. This might be related to access control or atleast related to the backend or might just not be sending the change or etc....

JoachimTislov:
I think its not possible to revert a released review. If this is correct, then `Released` should be disabled.

JosteinLindhom:
It should be possible, and the current behaviour is incorrect.
This is likely related to this issue: https://github.com/quickfeed/quickfeed/issues/570

PR #1306, opened 2025-03-30
Database - UpdateSubmission; Convert to select query

JoachimTislov:
Closes: #1278 

Setting the release to false is now reflected in the database. This was done with GORM solution: https://github.com/quickfeed/quickfeed/issues/569#issuecomment-1013729572

### DIFF
--- a/database/gormdb_submission.go
+++ b/database/gormdb_submission.go
@@ -217,8 +217,9 @@ func (db *GormDB) GetSubmissions(query *qf.Submission) ([]*qf.Submission, error)
 
 // UpdateSubmission updates submission with the given approved status.
 func (db *GormDB) UpdateSubmission(query *qf.Submission) error {
-	// full save associations is required to update the nested grades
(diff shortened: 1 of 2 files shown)


JoachimTislov:
> Using Model(query) relies on the query instance to infer the model, which could lead to unexpected behavior if query is nil or lacks necessary context. Consider explicitly specifying the model type (e.g., Model(&qf.Submission{})) to ensure the correct table is targeted.
> 
> ```
> return db.conn.Session(&gorm.Session{FullSaveAssociations: true}).Model(query).Select("*").Updates(query).Error
> ```

This does not work. Alot of, or all UpdateSubmission tests fail by using this suggestion.

meling (changes requested):
Please add a test the tries to set the release field to false and then run it with the old behavior to check that it fails to save it and fetch it back from the db... and then that this fix fixes it!! That's what we need to do to avoid regressions.

JoachimTislov:
@meling There's a test case for updaing release to false now.

meling (approved):

Issue #1302, opened 2025-03-28
NavbarActiveCourse - CourseFavoriteButton does not change/re-render onclick


JoachimTislov:
![Image](https://github.com/user-attachments/assets/5238bd1e-09b3-4b7f-9f53-5615dc816ed5)

Should be toggled between hollow and filled color.

PR #1309, opened 2025-03-30
NavbarActiveCourse - Add useAppState() to ensure Overmind re-renders the component when the state changes

JoachimTislov:
Resolves: #1302 

This can be viewed as a temporary solution to the issue, as Im not sure why its not re-rendering when the `enrollment.state` is modified. The course page is re-rendered when clicking the same button, but I think that is because of local state or a different part of the state is modified, Im not quite sure yet.

Update: I did as @JosteinLindhom mentioned; https://github.com/quickfeed/quickfeed/pull/1309#issuecomment-2775665875, but I dont assign useAppState to anything as this causes Webpack to trigger an error because the variable is not used... and ends up not building the ui because of it.


JoachimTislov:
Master:

gifPR-1309-master

This PR:

gifPR-1309

JosteinLindhom:
Seems like the component `NavBarActiveCourse` does not get tracked by Overmind correctly when fetching the enrollment. Similarly, the `CourseFavoriteButton` is not tracked either when passing the non-tracked enrollment to it.

Haven't been able to figure out why though.

Fetching the enrollment inside `CourseFavoriteButton` based on the passed enrollment will however cause Overmind to correctly track it, i.e.

```ts
    const state = useAppState()
    // we could pass only the courseID to the component and find the one with matching courseID
    const enrollment = state.enrollments.find(e => e.ID === <the passed enrollment>.ID)
    if (!enrollment) {
        return null
    }
```

JosteinLindhom:
> Seems like the component `NavBarActiveCourse` does not get tracked by Overmind correctly when fetching the enrollment. Similarly, the `CourseFavoriteButton` is not tracked either when passing the non-tracked enrollment to it.
> 
> Haven't been able to figure out why though.
> 
> Fetching the enrollment inside `CourseFavoriteButton` based on the passed enrollment will however cause Overmind to correctly track it, i.e.
> 
> ```ts
>     const state = useAppState()
>     // we could pass only the courseID to the component and find the one with matching courseID
>     const enrollment = state.enrollments.find(e => e.ID === <the passed enrollment>.ID)
>     if (!enrollment) {
>         return null
>     }
> ```

Turns out simply adding this line will make the component re-render correctly:

```ts
const CourseFavoriteButton = ({ enrollment, style }: { enrollment: Enrollment, style: React.CSSProperties }) => {
    // added state, but not used
++  const state = useAppState() <-----
    const actions = useActions()
    return (
        <span style={style} title="Favorite or unfavorite this course. Favorite courses will appear on your dashboard.">
            <i className={isVisible(enrollment) ? 'fa fa-star' : "fa fa-star-o"}
                onClick={() => actions.setEnrollmentState(enrollment)} />
        </span>
    )
}
```

Weird...

PR #1310, opened 2025-03-30
Add helpful launch profiles for debugging

JoachimTislov:
A launch profile is very useful for debugging, as it allows for developers to view the workflow and analyze different sections while the program is running. This is much more reliable and efficient than printing in-memory values to the console or the standard output.

"Launch Quickfeed" will let developers start Quickfeed in debug mode for both frontend and backend, which can in the future be combined with the Docker container, if that is diserable.

A user can run Quickfeed either directly or through the created binary or executable.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,8 +8,8 @@
             // Compound configuration to run both frontend and backend debug sessions
             "name": "Launch Quickfeed",
             "configurations": [
-                "Debug Frontend",
-                "Debug Backend - Quickfeed"
+                "Launch Quickfeed Frontend",
+                "Launch Quickfeed Backend"
             ],
         }
     ],
@@ -22,8 +22,7 @@
             "program": "${fileDirname}"
         },
         {
-            // Start a debugging session for the frontend
-            "name": "Debug Frontend",
+            "name": "Launch Quickfeed Frontend",
             "type": "chrome", // Others: "msedge", "firefox"
             "request": "launch",
             // The url is here for convenience, as you can also input the URL in the browser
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -3,13 +3,46 @@
...
+            "name": "Launch Quickfeed",
+            "configurations": [
+                "Debug Frontend",
+                "Debug Backend - Quickfeed"
+            ],
+        }
+    ],
     "configurations": [
         {
             "name": "Launch Package",
```
meling, line 18: What is this doing? Is it used for quickfeed?

meling (changes requested):
Haven't used these things much in the past... but it looks useful.

.vscode/launch.json
JoachimTislov: Im, note sure... but appears to be used to launch a certain go program

.vscode/launch.json
```diff
@@ -3,13 +3,46 @@
...
+            "type": "chrome", // Others: "msedge", "firefox"
+            "request": "launch",
+            // The url is here for convenience, as you can also input the URL in the browser
+            "url": "https://localhost:443", // Update PORT as needed. See .env file
+            "webRoot": "${workspaceFolder}/public",
+        },
+        {
+            // Start a debugging session for the backend
+            // Launches Quickfeed in dev and watch mode
+            // Debug Quickfeed directly by switching mode to "auto" and set program to "${workspaceFolder}/main.go"
```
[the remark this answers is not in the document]
JoachimTislov: > What's the benefit of auto vs exec?

You aren't able to run the Quickfeed binary without specifying the mode as `exec`. The mode `auto` isn't capable of determining the binary as a valid program to run. You will receive this error if you attempt to do so;

image

>  In what sense is this debugging? Debugging is when you run a debugger like gdb or delve

This runs delve in the vscode interface and allows you to interact and study what the system is doing by traversing with the usual delve functionalties, by clicking buttons instead of wrriting commands in a CLI.

This video walks through it; https://youtu.be/3HiLLByBWkg

The `dlv` binary is used;

image

> Does it make sense to have two alternatives?

I think its nice to inform the user of the different possibilites, and no, not really needed, but one might prefer one or the other. I think running Quickfeed directly is best, as you never need to build a binary or executable

meling (approved):

Issue #1218, opened 2025-02-11
suggestion(ui): consider changing build tool / bundler


JosteinLindhom:
Webpack is not the fastest, and it can be quite a chore to configure.

There are many alternatives that may be faster and / or easier to configure.

I stumbled upon this one, which has a Go API as well: https://github.com/evanw/esbuild

JoachimTislov:
> I stumbled upon this one, which has a Go API as well: https://github.com/evanw/esbuild

Esbuild would be a great addition to Quickfeed, improving both speed and compatibility with Golang. The API will be quite useful when compiling the frontend in the backend, converting from a CLI execution to a customizable API.

Esbuild is a reliable library used by 8.6 miliion people and has currently 110 contributors, quote "An extremely fast bundler for the web". I think any Go developer would be eager to switch from [Webpack](https://webpack.js.org/) to Esbuild.

Current usage of  the Webpack:

```makefile
ui: version-check
	@echo "Running npm ci and webpack"
	@cd public; npm ci; webpack

ui-update: version-check
	@echo "Running npm install and webpack"
	@cd public; npm i; webpack

webpack-dev-server:
	@cd public && npx webpack-dev-server --config webpack.config.js --port 8082 --progress --mode development
```

- /internal/reload/watcher.go: [webpack](https://github.com/quickfeed/quickfeed/blob/master/internal/reload/watcher.go#115)
- web/manifest/manifest.go: [runWebpack](https://github.com/quickfeed/quickfeed/blob/master/web/manifest/manifest.go#257)
- Related configuration to prevent webpack from running in tests, [runWebpack - boolean](https://github.com/quickfeed/quickfeed/blob/master/web/manifest/manifest.go#67)
- [Configuration file](https://github.com/quickfeed/quickfeed/blob/master/public/webpack.config.js)
- package.json;
```
"scripts": {
    "build": "webpack",
    "build:watch": "webpack --watch --mode development",
    "build:dev": "webpack --mode development"
}
```

The conversion appears to be quite straight forward;
- Naturally uninstall webpack and install esbuild
- Create an API for building the frontend for both manifest and the watcher 
- Alter the makefile to execute esbuild instead of webpack, and maybe use the API here as well, would be quite nice to centralize the esbuild usage to one API, resulting in easier maintenance.
- Write similar configurations in webpack.config.js

PR #1313, opened 2025-04-01
Esbuild for development and production

JoachimTislov:
Resolves: #1218 

Update: Esbuild is now ready for production 🎉 

Esbuild is only configured to run in the watcher instead of webpack. 

The metafile which is set to true and can be analyzed. [See metafile docs](https://esbuild.github.io/api/#metafile). Im not sure if that can be of any use.

This the best I managed to reduce the files sizes with splitting, minizing and bundling. When esbuild manages to remove the unused chunks and combine existing ones into larger ones, can esbuild be considered for use in production.

esbuild-prod

Webpack;

```console
index (1.01 MiB)
      npm.react.377fb85e61fa3e447a98.bundle.js
      npm.babel.915c77b4d68e22046833.bundle.js
      .....
  overmind (456 KiB)
      npm.bufbuild.192f88dc568548d3fa9a.bundle.js
      npm.connectrpc.682d9332bbd710b72b6f.bundle.js
      .....
  shared (387 KiB)
      runtime.0b2de02907ff03f71a52.bundle.js
      npm.bufbuild.192f88dc568548d3fa9a.bundle.js
      ....
```

Esbuild compiles in around 350ms while webpack use around 12 seconds, and that is what matters in development.


JoachimTislov:
That wasn't a good edit Copilot.. https://github.com/quickfeed/quickfeed/pull/1313/commits/800c4dc9b643f34ba1fc21930a6866626ec2c560

JoachimTislov:
> Reducing the reload delay from 500ms to 100ms might cause premature reloads in environments with slower response times. Please verify that 100ms is sufficient for reliable functionality.

Im not sure, but it worked Ok

internal/ui/esbuild.go
```diff
@@ -0,0 +1,160 @@
...
+// Watch starts a watch process for the UI, rebuilding on changes
+// The log level is set to info, so only warnings and errors are shown
+func Watch(ch chan<- error) {
+	ctx, err := api.Context(getOptions(nil))
+	defer ctx.Cancel()
+	if err != nil {
+		ch <- fmt.Errorf("failed to create build context: %v", parseMessages(err.Errors))
+		return
+	}
+	if err := ctx.Watch(api.WatchOptions{}); err != nil {
```
JosteinLindhom, line 146: If you edit some `ts` code in `-watch` mode I expect the related rebuilt file and it's hash to change.
I don't see the `index.html` file being updated anywhere here.

i.e `index.html` imports a script, `index-SOMEHASH.js`. After changing code, the file `index-SOMEHASH.js` will become `index-SOMENEWHASH.js`, while `index.html` will still refer to the former it seems.

internal/ui/esbuild.go
```diff
@@ -0,0 +1,160 @@
...
+	// Dynamically add entry points to the build options
+	// Adding more can reduce the size of the output files, but will increase the file count
+	entryPoints := []string{
+		"src/index.tsx",
+		"src/App.tsx",
+		"src/overmind/state.tsx",
+		"src/overmind/actions.tsx",
+		"src/overmind/effects.tsx",
+		"src/overmind/index.tsx",
+		"src/components/Results.tsx",
```
JosteinLindhom, line 50: Why is `Results.tsx` a separate entrypoint?

Makefile
```diff
@@ -35,11 +35,13 @@ endif
 
 ui: version-check
 	@echo "Running npm ci and webpack"
-	@cd public; npm ci; webpack
+	@cd public; npm ci
+	@go run cmd/esbuild/main.go
 
 ui-update: version-check
 	@echo "Running npm install and webpack"
```
JosteinLindhom, line 42:
```sh
@echo "Running npm install and esbuild"
```

JosteinLindhom:
Seems like most of the 700+ files created were related to `react-syntax-highlighter`.

Removing that dependency reduces the number of files to 16(!) and the size of the `dist` folder goes from 11 MB on master to 6.6 MB *before minifying*.

We should also consider not inlining the source maps to further reduce the size for users not interested in debugging code.

internal/ui/esbuild.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	var entries []string
+	for _, entry := range entryPoints {
+		entries = append(entries, fmt.Sprintf("%s/%s", env.PublicDir(), entry))
+	}
+	buildOptions.EntryPoints = entries
+	/*
+		Production options
+
+		These can be added for development, but is not recommended.
```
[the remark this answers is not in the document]
JoachimTislov: I added this statement because minizing the compiled files makes it harder to debug? See Esbuild docs: "..which means it downloads faster but is harder to debug"; [Minify](https://esbuild.github.io/api/#minify)

Should we minize the files for development as well?

internal/ui/esbuild.go
JoachimTislov: This reduced the total file size of the main compiled files, when I did some testing. I can do some secondary testing and confirm it

JosteinLindhom:
> Seems like most of the 700+ files created were related to `react-syntax-highlighter`.
> 
> Removing that dependency reduces the number of files to 16(!) and the size of the `dist` folder goes from 11 MB on master to 6.6 MB _before minifying_.
> 
> We should also consider not inlining the source maps to further reduce the size for users not interested in debugging code.

With external source maps the output is as follows:

```
 public/dist/index-UJEUN7QL.js        889.5kb
  public/dist/chunk-GFLYL2GX.js        379.3kb
  public/dist/chunk-YSZUEZ2D.js        354.6kb
  public/dist/chunk-AR6QXTF4.js        103.7kb
  public/dist/chunk-JITCGBRZ.js         67.4kb
  public/dist/chunk-VYODSVZE.js         28.4kb
  public/dist/chunk-OCOOIGCT.js         28.0kb
  public/dist/chunk-R4QRDZEV.js         18.9kb
  public/dist/chunk-WGQ6F7RE.js         14.1kb
  public/dist/index-HOG377CO.css         9.5kb
  public/dist/actions-B3Y2ZKIE.js        2.6kb
  public/dist/App-Q3436H4I.js            300b
  public/dist/index-SV6IKC6V.js          299b
  public/dist/Results-5EZ7QGEO.js        278b
  public/dist/effects-UMRTIZCF.js        183b
  public/dist/state-SDPUON4Q.js           97b
  public/dist/index-UJEUN7QL.js.map      1.3mb
  public/dist/chunk-GFLYL2GX.js.map    873.9kb
  public/dist/chunk-YSZUEZ2D.js.map    739.8kb
  public/dist/chunk-AR6QXTF4.js.map    170.7kb
  public/dist/chunk-JITCGBRZ.js.map    143.4kb
  public/dist/chunk-VYODSVZE.js.map     97.5kb
  public/dist/chunk-R4QRDZEV.js.map     94.6kb
  public/dist/chunk-OCOOIGCT.js.map     54.7kb
  public/dist/chunk-WGQ6F7RE.js.map     31.6kb
  public/dist/index-HOG377CO.css.map    18.4kb
  public/dist/App-Q3436H4I.js.map         93b
  public/dist/Results-5EZ7QGEO.js.map     93b
  public/dist/actions-B3Y2ZKIE.js.map     93b
  public/dist/effects-UMRTIZCF.js.map     93b
  ...and 2 more output files...
```

Where clients only fetch files ending in `js` unless inspecting the website.

We should probably use this option:

```go
buildOptions.Sourcemap = api.SourceMapLinked
```

Such that the source maps show up in `Sources`:

image

JosteinLindhom:
Enabling these further reduces the size of our bundled code:

```go
buildOptions.MinifyWhitespace = true
buildOptions.MinifyIdentifiers = true
buildOptions.MinifySyntax = true
```

Which should not cause any issues as we provide source maps to those who may wish to inspect readable code.

Size of bundled code with the options above:

```
  public/dist/index-U4E2PNZA.js        173.2kb
  public/dist/chunk-XMNYUSJ6.js        146.9kb
  public/dist/chunk-LXPNNTSA.js        134.1kb
  public/dist/chunk-FADCKEUP.js         52.6kb
  public/dist/chunk-7TMT2RJW.js         31.5kb
  public/dist/chunk-WJNJSJK7.js         14.9kb
  public/dist/chunk-B5DTDTUF.js         13.7kb
  public/dist/chunk-YP5ROZ5M.js          9.6kb
  public/dist/chunk-HWVKV3L4.js          7.1kb
  public/dist/index-45BNTIIX.css         6.8kb
  public/dist/actions-CTN4MOWX.js        1.7kb
  public/dist/index-LJIIZUHG.js          311b
  public/dist/App-CRHLXU3M.js            294b
  public/dist/Results-FA5ECP3G.js        275b
  public/dist/effects-EBZ7TRXY.js        192b
  public/dist/state-FIXTSOWJ.js          131b
```

JoachimTislov:
@JosteinLindhom Should esbuild now be considered for use in production?

The only remaining issue is related to watch mode, which is only for development. Refering to: https://github.com/quickfeed/quickfeed/pull/1313#discussion_r2022630299

internal/ui/esbuild.go
JoachimTislov: > as I had to manually insert them into the html file... after a build

Had to write code to do it*

JoachimTislov:
> I see no reason not to use it for both development and production at this point.

What about watch mode, isn't that an issue?

JoachimTislov:
Following is produced when doing changes in watch mode;

```console
[watch] build started (change: "public/src/Helpers.ts")
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/App-PNFHQ2DK.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/index-IY3MTJFY.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-BY7365N3.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/state-5J7EUEJX.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/state-5J7EUEJX.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-BY7365N3.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-V7OPYV5L.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-V7OPYV5L.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-K2A57MTU.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-OKOMPAWG.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/effects-HNKTHDXB.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/effects-HNKTHDXB.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-K2A57MTU.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-74A64KIF.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-74A64KIF.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-OHYGXPGH.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-OHYGXPGH.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-FQXSHO57.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-FQXSHO57.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-Z6PX25K4.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/index-YO3VPAPI.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/index-YO3VPAPI.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/App-PNFHQ2DK.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/Results-4TYHH2Q7.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/Results-4TYHH2Q7.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-Z6PX25K4.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/actions-3EDXNSUN.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/actions-3EDXNSUN.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/chunk-OKOMPAWG.js
2025/04/01 17:41:20 file changed: /home/yoa/quickfeed/public/dist/index-IY3MTJFY.js
```

internal/reload/watcher.go:47
```go
func (w *Watcher) start(ctx context.Context) {
	for {
		select {
		case event, ok := <-w.fsWatcher.Events:
			if !ok {
				return // Watcher closed
			}
			// We only care about writes and creates.
			if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) {
				// Broadcast event to all clients.
				w.broadcastMessage(event.Name)
added this -->		log.Println("file changed:", event.Name)
			}
		case err := <-w.fsWatcher.Errors:
			log.Println("error watching files:", err)
		case <-ctx.Done():
			w.fsWatcher.Close()
			return
		}
	}
}
```

This can be parsed and inserted into the index.html file. And reload can delayed just enough so the client gets the correct index.html

This should resolve the remaining issue; https://github.com/quickfeed/quickfeed/pull/1313#discussion_r2022630299

internal/ui/esbuild.go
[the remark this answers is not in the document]
JosteinLindhom: As long as we have source maps, minifying the code is fine.

If an error occurs, the error will be mapped to the source in our source map, not negatively affecting the debugging experience.

As commented elsewhere we should keep the source maps separate, but linked.

JosteinLindhom:
> This can be parsed and inserted into the index.html file. And reload can be delayed just enough so the client gets the correct index.html
> 
> This should resolve the remaining issue; [#1313 (comment)](https://github.com/quickfeed/quickfeed/pull/1313#discussion_r2022630299)

I suggest we use the plugin system for this.
I see we can create a plugin that runs `OnEnd`, which seem to be invoked after each rebuild.

I made a simple plugin that seem to work:

```go
Plugins: []api.Plugin{
	{
		Name: "HTML Plugin",
		Setup: func(setup api.PluginBuild) {
			setup.OnEnd(func(result *api.BuildResult) (api.OnEndResult, error) {
				if len(result.Errors) > 0 {
                                        // Could log the errors here as well, not sure if they are logged by
                                        // esbuild itself
					return api.OnEndResult{}, parseMessages(result.Errors)
				}
				createHtml(result.OutputFiles)
				return api.OnEndResult{}, nil
			})
		},
	},
},
```

JoachimTislov:
> I made a simple plugin that seem to work:

Nice!

Thats alot better than this;

```go
        // Watch for changes in the path channel
	// Rebuilds the index.html file and updates the links
	// This code can be optimized, but this works for now

	var cachedCssFile string
	var paths map[string]struct{}
	timer := time.NewTimer(0)
	for {
		select {
		case path, ok := <-pathCh:
			if !ok {
				log.Println("Path channel closed")
				return
			}
			if strings.HasSuffix(path, ".css") {
				cachedCssFile = path
			}
			// Reinitialize between builds to ensure no duplicates
			if paths == nil {
				paths = make(map[string]struct{})
			}
			if _, ok := paths[path]; !ok {
				paths[path] = struct{}{}
			}
			timer.Reset(2 * time.Millisecond)
		case <-timer.C:
			if len(paths) == 0 {
				continue
			}
			// Add all css files to the paths map
			// This is done to ensure that the css files are included in the index.html
			// Css files are not recompiled if they are not changed and thereby not included in rebuilds
			paths[cachedCssFile] = struct{}{}

			// Recreate index.html with the new links
			createHtml(paths)
			paths = nil // Reset paths to nil to ensure no duplicates
		}
	}
```

Code receives from the channel i commented earlier inside of the watcher method. This works fine, but the watcher broke with this solution...

JosteinLindhom:
We should rip out webpack related stuff in this PR as well.
`manifest.go` has some references to webpack and should be updated.

JoachimTislov:
I didn't remove the webpack-dev-server target because the selenium test depends on it?

```makefile
webpack-dev-server:
	@cd public && npx webpack-dev-server --config webpack.config.js --port 8082 --progress --mode development

# TODO Should check that webpack-dev-server is running.
selenium:
	@cd public && npm run test:selenium
```

 The user can install webpack externally use the target..

internal/ui/esbuild.go
```diff
@@ -0,0 +1,161 @@
...
+// Scss files are treated as css
+func Build(outputDir *string, dev bool) error {
+	if err := resetDistFolder(); err != nil {
+		return fmt.Errorf("failed to reset dist folder: %v", err)
+	}
+	result := api.Build(getOptions(outputDir, dev))
+	if len(result.Errors) > 0 {
+		return fmt.Errorf("failed to build UI: %v", parseMessages(result.Errors))
+	}
+	if err := createHtml(result.OutputFiles); err != nil {
```
JosteinLindhom, line 107: With the plugin HTML plugin this is no longer necessary here. Otherwise it will run twice, once here and once by the plugin.

internal/ui/esbuild.go
```diff
@@ -0,0 +1,161 @@
...
+					return api.OnEndResult{Errors: result.Errors, Warnings: result.Warnings}, err
+				})
+			},
+		},
+	},
+}
+
+// resetDistFolder removes the dist folder and creates a new one
+func resetDistFolder() error {
+	path := fmt.Sprintf("%s/dist", env.PublicDir())
```
JosteinLindhom, line 77: Can't this be replaced with `public("dist")`?

internal/ui/esbuild.go
```diff
@@ -0,0 +1,161 @@
...
+	if err := os.MkdirAll(path, 0o755); err != nil {
+		return fmt.Errorf("failed to create dist directory: %v", err)
+	}
+	return nil
+}
+
+// createHtml creates the index.html file from the index.tmpl.html template
+// Injects file links into the index template
+func createHtml(outputFiles []api.OutputFile) error {
+	file, err := os.Create(fmt.Sprintf("%s/assets/index.html", env.PublicDir()))
```
JosteinLindhom, line 92: Can use `public("assets/index.html")`

internal/ui/esbuild.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read index.html: %v", err)
+	}
+	funcMap := template.FuncMap{
+		"ext":  filepath.Ext,
+		"base": filepath.Base,
+	}
+	tmpl, err := os.ReadFile(fmt.Sprintf("%s/index.tmpl.html", env.PublicDir()))
```
JosteinLindhom, line 100: `public("index.tmpl.html")`

internal/ui/esbuild.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+var public = func(s string) string {
+	return fmt.Sprintf("%s/%s", env.PublicDir(), s)
+}
+
+// buildOptions defines the build options for esbuild
+// The api has write access and writes the output to public/dist
+var buildOptions = api.BuildOptions{
+	Outdir: fmt.Sprintf("%s/dist", env.PublicDir()),
```
JosteinLindhom, line 22: `public("dist")`

web/manifest/manifest.go
```diff
@@ -59,24 +60,24 @@ func CreateNewQuickFeedApp(srvFn web.ServerType, envFile string) error {
...
-		runWebpack: true,
+		domain:  domain,
+		envFile: envFile,
+		client:  github.NewClient(nil),
+		done:    make(chan error),
+		buildUI: true,
 	}
 	router := http.NewServeMux()
-	router.Handle("/manifest/callback", m.conversion())
+	router.Handle("/manifest/callback", m.conversion(dev))
```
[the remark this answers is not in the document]
meling: Can this function be tested?

web/manifest/manifest.go
```diff
@@ -59,24 +60,24 @@ func CreateNewQuickFeedApp(srvFn web.ServerType, envFile string) error {
...
-	handler    http.Handler
-	done       chan error
-	client     *github.Client // optional, for testing
-	runWebpack bool           // run webpack only for production
+	domain  string
+	envFile string
+	handler http.Handler
+	done    chan error
+	client  *github.Client // optional, for testing
+	buildUI bool           // build UI is only for production
```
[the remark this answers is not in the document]
meling: I'm wondering if the comment is a bit misleading now... I think the current semantics is that it builds the UI for production and development mode, but it should return nil for testing on GitHub. Do I have that right?

meling (changes requested):
I did another pass on this. There are a few issues that needs to be fixed...

internal/ui/esbuild.go
```diff
@@ -0,0 +1,161 @@
...
+	Outdir: public("/dist"),
+	EntryPoints: []string{
+		public("src/index.tsx"),
+		public("src/App.tsx"),
+		public("src/overmind/state.tsx"),
+		public("src/overmind/actions.tsx"),
+		public("src/overmind/effects.tsx"),
+		public("src/overmind/index.tsx"),
+		public("src/components/Results.tsx"),
+		// Adding more can reduce the size of the output files, but will increase the file count
```
JosteinLindhom, line 31: Adding more will not reduce the total size of our bundle, but it can decrease the size of the individual chunks at the cost of creating more chunks.

Analyzing the metafile gives this output:
image

The boxes that are red/purple-ish is our code. Everything else is dependencies.
With webpack we managed to split our dependencies into independent chunks.

I think by using different entrypoints we should be able to have chunks split into
- dependency chunks (these are more static)
- our own code (changes more often)

As an example, if `Helpers.ts` changes with the current endpoints, users will have to fetch a chunk that also includes:
- bufbuild/protobuf
- react-router

Even though these may not have changed at all.

internal/ui/esbuild.go
```diff
@@ -0,0 +1,161 @@
...
+		},
+	},
+}
+
+// resetDistFolder removes the dist folder and creates a new one
+func resetDistFolder() error {
+	path := fmt.Sprintf("%s/dist", env.PublicDir())
+	if _, err := os.Stat(path); err == nil {
+		if err := os.RemoveAll(path); err != nil {
+			return fmt.Errorf("failed to remove dist directory: %v", err)
```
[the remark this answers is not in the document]
meling: That said, it would probably be helpful to provide an identifying extra message to help debugging a problem. For example, the error message could be `("ui: failed to remove: %w", err)`... (depending on what the underlying error message will be)...

For `MkdirAll` below, perhaps: `("ui: failed to create: %w", err)` ... (depending on what the underlying error message will be)...

internal/ui/esbuild.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// Watch starts a watch process for the frontend, rebuilding on changes
+func Watch(ch chan<- error) {
+	ctx, err := api.Context(getOptions(nil, true))
+	if err != nil {
+		ch <- fmt.Errorf("failed to create build context: %v", parseMessages(err.Errors))
+		return
+	}
+	if err := ctx.Watch(api.WatchOptions{}); err != nil {
```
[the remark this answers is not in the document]
JoachimTislov: Its unclear to me as well, but I found [one reference of watcher.stop() inside of ctx.Dispose()](https://github.com/evanw/esbuild/blob/4475787eef4c4923b92b9fa37ebba1c88b9e1d9b/pkg/api/api_impl.go#L1146).

internal/ui/esbuild.go
```diff
@@ -0,0 +1,161 @@
...
+	Plugins: []api.Plugin{
+		{
+			Name: "Reset Plugin",
+			Setup: func(setup api.PluginBuild) {
+				setup.OnStart(func() (api.OnStartResult, error) {
+					err := resetDistFolder()
+					if err != nil {
+						log.Printf("failed to reset dist folder: %v", err)
+					}
+					return api.OnStartResult{}, err
```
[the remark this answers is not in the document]
JoachimTislov:
```go
if resetDistFolder() != nil {
    return api.OnStartResult{
        Warnings: []api.Message{
            {
                PluginName: "Reset",
                Text:       "Failed to clear the dist folder",
                Notes: []api.Note{
                    {
                        Text: fmt.Sprintf("The dist directory may now contain multiple builds\nLocation: %s", public("dist")),
                    },
                },
            },
        },
    }, nil
}
return api.OnStartResult{}, nil
```

This code outputs;

image

internal/ui/esbuild.go
```diff
@@ -0,0 +1,161 @@
...
+				})
+			},
+		},
+		{
+			Name: "HTML Plugin",
+			Setup: func(setup api.PluginBuild) {
+				setup.OnEnd(func(result *api.BuildResult) (api.OnEndResult, error) {
+					err := createHtml(result.OutputFiles)
+					if err != nil {
+						log.Printf("failed to create index.html: %v", err)
```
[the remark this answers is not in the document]
JoachimTislov: image

web/manifest/manifest.go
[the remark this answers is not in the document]
JoachimTislov: I think it has been misleading for a long time, as developers made their Github App, they would naturally build the UI. But yes, this is only here to prevent tests from building the UI.

`buildUI: false, // Disable esbuild for testing`

web/manifest/manifest.go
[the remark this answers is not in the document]
JoachimTislov: > Can this function be tested?

I tried, and it seems quite tricky and not worth it. Considering build can be any func which returns an error and the command "npm ci" can fail for many reasons and is difficult to force to fail, and in the same test do a fix and make it pass....

Testing it offers minimal benefits compared to the effort required, as users may experience different results due to the environment-dependent nature of "npm ci"...

I stopped at this error;

`2025/04/05 18:13:51 Running npm ci...
2025/04/05 18:13:54 [] exec: already started
2025/04/05 18:13:54 exit status 45
2025/04/05 18:13:54 Failed to run npm ci; giving up
--- FAIL: TestBuildUI (7.49s)
    --- FAIL: TestBuildUI/Build_returns_not_nil (7.49s)
        manifest_internal_test.go:242: Expected <nil> got failed to run npm ci: error
FAIL`


With corresponding test;

```go
func TestBuildUI(t *testing.T) {
	tests := []struct {
		name     string
		manifest *Manifest
		wantErr  error
	}{
		{name: "Build returns nil", manifest: &Manifest{build: func() error { return nil }}, wantErr: nil},
		{name: "Build returns not nil", manifest: &Manifest{build: func() error { return errors.New("error") }}},
	}
	for _, test := range tests {
		t.Run(test.name, func(t *testing.T) {
			err := test.manifest.buildUI()
			if err != test.wantErr {
				t.Errorf("Expected %v got %v", test.wantErr, err)
			}
		})
	}
}
```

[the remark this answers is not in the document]
JoachimTislov: Also, you need to replace the "public" assignment in `runNpmCi` with `env.PublicDir()` since the test isn't in quickfeed's root directory;

`c.Dir = env.PublicDir()`

internal/ui/esbuild.go
JoachimTislov: This is the best I've managed to configure the splitting;

image

The `Comment.tsx` component use `react-markdown` which use a bunch of libraries and therefore is the Comment an entrypoint, to group everything in one chunk. Top middle one.

Rest of the entry points are there for a good reason. I've reviewed them multiple times, and as of now, these are the optimal entry points.

JoachimTislov: There are a few small chunks that I would have liked to remove, but doing so affects the splitting of the other chunks.

meling (approved):

Issue #1231, opened 2025-03-07
fix: protect against nil pointer issue in getNextReviewer


meling:
We should fix this:
https://github.com/quickfeed/quickfeed/blob/b27cb42af5d059c52a19b06f3723d3e39e442347/assignments/pull_requests.go#L61

As mentioned [here](https://github.com/quickfeed/quickfeed/pull/1193/files#r1955757844).

PR #1321, opened 2025-04-06
getNextReviewer - Resolve potential panic

JoachimTislov:
Fixes: #1231 

@meling No, I don't want to write a test for AssignReviewers. The code written is quite confusing and suprisingly complex for the given functionality (Update: Im refering to getNextReviewer), plus is this code even used? 

AssignReviewers call hierachy:

image 

Only called here, in `handlePullRequestPush`

```go
if taskSum >= scoreLimit && !pullRequest.HasReviewers() {
  wh.logger.Debugf("Assigning reviewers to pull request #%d, in repository: %s", prNumber, repoName)
  if err := assignments.AssignReviewers(ctx, scmClient, wh.db, rd.Course, rd.Repo, pullRequest); err != nil {
	  wh.logger.Errorf("Failed to assign reviewers to pull request: %v", err)
	  return
  }
}
```

### DIFF
--- a/assignments/pull_requests.go
+++ b/assignments/pull_requests.go
@@ -36,6 +36,9 @@ func AssignReviewers(ctx context.Context, sc scm.SCM, db database.Database, cour
 	if err != nil {
 		return err
 	}
+	if teacherReviewer == nil || studentReviewer == nil {
+		return fmt.Errorf("failed to get reviewers for pull request %d", pullRequest.GetNumber())
+	}
 
 	opt := &scm.RequestReviewersOptions{
 		Organization: course.GetScmOrganizationName(),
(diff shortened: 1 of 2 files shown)


meling (approved):
> @meling No, I don't want to write a test for AssignReviewers. The code written is quite confusing and suprisingly complex for the given functionality (Update: Im refering to getNextReviewer), plus is this code even used?

I rewrote the test to use a table-driven approach and added a few cases to expose the index out of bounds... The previous tests for `getNextReviewer` did not catch the problem and it was tricky to expand it to be more useful.

It would be nice to have a test case for `AssignReviewers` but it requires adding a mock entry for the `scm.RequestReviewers()` method. Maybe in the future...

PR #1326, opened 2025-04-10
Resolve reference issue with Overmind - actions.tsx and effects.tsx files

JoachimTislov:
The references for methods and objects inside of actions.tsx and effects.tsx are now detected by the lsp for typescript

Keep in mind, this does not fix the underlying issue. Its simply a workaround to able to view references of the methods and objects...


JoachimTislov:
DeepSource complains about the confirms.... I won't address that issue in this branch.

Issue #1197, opened 2025-01-28
feature: Create a notification system


JoachimTislov:
@JosteinLindhom you mentioned about creating a system like this. I think one case was notifying the students when the teacher have made changes to the main repository. Can you elaborate more on this issue, what is your vision for the notification system?

JoachimTislov:
We can use this package: [go-email](https://github.com/wneessen/go-mail) to notify via mail.

meling:
I’m not a fan of dealing with email sending with all the concerns involved (spam filters etc.) The obvious idea is to use a frontend thing, but that is also not great… Perhaps Discord (we already have the helpbot) or Slack would be more relevant. Or we could create an issue on the student’s GitHub repository. Some users will the also get emails when issues are added to their repo.

JoachimTislov:
Understood

Issue #1131, opened 2024-08-17
ci: how to detect test environment failure?


meling:
This is a question without a solid solution. Sometimes, when creating tests, the test environment on the quickfeed server is slightly different from the local environment of the developer, or the developer forgets to add generated files or similar. This can result in challenging debugging sessions, e.g., due to compile errors that prevent the tests from running and producing useful output.

We could consider catching the exit status of the compilers and other tools from `run.sh` and notify the teacher pushing changes to the `tests` repository in the frontend or send an electric shock directly to his keyboard.

meling:
This issue has become more relevant recently with the new dat640 course being added to the fold and new TAs involved in debugging tests, Docker build failures, and similar issues. Therefore, I propose to resolve this issue in the simplest possible way.

Specifically, I suggest we create a teacher panel on the frontend to display the backend's log output for the selected course. This panel should be available as a clickable tile on the course page for teachers, e.g., `https://uis.itest.run/course/21`. When the teacher clicks this panel, the course log output should be shown. I suggest that we record a per-course log file that we can display to the teacher, possibly with some filter mechanism to allow focusing on the log output produced by different repositories, such as the `tests` repository. I think one approach could be to add a repository key to the logger and pass it down from the webhook-handling logic.

Note: The first version of this does not need a notification system. Instead, we can add a separate RPC method to fetch the full or partial log, e.g., based on dates.

Issue #265, opened 2020-06-04
Warn when student repo is behind the source assignments repo


meling:
Autograder should be able to detect when a student repo or a group repo is behind the original `assignments` repo, and give a warning, along with appropriate git instructions (in the frontend) for merging and resolving any conflicts.

JoachimTislov:
Refer to #1163, The simplest solution would be to create a pull request in all of the students repositories when a teacher pushes to the assignments repository.

meling:
Closing. The alternative is to implement #1497.

PR #1327, opened 2025-04-13
Notification System

JoachimTislov:
Closes: #1197 
Closes: #1131 
Closes: #265

The current design is questionable. Primarly the database solution as theres a many2many relationship between a user and a notification. Its end ups requiring alot of additional code (logic) to make it work, which I don't like, but I do not have a better solution atm, other than removing the "read" feature and store the userIDs as json in the notification table.

I currently need to update the functionaliy and design in the frontend and implement the predefined templates and integrate the notification system internally for certain events like for 265 - when a Teacher pushes to the assignment repository.


JoachimTislov:
Closing this PR since it requires to much combined work to be merged.

Issue #1291, opened 2025-03-27
Selected lab isn't displayed after a reload in Review Assignments


JoachimTislov:
The review/lab with id: 2213 should be displayed, but it isn't.

![Image](https://github.com/user-attachments/assets/ab616d6b-354b-4a16-937c-75600ada4fe7)

JoachimTislov:
I don't think that picture is accurate. Selected submission is null after a reload and displaying a review will therefore be difficult. I think this happens because of some other reason. I can't manage to reproduce this atm.

JosteinLindhom:
Closing as fixed in #1344

PR #1328, opened 2025-04-14
Render selected submission/lab after refresh

JoachimTislov:
Closes: #1291

### DIFF
--- a/public/src/components/Results.tsx
+++ b/public/src/components/Results.tsx
@@ -38,20 +38,18 @@ const Results = ({ review }: { review: boolean }) => {
         }
     }, [])
 
-    useEffect(() => {
-        if (!state.selectedSubmission) {
-            // If no submission is selected, check if there is a selected lab in the URL
-            // and select it if it exists
-            const selectedLab = new URLSearchParams(location.search).get('id')
-            if (selectedLab) {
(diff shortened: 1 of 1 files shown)


JoachimTislov:
> Copilot reviewed 1 out of 1 changed files in this pull request and generated no comments.
> 
> Comments suppressed due to low confidence (1)

I don't see why is should be inside of a useEffect hook to begin with. It should only be exectued once if there isn't a selected submission. Im not sure how the useEffect hook breaks the logic.

JoachimTislov:
PR-1291

JoachimTislov:
The submission isn't rendered since it is not present in either group or user submissions when the useEffect runs. Im don't how this race condition occurs.. but will look into it later. 

```typescript
   ByID(id: bigint): Submission | undefined {
        for (const submissions of this.userSubmissions.values()) {
            const submission = submissions.submissions.find(s => s.ID === id)
            if (submission) {
                return submission
            }
        }
        for (const submissions of this.groupSubmissions.values()) {
            const submission = submissions.submissions.find(s => s.ID === id)
            if (submission) {
                return submission
            }
        }
        return undefined
    }
```

JoachimTislov:
This have been resolved [here](https://github.com/quickfeed/quickfeed/pull/1344/commits/cd3f1c5b491341e3007e5a276c047def87079589)

Issue #1277, opened 2025-03-27
% and (r) is placed below number when width is too small


JoachimTislov:
This happens in the DynamicTable, but the issue might be located else where as well.

![Image](https://github.com/user-attachments/assets/17b47274-e8ad-48ed-9b7c-98f9a98b0df7)

![Image](https://github.com/user-attachments/assets/373def2e-ef16-4c4b-834d-d220892a9667)

![Image](https://github.com/user-attachments/assets/99feb974-4577-4d37-8d7d-c7d04db892f0)

DynamicTable rowCell;

```typescript
    const rowCell = (cell: RowElement, index: number) => {
        if (isCellElement(cell)) {
            const element = cell.link ? <a href={cell.link} target={"_blank"} rel="noopener noreferrer">{cell.value}</a> : cell.value
            return <td key={index} className={cell.className} onClick={cell.onClick}>{element}</td>
        }
        return index == 0 ? <th key={index}>{cell}</th> : <td key={index}>{cell}</td>
    }
```

PR #1329, opened 2025-04-14
Prevent inline wrapping when the width is too small in table elements

JoachimTislov:
Closes: #1277


JoachimTislov:
Master:

PR-1277-master (1)

This branch;

PR-1277-branch

public/src/components/submissions/SubmissionScore.tsx
```diff
@@ -14,15 +14,15 @@ const SubmissionScore = ({
 
     return (
         <tr>
-            <td className={`${className} pl-4`}>{score.TestName}</td>
-            <td className="text-right">
+            <td className={`${className} pl-4 nowrap-whitespace`}>{score.TestName}</td>
+            <td className="text-right nowrap-whitespace">
```
meling, line 18: Can you try this instead? (I didn't double-check every thing below...)
```tsx
  <td className="fixed-width-score">
    {score.Score}/{score.MaxScore}
  </td>

  <td className="fixed-width-percent">
    <span className={cellColor}>
      {percentage.toFixed(1)}%
    </span>
  </td>

  <td className="fixed-width-percent">
    <span
      style={{ opacity: 0.5 }}
      title={`Weight: ${score.Weight}`}
      aria-label={`Max weighted percentage is ${maxPercentage.toFixed(1)} percent, weight ${score.Weight}`}
    >
      {maxPercentage.toFixed(1)}%
    </span>
  </td>
```

```css
.fixed-width-score {
  width: 8ch;
  text-align: right;
  font-family: monospace;
  white-space: nowrap;
}

.fixed-width-percent {
  width: 7ch;
  text-align: right;
  font-family: monospace;
  white-space: nowrap;
}
```

JoachimTislov: image

JoachimTislov: Wait, that was wrong.

JoachimTislov: image

JoachimTislov: This is with the same css class `fixed-width-percent` for the table headers as well, to align them correctly;

image

JosteinLindhom:
Are there any instances where we would *not* want the `nowrap` on `td` and `th` elements?

If the answer is no, then I suggest making a general CSS rule targeting all `td` and `th` elements with `nowrap`.
This way we don't have to remember to add the class if creating a new table at some point.

Could likely add it here:

```css
.table td {
    padding: 0.5rem !important;
}

.table th {
    padding: 0.5rem !important;
}
```

These should also be nested under a single `.table`.

public/src/components/submissions/SubmissionScore.tsx
meling: The text alignment looks left to me on that image (the three columns to the right)

Issue #1330, opened 2025-04-14
Fix the Submissionstable


JoachimTislov:
The submissions table currently doesn't display any assignments.

![Image](https://github.com/user-attachments/assets/114fef25-9ba7-4528-8879-2d33c8b66a1b)

The following if statement inside of `NewSubmissionsTable` in `SubmissionsTable.tsx` appears to block all of the entries as `!""` is apperantly true

```typescript
  if (!deadline.message) {
      return
  }
```

![Image](https://github.com/user-attachments/assets/77796112-16cd-41e9-be36-4433e0813d55)

This case happens for the assignments because the `const timeToDeadline = timestampDate(deadline).getTime()` value is invalid. It currently doesn't take into consideration of the time for the todays date, resulting in a very large numbers

![Image](https://github.com/user-attachments/assets/47f994fd-6d27-4532-ab92-3f38827f63f7)

The `timeFormatter` therefore returns the last case: `{ className: "", message: "", daysUntil: days }` where the message is an empty string. Be aware the image above is after removing the `if (days < 14)` check around the second to last case.

PR #1331, opened 2025-04-14
Fix the Submissions table

JoachimTislov:
Closes: #1330 

The assignment entries in the table is only there to show what it would look like. These are not shown. I tried to write some dummy data, but that was surprisingly annoying to do...

I commented out `if (date.getFullYear() !== now.getFullYear() || date.getMonth() > now.getMonth() + 1) { return }`.....

image

Please view the related issue for the master branch view..


public/src/Helpers.ts
```diff
@@ -82,33 +81,36 @@ export interface Deadline {
...
-        return { className: "table-danger", message: `${hours} hours and ${minutes} minutes to deadline!`, daysUntil: 0 }
-    }
-
-    if (days < 3) {
-        return { className: "table-warning", message: `${days} day${days === 1 ? " " : "s"} to deadline`, daysUntil: days }
-    }
-
-    if (days < 14) {
-        return { className: "table-primary", message: `${days} days`, daysUntil: days }
+        const expiredSince = daysSince > 0 ? `${daysSince} days ago` : `${hoursSince} hours ago`
```
meling, line 98: This ternary isn't too difficult to read; can be inlined the `message` expression on Lne 101.

public/src/components/dashboard/SubmissionsTable.tsx
```diff
@@ -27,66 +26,75 @@ const SubmissionsTable = () => {
...
+        if (!submissions) {
+            return
+        }
+        if (state.enrollmentsByCourseID[courseID.toString()]?.status !== Enrollment_UserStatus.STUDENT) {
+            return
+        }
+        // Submissions are indexed by the assignment order - 1.
+        const submission = submissions.find(sub => sub.AssignmentID === assignment.ID) ?? create(SubmissionSchema)
+        const status = getStatusByUser(submission, state.self.ID)
+        if (!isApproved(status) && assignment.deadline) {
```
meling, line 42: Don't mix logic related to the deadline guard check and approved status. I think it will make more sense to do the deadline guard separately first (could be done on Line 31 before anything else in the `forEach` loop):
```tsx
if (isExpired(assignment)) {
    return  // ignore expired assignments
}
```

```tsx
// isExpired returns true if the assignment's deadline has expired.
// An assignment is considered expired if its deadline expired more than one month ago.
function isExpired(assignment) {
    const date = timestampDate(assignment.deadline)
    const now = new Date()
    return (
        date.getFullYear() !== now.getFullYear() ||
        date.getMonth() > now.getMonth() + 1
    )
}
```

public/src/components/dashboard/SubmissionsTable.tsx
```diff
@@ -27,66 +26,75 @@ const SubmissionsTable = () => {
...
+                <tr key={assignment.ID.toString()} className={`clickable-row ${deadline.className}`}
+                    onClick={() => history.push(`/course/${courseID}/lab/${assignment.ID}`)}>
+                    <th scope="row">{course?.code}</th>
+                    <td>
+                        {assignment.name}
+                        {assignment.isGroupLab ?
+                            <span className="badge ml-2 float-right"><i className={Icon.GROUP} title="Group Assignment" /></span> : null}
+                    </td>
+                    <td><ProgressBar courseID={courseID.toString()} submission={submission} type={Progress.OVERVIEW} /></td>
+                    <td>{getFormattedTime(assignment.deadline, true)}</td>
```
meling, line 67: Perhaps this should be called internally in `deadlineFormatter` and assigned to a field `{deadline.time}`.

public/src/components/dashboard/SubmissionsTable.tsx
```diff
@@ -27,66 +26,75 @@ const SubmissionsTable = () => {
...
-                            {assignmentStatusText(assignment, submission, status)}
-                        </td>
-                    </tr>
-                )
-            }
-        })
-        return table
-    }
+
+            const deadline = timeFormatter(date, submission.score >= assignment.scoreLimit)
```
meling, line 55: Seems like this should be called `deadlineFormatter` instead, since it returns a `Deadline` interface.

meling: Probably should also revert the `date` input to `assignment.deadline` since `isExpired(assignment)` hides the `date` variable.

meling: Or you could pass the `assignment` and do the `submission.score >= assignment.scoreLimit` comparison internally in `deadlineFormatter`:
```tsx
const deadline = timeFormatter(assignment, submission.score)
```

meling (changes requested):

meling:
The failing test should be easy to fix:
```
FAIL src/__tests__/lab.test.tsx
  ● Test suite failed to run

    src/components/Lab.tsx:50:63 - error TS2339: Property 'NoSubmission' does not exist on type 'typeof KnownMessage'.

    50                 return <CenteredMessage message={KnownMessage.NoSubmission} />
```

public/src/Helpers.ts
```diff
@@ -81,33 +96,38 @@ export interface Deadline {
  *
  * layoutTime = "2021-03-20T23:59:00"
  */
-export const timeFormatter = (deadline: Timestamp): Deadline => {
-    const timeToDeadline = timestampDate(deadline).getTime()
+export const deadlineFormatter = (assignment: Assignment, submissionScore: number): Deadline => {
+    const timeToDeadline = (new Date).getTime() - Date.now()
```
meling, line 100: Shouldn't this use the `assignment.deadline` here?

meling: Do we have tests for this stuff? It would be nice to have.

public/src/components/dashboard/SubmissionsTable.tsx
```diff
@@ -27,66 +26,66 @@ const SubmissionsTable = () => {
...
-                        <th scope="col">Due in</th>
-                        <th scope="col">Status</th>
-                    </tr>
-                </thead>
-                <tbody>
-                    {NewSubmissionsTable()}
-                </tbody>
-            </table>
-        </div>
+        table.length !== 0 ? (
```
meling, line 69: This ternary check is unnecessary. Prefer the guard style when there is a chance that something should return null.
```tax
if (table.length === 0) {
    return null
}
return (
     <div>
...
```

public/src/Helpers.ts
```diff
@@ -69,10 +69,25 @@ export const getFormattedTime = (timestamp: Timestamp | undefined, offset?: bool
     return `${deadline.getDate()} ${months[deadline.getMonth()]} ${deadline.getFullYear()} ${deadline.getHours()}:${zero}${minutes}`
 }
 
+
+// isExpired returns true if the assignment's deadline has expired.
+// An assignment is considered expired if its deadline expired more than one month ago.
+export const isExpired = (assignment: Assignment): boolean => {
+    if (!assignment.deadline) {
```
meling, line 76: What does this actually mean? If the deadline is not set, we return false. Meaning that the deadline hasn't expired, right? Is that the right behavior? 

Would be helpful if there was a test for this.

meling (changes requested):
See revised review comments...

public/src/Helpers.ts
JoachimTislov: > Is that the right behavior?

No mistake, it should return `true`, and the deadline field should in theory always be set. The protobuf.Timestamp is implicitly optional in proto3 (source - chatgpt)

meling (approved):

Issue #1336, opened 2025-04-17
Create Course page does not load because the "process is undefined" (Because of migration from webpack to esbuild)


JoachimTislov:
We currently only define the `process.env.NODE_ENV`, but we also need to define `process.env.QUICKFEED_ORGANIZATION_URL` and `process.env.QUICKFEED_APP_URL`. I think it handles the `NODE_ENV` automatically, and can't register custom environment variables.

Related issue on esbuild: https://github.com/evanw/esbuild/issues/12

PR #1337, opened 2025-04-17
Esbuild: Load the App Url and add key: process.env.QUICKFEED_APP_URL to Define map in build options

JoachimTislov:
Closes: #1336

### DIFF
--- a/internal/env/scm.go
+++ b/internal/env/scm.go
@@ -108,7 +108,3 @@ func HasAppID() bool {
 func GetAppURL() string {
 	return os.Getenv("QUICKFEED_APP_URL")
 }
-
-func GetOrganizationURL() string {
-	return os.Getenv("QUICKFEED_ORGANIZATION_URL")
-}
--- a/internal/env/scm.go
+++ b/internal/env/scm.go
@@ -104,3 +104,11 @@ func HasAppID() bool {
 	_, err := AppID()
 	return err == nil
 }
+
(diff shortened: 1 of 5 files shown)


internal/env/scm.go
```diff
@@ -104,3 +104,11 @@ func HasAppID() bool {
 	_, err := AppID()
 	return err == nil
 }
+
+func GetAppURL() string {
+	return os.Getenv("QUICKFEED_APP_URL")
```
JosteinLindhom, line 109: We currently only load the root `.env` file:

```go
// Load environment variables from $QUICKFEED/.env.
// Will not override variables already defined in the environment.
const envFile = ".env"
if err := env.Load(env.RootEnv(envFile)); err != nil {
	log.Fatal(err)
}
```

The `QUICKFEED_APP_URL` is written to `/public/.env`, and is not loaded into the environment at any point.

```go
publicEnvFile := env.PublicEnv(m.envFile)
if err := env.Save(publicEnvFile, map[string]string{
	"QUICKFEED_APP_URL": config.GetHTMLURL(),
}); err != nil {
	return err
}
```

I think this will return an empty string.

JosteinLindhom: We likely don't need the `/public/.env` file anymore as we no longer use webpack.
The app URL can be saved in the root `.env` file.

meling (approved):
Looks good to me!

Issue #1269, opened 2025-03-22
(Frontend) DeepSource, resolve performance and bug risks, 6 issues with a total of 125 occurences


JoachimTislov:
- [Avoid using an element's index as the key propJS-0437](https://app.deepsource.com/gh/quickfeed/quickfeed/issue/JS-0437/occurrences?listindex=0) - 6 
- [Unused return value from Array/Object prototype methodJS-D008](https://app.deepsource.com/gh/quickfeed/quickfeed/issue/JS-D008/occurrences?listindex=0) - 1
- [async function should have await expressionJS-0116](https://app.deepsource.com/gh/quickfeed/quickfeed/issue/JS-0116/occurrences?listindex=0) - 9
- [Found unhandled promiseJS-0328](https://app.deepsource.com/gh/quickfeed/quickfeed/issue/JS-0328/occurrences?listindex=0) - 1

- [Avoid .bind() or local functions in JSX propertiesJS-0417](https://app.deepsource.com/gh/quickfeed/quickfeed/issue/JS-0417/occurrences?listindex=0) - 104
- [Non-interactive elements assigned mouse/keyboard event listenersJS-0760](https://app.deepsource.com/gh/quickfeed/quickfeed/issue/JS-0760/occurrences?listindex=0) - 1

PR #1339, opened 2025-04-22
DeepSource bug risk and performance

JoachimTislov:
Closes: #1269
Supercedes: #1274


public/src/Helpers.ts
```diff
@@ -231,6 +231,7 @@ export const setStatusAll = (submission: Submission, status: Submission_Status):
 
 /** getCourseID returns the course ID determined by the current route */
 export const getCourseID = (): bigint => {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
JosteinLindhom, line 234: Instead of ignoring this particular one I think we should convert `getcourseID` to a proper hook.

We could move it to a separate file under `/public/hooks`:


```ts
import { useParams } from "react-router"


export const useCourseID = (): bigint => {
    const route = useParams<{ id?: string }>()
    return route.id ? BigInt(route.id) : BigInt(0)
}
```

public/src/components/navbar/NavBarTeacher.tsx
```diff
@@ -22,7 +22,7 @@ const NavBarTeacher = () => {
         links.unshift({ link: { text: "Review", to: `/course/${state.activeCourse}/review` } })
     }
 
-    const teacherLinks = links.map((link, index) => { return <NavBarLink key={index} link={link.link} icons={link.icons} /> })
+    const teacherLinks = links.map((link) => { return <NavBarLink key={link.link.text} link={link.link} icons={link.icons} /> })
```
meling, line 25: Unrelated to the change here, but it looks to me like the `link:` property of `NavLink` is unnecessary (it is always accessed via `link.link`. If I understand this correctly, perhaps we could open an issue about this and fix it in a separate PR.

meling:
```tsx
export interface NavLink {
    link: { text: string, to: string }
    icons?: ({ text: string | number, classname: string } | null)[],
    jsx?: React.JSX.Element
}
```
Something like this maybe?
```tsx
export interface NavLink {
    linkText: string
    linkTo: string
    icons?: ({ text: string | number, classname: string } | null)[],     // <-- I don't understand the " | null)[]," part
    jsx?: React.JSX.Element
}
```

public/src/pages/TeacherPage.tsx
```diff
@@ -36,7 +36,8 @@ const TeacherPage = () => {
     }
     const results = { title: "View results", text: "View results for all students in the course.", buttonText: "Results", to: `${root}/results` }
     const assignments = { title: "Manage Assignments", text: "View and edit assignments.", buttonText: "Assignments", to: `${root}/assignments` }
-    const updateAssignments = { title: "Update Course Assignments", text: "Fetch assignments from GitHub.", buttonText: "Update Assignments", onclick: () => actions.updateAssignments(courseID) }
+    const handleUpdateAssignments = useCallback(() => actions.updateAssignments(courseID), [actions, courseID])
+    const updateAssignments = { title: "Update Course Assignments", text: "Fetch assignments from GitHub.", buttonText: "Update Assignments", onclick: handleUpdateAssignments }
```
meling, line 40: Suggest to split over multiple lines; one field per line.

meling (changes requested):
Looks mostly good... But there are now some merge conflicts after I merged the `ternary-operations-refactor` PR.

public/src/components/Lab.tsx
```diff
@@ -63,7 +63,7 @@ const Lab = () => {
             let buildLog: React.JSX.Element[] = []
             const buildLogRaw = submission.BuildInfo?.BuildLog
             if (buildLogRaw) {
-                buildLog = buildLogRaw.split("\n").map((x: string, i: number) => <span key={i} >{x}<br /></span>)
+                buildLog = buildLogRaw.split("\n").map((x: string) => <span key={x} >{x}<br /></span>)
```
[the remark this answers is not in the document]
JoachimTislov: The formatter does not want <br/>. It looks odd in this case, but in other cases does make sense to a have space there.

public/src/components/navbar/NavBarTeacher.tsx
JoachimTislov: > icons?: ({ text: string | number, classname: string } | null)[],     // <-- I don't understand the " | null)[]," part

It has to allow null entries because the following variables are conditionally initialized:

```typescript
const pending = state.pendingEnrollments.length > 0 ? { text: state.pendingEnrollments.length.toString(), classname: "badge badge-danger" } : null
const pendingGroups = state.pendingGroups.length > 0 ? { text: state.pendingGroups.length.toString(), classname: "badge badge-danger" } : null
```

JoachimTislov:
For the ones wondering why I have forced pushed three times... 

I did multiple formatting mistakes after merging with master. Im not certain to how it became so clunky formatted...

meling (approved):
Looks good to me!

PR #1345, opened 2025-04-25
Implement tailwindcss into the current Esbuild setup for the UI

JoachimTislov:
This saves the hassle of moving the whole Esbuild implementation to the frontend and prevents us from needing a download script for retrieving the correct binary for tailwind. I am quite certain that this is the most optimal solution for integrating tailwindcss into the current state of QuickFeed.


internal/ui/tailwind/tailwind.go
```diff
@@ -0,0 +1,33 @@
...
+
+func Build() error {
+	return run(build)
+}
+
+func Watch() {
+	err := run(watch...)
+	if err != nil {
+		fmt.Printf("Error running tailwind watcher: %v\n", err)
+	}
```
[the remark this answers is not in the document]
JoachimTislov: That would require a channel as Watch is executed in a go routine, but I think that this type of logging is sufficient..

JoachimTislov:
I have now realized that some of the current classes used for bootstrap is similar to the tailwind classes. This causes unintended styling on multiple elements in the UI:

Dashboard:

image

Course view (teacher):

image

Student lab view:

image

internal/ui/tailwind/tailwind.go
```diff
@@ -0,0 +1,33 @@
+package tailwind
```
meling, line 1: We don't need package for tailwind. We can just keep it in the `ui` package.

meling: The file can still be called `tailwind.go` when moved to the `ui` package. You can rename the functions to:
- `tailwindBuild` or `buildTailwind`
- `tailwindWatch` or `watchTailwind`

meling (changes requested):

Glattshaved:
Classes between bootstrap and tailwind doesn't get mixed up in my implementation somehow.
https://github.com/quickfeed/quickfeed/pull/1352

JoachimTislov:
Issues in https://github.com/quickfeed/quickfeed/pull/1345#issuecomment-2831551591 have been resolved by https://github.com/quickfeed/quickfeed/pull/1345/commits/3d7af83dbfc7537bb9c267a589397dfaa2a2928f. 

The watcher was somehow deleting the output file and was generally not persistent, [view related discussion here](https://github.com/tailwindlabs/tailwindcss/discussions/9031). I replaced the watcher and CLI binary with the main binary. This will persistently rebuild the output file: `tailwind.css`, after the dist folder is cleared. Note that it won't rebuild when changes are made to the actual `tailwind.css` file, since Esbuild only watches the files imported into the application.

Even though most of the issues have been resolved, are there potentially a few more I haven't noticed. For example, the toggle switch for teachers changed for some reason:

Master:
imageimage
This PR:
imageimage

This is the only `known` issue left ... and it's most likely caused by the lack of compatibility between Tailwind 4 and the preprocessor Sass, as mentioned in https://github.com/quickfeed/quickfeed/pull/1345#issuecomment-3376695423

public/src/tailwind.css
```diff
@@ -0,0 +1,5 @@
+@import 'tailwindcss';
+
+@plugin 'daisyui' {
+  themes: light --default, dark --prefersdark;
+}
```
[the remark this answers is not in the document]
JoachimTislov: False positive; I believe your information is outdated.

internal/ui/esbuild.go
```diff
@@ -57,14 +58,14 @@ var buildOptions = api.BuildOptions{
...
 				setup.OnStart(func() (api.OnStartResult, error) {
 					if err := resetDistFolder(); err != nil {
 						return api.OnStartResult{
 							Warnings: []api.Message{
 								{
-									PluginName: "Reset",
+									PluginName: "Reset dist",
 									Text:       "Failed to clear the dist folder",
 									Notes: []api.Note{
 										{Text: fmt.Sprintf("The dist directory may now contain multiple builds\nLocation: %s", distDir)},
```
meling, line 71: `dist directory` -> `dist folder`

internal/ui/esbuild.go
```diff
@@ -57,14 +58,14 @@ var buildOptions = api.BuildOptions{
...
-			Name: "Reset Plugin",
+			Name: "Reset dist & Generate Tailwind CSS",
 			Setup: func(setup api.PluginBuild) {
 				setup.OnStart(func() (api.OnStartResult, error) {
 					if err := resetDistFolder(); err != nil {
 						return api.OnStartResult{
 							Warnings: []api.Message{
 								{
-									PluginName: "Reset",
+									PluginName: "Reset dist",
```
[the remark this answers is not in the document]
meling: btw: shouldn't this match the name of the plugin named above?

internal/ui/esbuild.go
```diff
@@ -57,14 +58,14 @@ var buildOptions = api.BuildOptions{
 	},
 	Plugins: []api.Plugin{
 		{
-			Name: "Reset Plugin",
+			Name: "Reset dist & Generate Tailwind CSS",
```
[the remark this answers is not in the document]
meling: On the other hand, would it be possible to separate this into two plugin entries? One for resetting the dist folder and one for generating Tailwind? Both would have their own name and setup function. It would look nicer.

internal/ui/esbuild.go
```diff
@@ -57,14 +58,14 @@ var buildOptions = api.BuildOptions{
...
 			Setup: func(setup api.PluginBuild) {
 				setup.OnStart(func() (api.OnStartResult, error) {
```
meling, line 63: The `buildOptions` variable definition is too long (~100 lines).

Consider creating helper setup functions and a separate variable for the entry points.

meling (changes requested):
@JoachimTislov Thanks for this refresh. Looks good to me, but see my nits in `esbuild.go`.

I defer to @JosteinLindhom for hands-on input and a second opinion on the tsx and css stuff. Could you also fire this up in the browser so we can look at it tomorrow or later this week?

internal/ui/esbuild.go
JoachimTislov: I decided to remove the note with this statement, as it was redundant. The presence of old files is already implied when a folder isn’t reset.

meling (approved):
Thanks again. Nice organization of the plugins. Still waiting for @JosteinLindhom

JoachimTislov:
I'm aware of the issue with applying h-full without removing the `height: 100vh` inside of style.scss, which appears to have a higher priority. Removing it resolves the conflict.

meling:
Superseded by #1505

PR #1352, opened 2025-04-27
Implements Tailwind using a plugin

Glattshaved:
This PR implements Tailwind using the esbuild-plugin-tailwind and also daisyUI a plugin for Tailwind to deliver complete UI elements for faster development


Glattshaved:
Still need to setup some splitting and/or minify for the assets/index.js file that is generated.

PR #1367, opened 2025-05-05
Use mocked scm client in development and add dummy data generator

### LINKED ISSUE #1341: Use the mocked scm client for development and generate related dummy data
QuickFeed is highly integrated with GitHub and this has caused many restrictions for manually testing and interacting with QuickFeed in development. Any operation which requires synchronization with GitHub is not possible to perform unless the webhook is configured. Heres a list of endpoints which use the SCM client: 

These are not functional without a working webhook (web/quickfeed_service.go)

- UpdateAssignments
- IsEmptyRepo
- UpdateCourse
- UpdateEnrollments
- UpdateGroup
- DeleteGroup
- UpdateAssignments

The endpoints which purely updates GitHub data is not the focus. Its the certain endpoints which rely on GitHub to perform the requested operation that requires a "fake" scm client to mock the response.

The amount of work required is dependent on the existing mock of GitHub, and I assume it would be natural to generate mock data which can be used along with the existing mocked/dummy organizations.

This will close #40 and render the anonymizer script obsolete, I think.

JoachimTislov:
Closes: #1341


meling:
Just for the record. I support developing this. It is not my own priority, but I appreciate your efforts on this @JoachimTislov. One small comment about naming; I suggest calling the internal package `mockdata` and the command `cmd/mockgen` or `cmd/mockdatagen`. The only issue with `mockgen` is that there's another tool with the same name. But that shouldn't be a problem if we only run it from the `Makefile` using `go run`. The make target could be `make mockdata`, which reads nicely.

meling:
Closing in favor of the new initiative outlined in #1638.

Issue #1368, opened 2025-05-05
Members page. Logic of promoting and demoting was reversed in ternary refactor - pr


JoachimTislov:
I made a mistake in #1325 

The role action was reversed;

![Image](https://github.com/user-attachments/assets/810ef09c-ce16-488f-a7e8-b3a29d77de5b)

The green line 48 and red line 78 should have matching order.

So the naming - `currentRole` makes sense, but in order to update the enrollment by promoting the user, will this will fail. Because it attempts to update the enrollment to the same status...

I encountered this bug after configuring #1367 and it allowed me to perform the action without receiving the scm client error. So im both happy and salty/mad, but I've atleast found this out before it entered production. That would break a very important functionality of QuickFeed..

PR #1369, opened 2025-05-05
Fix previous slip up in members.tsx

JoachimTislov:
Closes: #1368

I debated on creating a test for the members component. To avoid regression, but it appears to be complex and fragile. Any further changes made to member.tsx will require updates in the test.


JosteinLindhom:
I made some changes to get rid of many of the ternary operators and use a switch instead to determine text, color and status.
While slightly more verbose, it's now more clear what is going on.

PR #1396, opened 2025-08-12
Environment variables are now properly loaded before building frontend

### LINKED ISSUE #1395: building frontend with cmd esbuild does not include (load) environment variables
When building the frontend we need to pass an environment variable `QUICKFEED_APP_URL` in order for the correct URL to show in our course creation component.

Using our `cmd/esbuild/main.go` in our makefile (or directly) will *not* include any environment variables.
We should make sure to load the environment variables in `cmd/esbuild/main.go` to make environment variables available during build.

JosteinLindhom:
- **feat(state): add settings namespace to Overmind configuration**
- **feat(ui): set up route and button for a settings page**
- **feat(ui): add settings components for changing UI elements**
- **feat(ui): add theme provider that update CSS variables when state changes**
- **fix: add missing settings route**
- **feat(ui): add settings to select table cell colors**
- **fix(ui): increase default width of test/criterion bar indicator**
- **fix(ui): restrict settings for results/review table to teachers only**
- **fix(ui): refactor color pickers into a single generic color picker component**
- **fix(ui): simplified how settings are applied**
- **fix(ui): add rule for fixed-width prepends**
- **fix(ui): replace color pickers with generic ColorPicker component**
- **fix(ui): change state keys to match CSS variable style**
- **fix(ui): change placement of restore default settings button**
- **fix: use reaction for updating range element**
- **fix: export overmind reaction hook**
- **style: update color variables hover highlight in submission view**
- **style: adjust box-shadow and padding for not-graded class**
- **style: refactor table and status styles**
- **fix: update minimum range value for bar width setting**
- **fix: reset bar-width to default width  if out of bounds**
- **style: remove unnecessary semicolons for consistency**
- **fix: handle null newSettings in updateSettings and update loadSettings return type**
- **style: remove redundant table cell styles**
- **fix: resolve various deepsource issues (self-closing tags, template literals, wildcard import)**
- **building frontend with cmd esbuild does not include (load) environment variables Fixes #1395**


JosteinLindhom:
This included about 450 more lines than intended.. Closing..

Issue #1430, opened 2025-10-10
Bootstrap 404 source map warning in webconsole


JoachimTislov:
<img width="1276" height="77" alt="Image" src="https://github.com/user-attachments/assets/d4bd7849-26f7-4d51-8683-8c7e758022e6" />

When I previously added `bootstrap.min.css`, I forgot to add the source map. This can be helpful now that Tailwind is entering development.

PR #1431, opened 2025-10-10
chore(Bootstrap): Add source map

JoachimTislov:
Closes: #1430


JoachimTislov:
Yes, it makes sense to keep it. Bootstrap will probably exist for a while after the Tailwind PR has been merged. In terms of actual usage, it's not really needed, but it prevents the 404 warning. Optionally, the source map link comment in bootstrap.min.css could be removed.

With source map:
image

Without:
image

PR #1478, opened 2026-01-13
fix alerts missing on some pages

JosteinLindhom:
- **feat(alerts): add toast-style alerts container with slide-in animation**
- **refactor(alerts): update alerts styling**
- **fix(alerts): add Alerts component to App and remove from other pages**


JosteinLindhom:
Alerts are also made sticky in this PR, showing them regardless of scroll position:

alerts

PR #1491, opened 2026-01-20
Fix CA certificate installation for local development

### LINKED ISSUE #1443: Add support for receiving Github webhooks locally
We can use https://docs.github.com/en/webhooks/testing-and-troubleshooting-webhooks/using-the-github-cli-to-forward-webhooks-for-testing to receive webhook events locally.

However after some initial testing it seems that we cannot forward events to our server due to us using TLS with self-signed certificates in dev mode.
We should consider adding a separate server that does not use TLS to handle these webhook events.

Errors received when trying to run as-is:
```bash
$ gh webhook forward --events=push --org=qf105 --url "https://127.0.0.1/hook"
Forwarding Webhook events from GitHub...
warning: error forwarding event: Post "https://127.0.0.1/hook": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

JosteinLindhom:
- **fix: extract CA cert from fullchain for trust store**
- **feat: add cmd for self-signed certificate generation and trust store update**
- **docs: update deployment instructions for GitHub webhooks with GitHub CLI and ngrok**


doc/deploy.md
```diff
@@ -241,17 +242,79 @@ Note that you will need to repeat this step each time you recompile the server.
...
+#### Option 1: Using GitHub CLI (Recommended for Development)
 
-1. Start ngrok: `ngrok http 443` - assuming the server runs on port `:443`.
-2. ngrok will generate a new endpoint URL.
-   Copy the urls an update webhook callback information in your GitHub app to point to this URL.
+The [GitHub CLI](https://cli.github.com/) provides a built-in webhook forwarding feature that is ideal for local development.
+
+**Prerequisites:**
+
+1. Install GitHub CLI and authenticate:
```
meling: I would prefer to write these commands as a makefile target.

internal/cert/add_cert_linux.go
```diff
@@ -3,21 +3,45 @@
...
-	out, err := sh.OutputA("sudo", "cp", certFile, certPath)
+	const caCertFile = "quickfeed-ca.crt"
+
+	// Extract only the CA certificate from the fullchain
+	caCertPEM, err := extractCACert(certFile)
+	if err != nil {
+		return err
+	}
+
+	// Write the CA certificate to a temporary file first
```
meling: Much of this writing to temp looks to be shared across platforms? Maybe it could be moved to a `util.go` file or something?

meling: Or in the `extract.go` file.

meling (changes requested):

internal/cert/add_cert_darwin.go
```diff
@@ -12,12 +14,60 @@ const keychain = "/Library/Keychains/System.keychain"
...
+
+// isTrustedCertInstalled reports whether the certificate with the given SHA-1
+// thumbprint is present in the system keychain. Given -Z, security(1) prints a
+// "SHA-1 hash:" line for each certificate, so the certificate is matched by
+// thumbprint rather than by subject name, which is not unique.
+func isTrustedCertInstalled(thumbprint string) bool {
+	// find-certificate exits non-zero when the keychain holds no certificates,
+	// which is not an error here; an empty listing simply matches nothing.
+	out, _ := sh.OutputA("security", "find-certificate", "-a", "-Z", keychain)
+	return strings.Contains(strings.ToUpper(out), thumbprint)
```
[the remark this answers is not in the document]
JosteinLindhom: I'm not able to test any fix for this currently, so leaving this for now.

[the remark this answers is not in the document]
meling: Not sure if this is important. Feels a bit nitpicky. We can fix it later if we find it is a problem.

PR #1518, opened 2026-04-10
feat: Implement slip days for groups

### LINKED ISSUE #1555: Implement slipdays for groups
This issue tracks the implementation of group slip days support in QuickFeed.

Group submissions should deduct slip days from a shared group slip-day pool, similar to how individual enrollments track slip days.

## Changes Required

- Extend `Group` and `UsedSlipDays` protobuf messages with group slip-day tracking fields
- Implement group slip-day calculation and update logic
- Integrate group slip days into CI `RecordResults` for group submissions
- Preload group `UsedSlipDays` and populate remaining slip days
- Surface group slip day values in the UI

## Design Considerations

- Refactor shared slip-day logic to avoid duplication between `Enrollment` and `Group` types
- Consider extracting slip-day logic into a shared helper that both `Group` and `Enrollment` can use
- Ensure `SlipDaysRemaining` (computed field) is populated when returning groups from the database

Closes #1518

JosteinLindhom:
- **feat: add usedSlipDays and slipDaysRemaining fields to Group message**
- **feat: preload UsedSlipDays for Groups**
- **feat: set remaining slip days for groups in PopulateSlipDays**
- **feat: implement group slip days update logic in RecordResults**
- **test: add unit test for fetching groups with slip days**
- **feat: implement UpdateSlipDays and related methods for group slip day management**

Fixes #1555


qf/group_slipdays.go
```diff
@@ -0,0 +1,72 @@
+package qf
+
+import "fmt"
+
+// UpdateSlipDays updates the number of slip days for the given assignment/submission.
+// This method is for group submissions.
+func (m *Group) UpdateSlipDays(assignment *Assignment, submission *Submission) error {
```
meling, line 5: Most of the logic in this file appears to overlap with `qf/enrollment.go`'s implementation of slip days. I suggest extracting the slip days handling into a separate type that can be constructed based on either a `Group` or `Enrollment`. Alternatively, switch to functions instead of methods on a type. I don't have a specific design in mind right now, but I think it is better to keep the "slip days" logic in one place rather than duplicating it in two places, which risks them diverging.

meling: Keeping the slip days logic separate from its two instances would also reduce the number of tests we have to maintain.

qf/group_slipdays_test.go
```diff
@@ -0,0 +1,246 @@
...
+		[][]int32{
+			{5, 5, 5, 5, 5},
+			{5, 5, 5, 5, 5},
+			{5, 5, 4, 3, 2},
+			{2, 2, 2, 2, 1, 0, -1},
+		},
+	},
+}
+
+func TestGroupSlipDays(t *testing.T) {
```
meling: Consider if these tests overlap with the existing `qf/slipdays_test.go`...

qf/group_slipdays.go
meling: PS: If it makes it easier, we can also keep the methods on both `Group` and `Enrollment`, and just call out to the underlying (shared) logic functions. But we should also consider if there are better structures for this shared logic.

public/src/components/submissions/SubmissionInfo.tsx
```diff
@@ -16,6 +16,12 @@ const SubmissionInfo = ({ submission, assignment }: SubmissionInfoProps) => {
     const built = getFormattedTime(buildInfo?.BuildDate)
     const executionTime = buildInfo ? `${buildInfo.ExecTime / BigInt(1000)} seconds` : ""
 
+    const isGroupSubmission = submission.groupID > 0n
+    const group = isGroupSubmission
+        ? (state.groups[assignment.CourseID.toString()]?.find(g => g.ID === submission.groupID)
```
meling, line 20: This looks like logic we might have in several places in the codebase, which could be pulled into a new method on `state` called `FindGroup(courseID, groupID)`.

ci/record_results.go
```diff
@@ -99,30 +99,36 @@ func (r *RunData) newTestRunSubmission(previous *qf.Submission, results *score.R
 	}
 }
 
-func (r *RunData) updateSlipDays(db database.Database, submission *qf.Submission) error {
-	enrollments := make([]*qf.Enrollment, 0)
-
+func (r *RunData) updateSlipDays(logger *zap.SugaredLogger, db database.Database, submission *qf.Submission) error {
```
meling: Looks like the two code paths in this method share similar logic:
```go
if err := group.UpdateSlipDays(r.Assignment, submission); err != nil {
	return fmt.Errorf("failed to update slip days for group %d in course %d: %w", group.GetID(), r.Assignment.GetCourseID(), err)
}
if err := db.UpdateSlipDays(group.GetUsedSlipDays()); err != nil {
	return fmt.Errorf("failed to update slip days for group %d (course %d): %w", group.GetID(), group.GetCourseID(), err)
}
```
And:
```go
if err := enroll.UpdateSlipDays(r.Assignment, submission); err != nil {
	return fmt.Errorf("failed to update slip days for user %d in course %d: %w", enroll.GetUserID(), r.Assignment.GetCourseID(), err)
}
if err := db.UpdateSlipDays(enroll.GetUsedSlipDays()); err != nil {
	return fmt.Errorf("failed to update slip days for enrollment %d (user %d) (course %d): %w", enroll.GetID(), enroll.GetUserID(), enroll.GetCourseID(), err)
}
```
We should consider whether these code paths can be combined into a single path, given potential refactoring in the `qf` package.

meling (changes requested):
I suggest that we consider making `SlipDays` a separate type with methods that can fit both the `Enrollment` and `Group` uses. Possibly with wrappers on the two usage types. I have not thought carefully about the best design for this type, but I have a strong hunch that it should be relatively easy to come up with a reasonable refactored design.

meling:
@copilot create an issue for this PR and link this PR to it.

qf/slipdays.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// gracePeriod is the grace period for submissions after the deadline.
+// The grace period should be hours in the range 0-23.
+// Note grace period applies to all enrollments and groups (courses).
+const gracePeriod = 2 * time.Hour
+
+// slipDayHolder is implemented by *Group and *Enrollment, allowing the slip-day
+// bookkeeping below to be shared between the two instead of duplicated.
+type slipDayHolder interface {
```
meling, line 15: I guess this is not strictly needed anymore, but it would require that the two interface assertion lines I added earlier be removed. That's fine with me. We would notice a mistake if two implementations deviate, since we have consumer-side interfaces that would yell at you if their contract is violated.

qf/enrollment.go
```diff
@@ -1,81 +1,32 @@
...
-	"time"
-)
-
-// gracePeriod is the grace period for submissions after the deadline.
-// The grace period should be hours in the range 0-23.
-// Note grace period applies to all enrollments (courses).
-const gracePeriod time.Duration = time.Duration(2 * time.Hour)
-
 // UpdateSlipDays updates the number of slip days for the given assignment/submission.
 func (m *Enrollment) UpdateSlipDays(assignment *Assignment, submission *Submission) error {
```
meling: We could consider moving these `Enrollment` forwarding methods to a new file `enrollment_slipdays.go` to match the `group_slipdays.go` ...

PR #1539, opened 2026-05-21
chore: fix TypeScript and ESLint warnings across frontend

JosteinLindhom:
- **fix: update module resolution to 'bundler' and set module to 'ESNext' in tsconfig**
- **chore: fix TypeScript and ESLint warnings across frontend**
- **refactor: removed fragments and return elements directly**
- **feat: add global TypeScript declarations for SCSS modules and process.env**
- **fix: update error handler signatures to require payload and handle ConnectError**
- **fix: update BigInt prototype toJSON method type assertion**
- **fix: simplify error handling in useCourseID hook**
- **refactor: change imports to type imports in state and errors modules**


meling (approved):
Nice!!

PR #1546, opened 2026-06-11
fix: detach some handlers from connection contexts

JosteinLindhom:
Fixes #1498


web/interceptor/detach_test.go
```diff
@@ -0,0 +1,102 @@
...
+// observe reports true if ctx is canceled within the grace period after the
+// client disconnects, and false if the handler outlives the disconnect.
+func (c *cancellationProbe) observe(ctx context.Context) {
+	close(c.started)
+	select {
+	case <-ctx.Done():
+		c.canceled <- true
+	case <-time.After(2 * time.Second):
+		c.canceled <- false
+	}
```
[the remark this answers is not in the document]
JosteinLindhom: After Go 1.23 the docs for `time.After` state this:

`
Before Go 1.23, this documentation warned that the underlying Timer would not be recovered by the garbage collector until the timer fired, and that if efficiency was a concern, code should use NewTimer instead and call Timer.Stop if the timer is no longer needed.
 As of Go 1.23, the garbage collector can recover unreferenced, unstopped timers. There is no reason to prefer NewTimer when After will do.
`

web/interceptor/detach.go
```diff
@@ -0,0 +1,60 @@
...
+const detachTimeout = 2 * time.Minute
+
+// detachedMethods lists methods that perform multi-step SCM and database
+// mutations without rollback support. Aborting such a method midway, e.g.,
+// because the client closed the connection, can leave the GitHub organization
+// and the database in inconsistent states.
+// Once started, these methods should be allowed to complete, bounded by detachTimeout.
+var detachedMethods = map[string]bool{
+	"UpdateGroup":       true, // may create the group repo and add/remove collaborators on GitHub
+	"DeleteGroup":       true, // deletes database records before deleting the GitHub repo
```
meling: I think these two delete operations are reversed by the other PR, which should be ready to merge. Suggest updating this comment to reflect future behavior.

web/interceptor/detach.go
```diff
@@ -0,0 +1,60 @@
...
+// cancellation signal from the client connection. It must be generous enough
+// to allow several sequential GitHub API calls, e.g. one per group member.
+const detachTimeout = 2 * time.Minute
+
+// detachedMethods lists methods that perform multi-step SCM and database
+// mutations without rollback support. Aborting such a method midway, e.g.,
+// because the client closed the connection, can leave the GitHub organization
+// and the database in inconsistent states.
+// Once started, these methods should be allowed to complete, bounded by detachTimeout.
+var detachedMethods = map[string]bool{
```
meling, line 21: Should this map also hold `RejectEnrollment` or other methods?

meling (approved):
Looks good to me!

meling:
I think this looks good. Any reason not to remove the Draft status and merge?

web/interceptor/detach.go
JosteinLindhom: We don't have a `RejectEnrollment` RPC. Rejection is handled by passing an enrollemnt with status = `NONE` in `UpdateEnrollments`.
I will update the comment below

PR #1548, opened 2026-06-15
fix: delete SCM repos before database repos when removing groups, enrollments

JosteinLindhom:
- **refactor: improve error handling and repository deletion logic in SCM operations**
- **test(groups): add TestDeleteGroupRepoAlreadyDeleted to verify deletion logic for non-existent repositories**


scm/github.go
meling: Nit: by ID.

meling (approved):
Looks good to me!

PR #1552, opened 2026-07-02
feat: add support for adding internal notes to students

### LINKED ISSUE #1547: TAs should be able to add internal notes visible only to course staff
In some cases a submission reaches the required score threshold, but a TA identifies issues that must be addressed before the assignment can be approved. The TA may therefore ask the student to make specific improvements and return for a new review.

When the student later approaches a different TA, there is currently no way for that TA to see the previous review comments or approval conditions. As a result, important review decisions may be lost, leading to inconsistent grading and approval decisions.

One example scenario:

- A student submits an assignment and gets gets a score greater than or equal to the score treshold.
- TA 1 reviews the submission and determines that the student must fix issues A and B before receiving approval.
- The student fixes issue A but not issue B.
- The student later requests approval from TA 2.
- TA 2 has no visibility into TA 1's previous review and may approve the submission even though issue B remains unresolved.

We should allow TAs to write internal notes for a submission / enrollment / group.

Notes could exist at multiple levels:

Submission notes
Group notes
Enrollment notes

All notes should:

Be hidden from students.
Record the author and timestamp.
Be visible to all TAs.
Remain available for future reviews.

When reviewing a submission, TAs should be able to see all relevant notes from the submission itself as well as any associated group and enrollment notes.

JosteinLindhom:
Implements internal notes for course staff, addressing #1547.

TAs and teachers can attach a note to a submission, a group, or an
enrollment. Notes are never shown to students, record their author and
timestamp, and are visible to all teaching staff so a later reviewer can
see what a previous reviewer flagged. Notes appear in the review form
(`ReviewForm`/`LabResult`) alongside the relevant submission, and the
Members view shows a per-student note count with a new student-details
page summarizing a student's notes and submission history.

## Deviations from #1547

- **Editing is author-only, not staff-wide.** The issue only requires
  notes to be *visible* to all TAs, which this PR does. Editing/deleting
  was initially also opened up to site admins, but nothing asked for it
  and the access-control interceptor already gates the note RPCs to
  course teachers, so that branch only ever applied to an admin who was
  also a teacher on the course. It was removed; only a note's author can
  edit or delete it.
- **Members overview page.** The per-student note count and the new
  `StudentDetails` page aren't requested by the issue but were added to
  make notes discoverable from the Members view rather than only from a
  submission's review.

Also documents QuickFeed's request/response message type policy in
[AGENTS.md](AGENTS.md), generalizing the pattern this feature follows
(passing `qf.Note` directly instead of a wrapper request type).


qf/types.proto
```diff
@@ -307,3 +308,24 @@ message FeedbackReceipt {
...
+
+// Note is an internal note written by course staff (teachers/TAs).
+// Notes are never exposed to students; access is restricted to teachers via the access control interceptor.
+// Exactly one of SubmissionID, GroupID, or EnrollmentID identifies the target the note is attached to.
+message Note {
+    uint64 ID                           = 1;
+    uint64 CourseID                     = 2;  // foreign key; required for access control
+    uint64 AuthorID                     = 3;  // UserID of the teacher/TA who wrote the note
+    string body                         = 4;
+    uint64 SubmissionID                 = 5;  // set if the note is attached to a submission
```
meling, line 325: Should we use `oneof` here? If not, perhaps we should add a comment to explain why it is annoying to use here?

qf/types.proto
```diff
@@ -307,3 +308,24 @@ message FeedbackReceipt {
 message AssignmentFeedbacks {
     repeated AssignmentFeedback feedbacks = 1;
 }
+
+//   INTERNAL NOTES   //
+
+// Note is an internal note written by course staff (teachers/TAs).
```
meling: `course staff (teachers/TAs).` --> `teaching staff.`

web/notes.go
```diff
@@ -0,0 +1,177 @@
...
+func checkNoteBody(body string) error {
+	if strings.TrimSpace(body) == "" {
+		return connect.NewError(connect.CodeInvalidArgument, errors.New("note body must not be empty"))
+	}
+	return nil
+}
+
+// hasSingleTarget returns true if the note references exactly one of a
+// submission, group, or enrollment.
+func hasSingleTarget(note *qf.Note) bool {
```
meling: This could be covered by the `NoteRequest.IsValid()` method, like it is for the `NoteRequests.IsValid()` method. Any reason to keep this check in the handler and not in the `IsValid()` method?

The same applies to the `checkNoteBody` above.

web/notes.go
```diff
@@ -0,0 +1,177 @@
...
+		if errors.Is(err, database.ErrEmptyNoteID) {
+			return nil, connect.NewError(connect.CodeInvalidArgument, errors.New("note ID is required"))
+		}
+		return nil, connect.NewError(connect.CodeNotFound, errors.New("note not found"))
+	}
+	if existing.GetCourseID() != in.GetCourseID() {
+		return nil, connect.NewError(connect.CodePermissionDenied, errors.New("note does not belong to the course"))
+	}
+
+	// TODO(jostein): Currently the interceptor only checks that the caller is a teacher in the course, so any call that ends up here is from a teacher in the course.
```
meling: I think allowing admins to edit other teachers' notes is a YAGNI feature. It seems unlikely that we will need it any time soon.

web/notes.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	courseID := in.GetCourseID()
+	// The interceptor only verifies the caller teaches courseID, not that the
+	// note's target lives in that course; reject cross-course targets here.
+	if !s.noteTargetInCourse(courseID, in.GetNote()) {
+		return nil, connect.NewError(connect.CodePermissionDenied, errors.New("note target does not belong to the course"))
+	}
+	// Build the note from only the fields a client may set; ID, author, course,
+	// and timestamps are server-owned and must not be taken from the request.
+	note := &qf.Note{
```
meling, line 34: This could be replaced by:
```go
note := in.GetNote()
note.CourseID = courseID
note.AuthorID = userID(ctx)
```
But see the other comment above.

web/notes.go
```diff
@@ -0,0 +1,177 @@
...
+	"strings"
+
+	"connectrpc.com/connect"
+	"github.com/quickfeed/quickfeed/database"
+	"github.com/quickfeed/quickfeed/qf"
+)
+
+// CreateNote creates a new internal staff note attached to a submission, group, or enrollment.
+// The author and timestamps are set server-side; the access control interceptor restricts this to teachers.
+func (s *QuickFeedService) CreateNote(ctx context.Context, in *qf.NoteRequest) (*qf.Note, error) {
```
meling: This is sort of a new "policy" being introduced here. We would normally pass just the `qf.Note` as the argument here since it already holds the `CourseID`. Is there any reason to start deviating from the original quickfeed style? I know we are flying against so-called established best practices, but I'm not convinced that we need to deviate from our own best practices.

I see two paths forward:
1. Either we stick to our guns and define the policy clearly in the `AGENTS.md` file, so that coding agents know about it and stick to it. 
2. Or we switch to "duplicated" types: message types (for `TypeNameRequest` and `TypeNameResponse`) and storage types (`TypeName`) across the board. This would be a separate PR.

Unless there are strong arguments against option 1, I'm in favor of sticking with it.

meling (changes requested):
I've done a preliminary review of the Go code; it mostly looks fine, but I'm not convinced we need both the message types and the storage types (NoteRequest, NoteResponse, Note, etc.). I don't mind these if there are compelling reasons for duplicate message types.

PR #1556, opened 2026-07-08
Fix frontend TypeScript errors

JosteinLindhom:
- **fix: update recharts version and improve tooltip formatting in FeedbackGraph component**
- **fix: update global.d.ts to declare ProcessEnv interface for environment variables**
- **fix: simplify mock function signature in TestHelpers.ts**
- **refactor: remove unnecessary React imports from components and tests**
- **fix: handle optional user parameter in updateGroupUsers function**
- **refactor: define SearchProps type for improved type safety in Search component**
- **fix: ensure owner is defined before updating submission owner in Results component**


JosteinLindhom:
`tsc` reported ~100 errors when compiling the frontend. 
The vast majority of these were unused imports, specifically `React`, which is no longer required.

Fixed a few instances where some functions were used with invalid inputs (e.g. T | undefined, where only T was accepted).
Fixed some instances where function signatures were incorrect as well.

Upgraded `recharts` to latest version due to an error in the library. 
(This could be ignored by setting `skipLibCheck = true` in the compiler options. That way `tsc` would ignore errors from third party libraries we do not contorl.)

PR #1566, opened 2026-08-08
logging: migrate backend to structured, context-aware slog

### LINKED ISSUE #489: Add a logging interceptor

### LINKED ISSUE #670: Custom logger with context

meling:
## Summary

Replaces zap and `SugaredLogger` across the backend with `log/slog`. Loggers now travel in the request context instead of being passed alongside it, so request scope survives calls into SCM, CI, and background work.

Fixes #670
Fixes #489

## What's new

**`internal/qlog`** — logger construction and context plumbing:

- `New(w)` builds a text handler at debug level with `AddSource`, rewriting source paths to be repository-relative so they stay readable in test output.
- `FromContext(ctx)` / `NewContext(ctx, logger)` carry the logger.
- `WithLogger(ctx, attrs...)` returns both an enriched context and its logger, for functions that log repeatedly and pass the context downstream.
- `With` is the context-only variant.
- `FromContext` falls back to the process logger when none is attached, so helpers work without ceremony in tests.

**`internal/qlog/label`** — canonical attribute keys (`course_id`, `rpc_method`, `target_user_id`, …) for fields that are queried or shared across packages. Component-specific names stay local to the component.

**`web/interceptor/logging.go`** — RPC completion logging: procedure, Connect code, and duration, with no request or response bodies. The interceptors attach `rpc_method`, plus `user_id` and `course_id` once authentication and access control have accepted the request, so handlers never repeat them.

## Scoping

Attributes are recorded once per record, and slog keeps duplicate keys — a repeated attribute shows up twice in the same line. Scopes are therefore derived once with `qlog.WithLogger` where a function logs the same attributes more than once or hands them to functions that log them again. Handlers leave the request scope to the interceptors, and use `target_user` / `target_user_id` when acting on someone other than the caller. `web.TestRPCLoggingNoDuplicateScope` guards this.

## Sensitive data

- Stop logging RPC bodies, refresh tokens, and session secrets.
- Redact generated per-run secrets from captured command output before logging.
- Keep the stack trace when recovering a panic while recording results — the run attributes alone are not enough to locate the cause.

## Also

- Drops the `go.uber.org/zap` and `go.uber.org/multierr` dependencies.
- GORM logger adapter in `database/logger.go`; test adapter in `internal/qtest/logger_helper.go`.
- Constructors that need a logger for the lifetime of the object they return take one directly, since a context cannot be stored — the SCM client constructors and the mocked SCM client, whose handlers capture the logger.
- `doc/dev.md` documents the structured-logging and sensitive-data rules.

---


assignments/assignments.go
```diff
@@ -26,12 +27,21 @@ const MaxWait = 5 * time.Minute
...
 
-	logger.Debugf("Updating %s from '%s' repository", course.GetCode(), qf.TestsRepo)
-	ctx, cancel := context.WithTimeout(context.Background(), MaxWait)
+	// Scope every record below to the tests repository, so that the individual
+	// statements do not repeat these attributes. The caller is expected to have
+	// scoped the context to the course already.
+	ctx, logger := qlog.WithLogger(
+		ctx,
+		label.Repository, qf.TestsRepo,
+		label.RepositoryType, qf.Repository_TESTS.String(),
```
meling: This one seems redundant. Why do we need this one?

assignments/assignments.go
```diff
@@ -40,57 +50,58 @@ func UpdateFromTestsRepo(logger *zap.SugaredLogger, runner ci.Runner, db databas
...
 		for _, assignment := range assignments {
-			logger.Debugf("Failed to update database for: %v", assignment)
+			logger.Debug("assignment not updated in database", label.Assignment, assignment.GetName())
 		}
-		logger.Errorf("Failed to update assignments in database: %v", err)
+		logger.Error("failed to update assignments in database", label.Error, err)
 		return
 	}
-	logger.Debugf("Assignments for %s successfully updated from '%s' repo", course.GetCode(), qf.TestsRepo)
+	logger.Debug("updated from tests repository")
```
meling: This should still say `assignments successfully updated from ...`

ci/run_tests.go
```diff
@@ -94,24 +102,35 @@ func (r *RunData) RunTests(ctx context.Context, logger *zap.SugaredLogger, sc sc
...
 		testsFailedExtractResultsCounter.WithLabelValues(r.JobOwner, r.Course.GetCode()).Inc()
-		logger.Debugf("Session secret: %s", randomSecret)
-		logger.Errorf("Failed to extract (some) results for assignment %s for course %s: %v", r.Assignment.GetName(), r.Course.GetName(), err)
+		logger.Error("failed to extract some test results", label.Error, err)
 		// don't return here; we still want partial results!
 	}
 
 	testsSucceededCounter.WithLabelValues(r.JobOwner, r.Course.GetCode()).Inc()
-	logger.Debug("ci.RunTests", zap.Any("Results", qlog.IndentJson(results)))
+	logger.Debug("test results extracted", "score", results.Sum(), "tests", len(results.Scores))
```
meling, line 118: Should we still log the results json? (without `IndentJson()`) We have used this for debugging in the past, but it generates too much noise for the common case. Perhaps this should be added as a separate logger destination.

internal/qlog/logger.go
```diff
@@ -1,14 +1,76 @@
...
-	return cfg.Build(zap.AddStacktrace(zapcore.PanicLevel))
+type contextKey struct{}
+
+// New returns a structured logger that writes debug and higher-level records to w.
+func New(w io.Writer) *slog.Logger {
+	root := env.Root()
+	handler := slog.NewTextHandler(w, &slog.HandlerOptions{
+		AddSource: true,
+		Level:     slog.LevelDebug,
+		ReplaceAttr: func(_ []string, attr slog.Attr) slog.Attr {
```
meling, line 21: Check this logic later

internal/qtest/logger_helper.go
```diff
@@ -1,25 +1,27 @@
...
-	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
+	"github.com/quickfeed/quickfeed/internal/qlog"
 )
 
-func Logger(t *testing.T) *zap.SugaredLogger {
+// Context returns a background context containing the test logger.
+func Context(t *testing.T) context.Context {
+	t.Helper()
+	return qlog.NewContext(context.Background(), Logger(t))
```
meling: This should use t.Context().

web/users.go
```diff
@@ -3,14 +3,16 @@ package web
...
 	"github.com/quickfeed/quickfeed/qf"
 	"github.com/quickfeed/quickfeed/scm"
 )
 
 // editUserProfile updates the user profile according to the user data in
 // the request object. If curUser is admin, and the request may also
 // promote the user to admin.
-func (s *QuickFeedService) editUserProfile(curUser, request *qf.User) error {
+func (s *QuickFeedService) editUserProfile(ctx context.Context, curUser, request *qf.User) error {
 	updateUser, err := s.db.GetUser(request.GetID())
```
meling: Suggest `updateUser` -> `targetUser` is a better name for this.

web/hooks/github.go
```diff
@@ -1,24 +1,39 @@
...
 	"github.com/quickfeed/quickfeed/web/stream"
-	"go.uber.org/zap"
 )
 
 // maxConcurrentTestRuns is the maximum number of concurrent test runs.
 const maxConcurrentTestRuns = 5
 
+// webhookTimeout bounds the handling of a single push event, which clones the
+// repository and runs the assignment's tests in a container.
+const webhookTimeout = 30 * time.Minute
```
meling, line 31: This seems excessive. GitHub will fail a webhook push event much more rapidly than 30 minutes.

web/rebuild.go
```diff
@@ -73,19 +90,24 @@ func (s *QuickFeedService) internalRebuildSubmission(request *qf.RebuildRequest)
...
 
 	// counting semaphore: limit concurrent rebuilding to maxContainers
 	sem := make(chan struct{}, maxContainers)
 	errCnt := int32(0)
 	var wg sync.WaitGroup
 	wg.Add(len(submissions))
+	// The rebuilds must complete even if the client disconnects, so detach
+	// them from the request's cancellation; each rebuild is bounded by the
+	// assignment's container timeout.
+	workerCtx := context.WithoutCancel(ctx)
```
JosteinLindhom, line 110: Should we also have this detached context for single-submission rebuilds (`internalRebuildSubmission`)?
Otherwise I think a client rebuilding a single submission will cancel the context if the client disconnects.

Maybe we should look into adding these to the detach interceptor somehow.

meling: Should be possible to just add `RebuildSubmissions` to this map:
```go
var detachedMethods = map[string]bool{
```
Maybe then we could avoid the `workerCtx` stuff herein.

PR #1569, opened 2026-08-15
logging: add the course-scoped log sink

meling:
Adds a `slog` handler that copies course-scoped log records, such as webhook processing, CI and Docker output, teacher-triggered rebuilds and assignment syncs, to per-course files teachers can read, without exposing the shared operator log. This is the storage layer only: no new API surface yet.

## What's new

**`internal/qlog`** — `WithCourse`/`WithCourseLog` scope a logger to a course, attaching `course_id`, `course_code`, and an opt-in `course_log` marker (the course's SCM organization name) that a later handler keys a course's log file on. `WithCourseLog` is the same minus `course_id`, for the RPC call paths where the interceptors already attached it from the caller's claims. `CourseAttrs` returns the same attributes for callers extending an existing logger rather than a context. `WithSink` fans a logger out to a second handler with `slog.NewMultiHandler`. `RelativeSource` is `New`'s source-path rewriter, pulled out for reuse.

**`internal/qlog/courselog`** — `Store` keeps one newline-delimited JSON file per course per UTC day under `<dir>/<organization>/<date>.jsonl` (directories `0750`, files `0600`). It's a single process-wide instance: a course's directory and file are created lazily on its first write, so a course created after the server started needs no restart. Retention is 14 days, swept at startup, at daily rollover, and by a ticker `Close` stops. Store failures go to a plain operator logger that never carries the marker, so reporting a failure can't recurse.

`Handler` copies a record only once a scope's `WithAttrs` call has carried the marker; attributes attached earlier (e.g. `rpc_method`) are held pending and replayed once it appears. Everything else is dropped. A message or attribute over 64 KiB is cut down to size and marked `truncated`.

**Wired in** — `main.go` builds the store at `$QUICKFEED/logs/courses` and fans the process logger through it; the store closes alongside the database and runner.

**Opted in** — `web/hooks/push.go`, `pull_request.go`, `installation.go` (after the course is persisted); `web/rebuild.go`'s single-submission and rebuild-all paths; `web/quickfeed_service.go`'s `UpdateAssignments`. `ci/run_tests.go`, `assignments/`, `courses_new.go`, and `sync.go` needed no changes — they always run under a `ctx` one of the above already scoped.

---

### DIFF
--- a/internal/env/path.go
+++ b/internal/env/path.go
@@ -106,3 +106,8 @@ func DatabasePath() string {
 func TestdataPath() string {
 	return Root("testdata", "courses")
 }
+
+// CourseLogDir returns the path to the directory holding per-course teacher logs.
+func CourseLogDir() string {
+	return Root("logs", "courses")
+}
--- a/internal/qlog/course.go
+++ b/internal/qlog/course.go
@@ -1,0 +1,49 @@
+package qlog
+
+import (
+	"context"
+	"log/slog"
+
+	"github.com/quickfeed/quickfeed/internal/qlog/label"
+)
+
+// Course is the subset of qf.Course that WithCourse and WithCourseLog need.
+// Declaring it here, rather than importing qf, keeps qlog free of a
+// dependency on the database model package.
+type Course interface {
+	GetID() uint64
+	GetCode() string
+	GetScmOrganizationName() string
+}
+
+// CourseAttrs returns the attributes WithCourse attaches, for the rare caller
+// that already holds an enriched logger it only wants to extend, rather than
+// a context to enrich.
+func CourseAttrs(course Course) []any {
+	return []any{
+		label.CourseID, course.GetID(),
+		label.CourseCode, course.GetCode(),
+		label.CourseLog, course.GetScmOrganizationName(),
+	}
+}
+
+// WithCourse returns a context and logger scoped to course, in addition to
+// attrs. Records logged through the returned logger are copied to the
+// course's teacher-visible log, so course must come from the database, never
+// from request data: nothing else can mark a scope for a course's log.
+func WithCourse(ctx context.Context, course Course, attrs ...any) (context.Context, *slog.Logger) {
+	return WithLogger(ctx, append(CourseAttrs(course), attrs...)...)
+}
+
+// WithCourseLog is WithCourse without CourseID, for the handful of RPC call
+// paths where the logging interceptors already attached CourseID to the
+// request logger from the caller's claims; see enrichRequestLogger. It still
+// attaches CourseCode, which is not part of that RPC scope, and the course
+// log marker.
+func WithCourseLog(ctx context.Context, course Course, attrs ...any) (context.Context, *slog.Logger) {
+	scoped := append([]any{
+		label.CourseCode, course.GetCode(),
+		label.CourseLog, course.GetScmOrganizationName(),
+	}, attrs...)
+	return WithLogger(ctx, scoped...)
+}
--- a/internal/qlog/courselog/handler.go
+++ b/internal/qlog/courselog/handler.go
@@ -1,0 +1,123 @@
+package courselog
+
+import (
+	"context"
+	"log/slog"
+
+	"github.com/quickfeed/quickfeed/internal/env"
+	"github.com/quickfeed/quickfeed/internal/qlog"
+	"github.com/quickfeed/quickfeed/internal/qlog/label"
+)
+
+// maxFieldBytes is the largest message or string attribute value kept in a
+// course log record; anything longer is cut down to size and the record is
+// marked label.Truncated.
+const maxFieldBytes = 64 * 1024
+
+// handler is a slog.Handler that copies records carrying label.CourseLog to a
+// Store, and drops every other record. Attach it alongside the operator's
+// handler with slog.NewMultiHandler; see qlog.WithSink.
+//
+// A handler is immutable: WithAttrs and WithGroup return a new value, as
+// slog.Handler requires. Until a scope's WithAttrs call carries
+// label.CourseLog, its attributes are held pending; the call that carries it
+// selects the course's file and replays the pending attributes into it.
+type handler struct {
+	store *Store
+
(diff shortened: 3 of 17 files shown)


meling (commented):
## Summary

This is a carefully designed storage-only course log sink: an opt-in `course_log` marker, lazy per-course JSONL files, path sanitization, 14-day retention, and a separate operator logger so store failures cannot recurse. The handler’s pending-attr replay, `WithCourse` vs `WithCourseLog`, and the call-site wiring (webhooks after the course is loaded, rebuild/UpdateAssignments under interceptor `course_id`) all match the stated intent, and the store/handler tests cover the important properties. The remaining issues are schema/duplication regressions on the operator stream, a process-wide MultiHandler cost, and a missing ignore/docs note for the new on-disk location — not correctness bugs in the sink itself.

## Issue counts by severity

- bugs: 0
- suggestions: 4
- nits: 1

internal/qlog/course.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// CourseAttrs returns the attributes WithCourse attaches, for the rare caller
+// that already holds an enriched logger it only wants to extend, rather than
+// a context to enrich.
+func CourseAttrs(course Course) []any {
+	return []any{
+		label.CourseID, course.GetID(),
+		label.CourseCode, course.GetCode(),
+		label.CourseLog, course.GetScmOrganizationName(),
```
meling, line 26: **[suggestion]** `CourseAttrs` / `WithCourse` attach `course_id`, `course_code`, and the internal `course_log` marker, but not `label.Organization`. `handlePush` and `UpdateAssignments` previously scoped the entire downstream tree (CI, Docker, assignment sync, clone) with `organization=<scm org>`. Those records now carry `course_log=` instead, so operator queries on the documented `organization` key miss the hottest course-scoped paths, and a routing field the handler comments call internal leaks onto stderr.

**Suggestion:** Keep attaching `label.Organization` (same SCM org name) in `CourseAttrs` / `WithCourseLog` alongside `CourseLog`. Leave `CourseLog` as the sink-only opt-in marker that `extractCourseLog` strips from teacher files.

web/hooks/installation.go
```diff
@@ -53,7 +53,9 @@ func (wh GitHubWebHook) handleInstallationCreated(ctx context.Context, event *gi
 		logger.Error("failed to create course", label.Error, err)
 		return
 	}
-	logger.Info("created course", label.CourseID, c.GetID())
+	// The course now exists, so its records can be copied to its log.
+	_, logger = qlog.WithCourse(ctx, c)
```
meling, line 60: **[suggestion]** The logger is already scoped with `course_code` (and `organization`) at line 40–44. `WithCourse` then appends `CourseAttrs`, which always includes `CourseCode` again. slog emits both, so the “created course” / token-refresh records have a duplicate `course_code` key in the operator text log and in the course JSONL (invalid-but-common duplicate JSON keys; last value wins). That violates the package rule in `label` that handlers must not repeat attributes.

**Suggestion:** After persist, add only what is missing (`CourseID` and `CourseLog`), or drop `CourseCode` from the earlier `WithLogger` and let `WithCourse` be the single course scope.

internal/qlog/courselog/handler.go
```diff
@@ -0,0 +1,123 @@
...
+	pending []slog.Attr
+}
+
+// NewHandler returns a Handler that writes through store.
+func NewHandler(store *Store) slog.Handler {
+	return &handler{store: store}
+}
+
+func (h *handler) Enabled(_ context.Context, level slog.Level) bool {
+	return level >= slog.LevelDebug
```
meling: **[suggestion]** `Enabled` is `level >= Debug` even when `inner == nil`. `slog.NewMultiHandler` ORs `Enabled` and, in `Handle`, clones the record for every handler that reports enabled. The process logger is fanned through this sink, so every unscoped debug+ line (most RPCs) is cloned and then dropped in `Handle`. After `WithCourse`, `inner` is set and the extra work is necessary; before that, it is not.

**Suggestion:** `return h.inner != nil && level >= slog.LevelDebug`. Unscoped loggers then skip this handler entirely; opted-in scopes are unchanged.

internal/env/path.go
```diff
@@ -101,3 +101,8 @@ func DatabasePath() string {
 func TestdataPath() string {
 	return Root("testdata", "courses")
 }
+
+// CourseLogDir returns the path to the directory holding per-course teacher logs.
+func CourseLogDir() string {
+	return Root("logs", "courses")
```
meling, line 112: **[suggestion]** The store is created at `$QUICKFEED/logs/courses/<org>/<date>.jsonl`. `.gitignore` only ignores `*.log` / `*.db`, and `doc/deploy-prod.md` only documents `qf.log` + logrotate. A local `quickfeed -dev` (or any run with `QUICKFEED` pointing at the repo) will create untracked `logs/` trees, and operators get no note that this tree is self-rotated at 14 days and can grow with CI output.

**Suggestion:** Ignore `logs/` (or `logs/courses/`) and add a short note next to the existing server-log section describing the path, permissions (0750/0600), and retention.

internal/qlog/courselog/store.go
```diff
@@ -0,0 +1,243 @@
...
+
+func (s *Store) write(org string, p []byte) (int, error) {
+	cf := s.courseFileFor(org)
+	cf.mu.Lock()
+	defer cf.mu.Unlock()
+
+	today := s.now().UTC().Format(dateLayout)
+	if cf.file == nil || cf.date != today {
+		if cf.file != nil {
+			_ = cf.file.Close()
```
meling, line 138: **[nit]** On date rollover, `write` closes the current file and only then calls `openFile`. If `openFile` fails, `cf.file` still points at the closed `*os.File` and `cf.date` is unchanged. The next write retries (and `Close`s an already-closed file, error ignored), so this is not a leak or a lost-handle bug, just a sloppy failure path.

**Suggestion:** Set `cf.file = nil` after a successful close, and only assign `cf.file`/`cf.date` after `openFile` succeeds.

PR #1571, opened 2026-08-15
web: add the GetCourseLog RPC

meling:
Adds `GetCourseLog`, a teacher-only RPC that queries a course's log (from `meling/course-log-sink`) by time interval, repository, and minimum level.

## What's new

**Proto** — `CourseLogRequest` (`courseID`, `from`/`to` defaulting to the last 24 hours, an exact `repository` match, a minimum `CourseLogEntry.Level`, and a `limit` defaulting to 2000) and `CourseLog` (chronological `entries`, every `repositories` value seen in the interval, and a `truncated` flag). `CourseLogEntry.Level` is nested the way `Enrollment.UserStatus` already is, and referenced from `CourseLogRequest` across messages the same way.

**Reader** — `Store.Query` in `internal/qlog/courselog` decodes a course's JSONL date files, filters, and keeps the newest `limit` matches in chronological order. The `repositories` list reflects everything logged in the time window regardless of the repository or level filter, so narrowing either one never shrinks a teacher's repository dropdown. A malformed final line (a partial write) is ignored; a malformed line elsewhere is a real failure.

**Handler** — `web/course_log.go` resolves the course, computes the interval and limit, and maps failures to `CodeNotFound` (unknown course) or `CodeInternal` (read failure). Rather than rejecting an interval past retention or a limit above the maximum, both are silently clamped — consistent with how an over-limit result is reported via `truncated` instead of an error.

**Access control** — `checkTeacher`, matching `GetSubmissionsByCourse`: a site admin who isn't a teacher of the course is denied.

`QuickFeedService` gains a `courseLogs *courselog.Store` field (`NewQuickFeedService`'s sixth argument) — the same process-wide store the sink writes through, not one per course.

## Also

- `make proto` picked up an unrelated `protoc-gen-es` version bump for the unpinned remote TS plugin, touching `score_pb.ts`/`types_pb.ts` with cosmetic diffs unrelated to this change.
- Four method tables needed updating for the new RPC: the access-control method list, `TestMethodCheckerRequestTypes`, `TestRPCLoggingNoDuplicateScope`'s handler list, and `TestImplementsValidation`, which the original plan hadn't anticipated — it enumerates every registered proto message.

---


web/course_log.go
```diff
@@ -0,0 +1,131 @@
...
+		Repositories: repositories,
+		Truncated:    truncated,
+	}, nil
+}
+
+// courseLogInterval applies in's defaults (the last 24 hours) and clamps From
+// to the retention window, rather than rejecting a request for an interval
+// wider than what the store can ever hold.
+func courseLogInterval(in *qf.CourseLogRequest) (from, to time.Time) {
+	to = time.Now()
```
meling: **[suggestion]** `courseLogInterval` clamps `From` to `now - Retention` but never clamps `To` to now (or to `From + Retention`). The handler comment on lines 26–27 says From/To/Limit are “all three” clamped; only From and Limit actually are. `IsValid` only rejects an inverted explicit interval, so a teacher can send `From` inside the retention window and `To` at protobuf’s far future (year 9999). `datesBetween` then allocates a string per UTC day and `os.Open`s each missing file — on the order of millions of iterations — before returning an empty result.

**Suggestion:** Clamp `To` to `time.Now()` (and, if `To` then precedes `From`, return an empty result). In `datesBetween`, also refuse to walk past today / beyond `Retention`, so `Store.Query` cannot be used as an unbounded loop even if a future caller skips the handler.

internal/qlog/courselog/query.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	defer f.Close()
+
+	var lines []string
+	scanner := bufio.NewScanner(f)
+	// A line may hold a message or attribute truncated up to 64 KiB, plus its
+	// surrounding JSON, so grow well past bufio.Scanner's 64 KiB default cap.
+	scanner.Buffer(make([]byte, 0, 64*1024), 256*1024)
+	for scanner.Scan() {
+		lines = append(lines, scanner.Text())
```
meling: **[suggestion]** `scanFile` reads every line into a `[]string` solely so it can treat a decode error on the last line as a partial write. The sink truncates a message or string attribute to 64 KiB, so a busy course’s daily file (CI/Docker output) can be tens to hundreds of megabytes. That allocation sits on top of the 5 000-entry ring and is repeated for each day in the interval. A single line over the 256 KiB scanner cap also fails the whole day’s query.

**Suggestion:** Stream the file: decode line N-1 as soon as line N arrives, and apply the partial-write exception only to a decode error on the final token. That keeps the last-line policy without holding the file in memory. Consider raising the scanner max or skipping an overlong line instead of failing the request.

web/course_log.go
```diff
@@ -0,0 +1,131 @@
...
+	if t := in.GetTo(); t != nil {
+		to = t.AsTime()
+	}
+	from = to.Add(-defaultCourseLogInterval)
+	if f := in.GetFrom(); f != nil {
+		from = f.AsTime()
+	}
+	if cutoff := time.Now().Add(-courselog.Retention); from.Before(cutoff) {
```
meling: **[suggestion]** The handler clamps `From` to `time.Now().Add(-courselog.Retention)`, but the store deletes a date file when that date’s UTC midnight is before the same cutoff (`cleanupCourseDir`: `t.Before(cutoff)` with `t` parsed as `00:00Z`). If now is 15:00 UTC, the file for (today − 14d) is already gone, including records from 15:00–23:59 that are still inside the 14-day window the handler just advertised. A “last 14 days” query therefore silently misses up to almost 24 hours at the old end, with no `truncated` signal.

**Suggestion:** Clamp `From` to the UTC midnight of the oldest date cleanup would still keep (the first date `d` where `d.00:00Z >= now-Retention`), or change cleanup (in the sink PR) to keep a day’s file until the entire day is past retention. Either way the handler window and on-disk files should agree.

internal/qlog/courselog/query.go
```diff
@@ -0,0 +1,254 @@
...
+		if err != nil {
+			if i == len(lines)-1 {
+				// The process may have been killed mid-write; the last line
+				// of the most recent file is the only one this can affect.
+				return nil
+			}
+			return err
+		}
+		if entry.Repository != "" && !entry.Time.Before(q.From) && !entry.Time.After(q.To) {
+			repos[entry.Repository] = true
```
meling: **[nit]** The comment says a partial write can only affect “the last line of the most recent file,” but the code (correctly) ignores a malformed last line of every date file. A crash on day N leaves a torn last line on that day’s file, which is no longer the most recent once the process writes day N+1.

**Suggestion:** Drop “of the most recent file” so the comment matches the code and the crash-at-rollover case.

meling (commented):
## Summary

This is a solid, well-tested addition of a teacher-only `GetCourseLog` RPC: proto types, `Store.Query`, handler defaults/clamps, `checkTeacher` access control, and MockClient wiring are consistent and match the stated design. Success, permission, validation, filter, truncation, and storage-error paths are covered. The main risks are robustness rather than logic holes: `To` is never capped, so `datesBetween` can walk an unbounded number of calendar days, and `scanFile` materializes each day's JSONL in memory before decoding.

## Issue counts by severity

- bugs: 0
- suggestions: 3
- nits: 1

JosteinLindhom (approved):

PR #1572, opened 2026-08-15
teacher: add the Course Logs page

### LINKED ISSUE #1131: ci: how to detect test environment failure?

meling:
Adds a teacher-facing "Course Logs" page that queries `GetCourseLog` (from `meling/get-course-log`) and lets a teacher narrow, search, copy, and download their course's log.

## What's new

**`LogOutput`** — the build-log card shell extracted out of `Lab.tsx`: a header with the terminal icon, a title, an optional right-hand `controls` slot, and the scrollable, monospaced body. The previously hardcoded `#f87171` inline style is now a `codeClassName` prop (default `text-error`, so `Lab`'s own look is unchanged), letting a caller with differently colored lines — level badges, in `CourseLogs` — override it. `Lab` passes its existing pre-rendered spans as children; `lab.test.tsx`, which queries by text, is untouched. The body is capped at `max-h-[70vh]` and scrolls within the card: a course log defaults to 2000 flex rows, and laying those out at full page height would push the page's own controls out of reach. This applies to `Lab`'s build log too, which is the one deliberate change to the submission page.

**`CourseLogs`** — a new page at `/course/:id/logs`, holding its filters and the fetched result in local `useState` and calling `useGrpc().global.api.client.getCourseLog(...)` from a `useEffect` — the pattern already used in `admin/User.tsx` and `AssignmentFeedbackView.tsx`; no new Overmind namespace.

- `datetime-local` From/To inputs (defaulting to the last 24 hours), a repository selector built from the response's `repositories`, and a minimum-level selector — all applied together by an explicit Refresh button.
- An unedited To means "up to now": it is left out of the request so the server bounds the interval by its own clock, and the field advances to match. A page left open therefore picks up whatever was logged since it loaded. A From that falls after To is reported and blocks the request, rather than being sent for the server to answer with an empty result that reads as "nothing happened".
- The effect refetches when the route names another course, since the router keeps the page mounted across that change, and clears the repository selection, which belongs to a single course. A selected repository that falls quiet stays in the option list, so the select never sits blank while still filtering.
- Each row shows the time, a level badge, the repository, the message, the entry's structured `fields`, and its `source`. Almost all useful context — CI and Docker `output`, `error`, assignment, commit, submission id, score, exit status — lives in `fields`, so a row that omitted them showed nothing the teacher docs promise. Fields are sorted by key, since a protobuf map has no order of its own.
- A free-text box, built by reusing `Search`'s existing local `setQuery` option, filters the already-fetched entries live and never triggers a request.
- Copy (`navigator.clipboard`) and Download (a `Blob` + a synthetic `<a download>` click) act on whatever is currently filtered, through the same line formatter the rows are built from, so the screen, the copied text, the downloaded file, and the search all agree. Both report failure to the teacher: `navigator.clipboard` is undefined outside a secure context, and `writeText` rejects when access is denied.
- Loading, error, empty, and truncated states, built from the existing `CenteredMessage` component and a daisyUI `alert`. The truncation banner reports the server's own entry count and stands whenever the response is truncated, including when the local filter leaves nothing on screen — it is a server-side cutoff, which the free-text box neither causes nor undoes.

**Wired in** — a "Course Logs" tile on `TeacherPage` and a matching `NavBarTeacher` sidebar link.

## Also

- **Deviation from the plan:** entries display via `timestampDate(...).toLocaleString()` rather than the existing `getFormattedTime` helper, which has no seconds and would visually collapse several log lines from the same minute together — common for webhook/CI bursts.
- Rows are `<span>`s, not `<div>`s: `LogOutput` renders them inside a `<code>` element, which only admits phrasing content.
- `npx tsc --noEmit` is clean apart from a pre-existing error in `src/__tests__/TestHelpers.ts`, and the Vitest suite passes with 95 tests (up from 88 on `master`).
- `npm run lint` still cannot start, and the cause is now pinned down: `public/eslint.config.js` uses `reactHooksPlugin.configs["recommended-latest"]`, but in eslint-plugin-react-hooks 7.1.1 that preset is legacy eslintrc-shaped; the flat-config one is `configs.flat["recommended-latest"]`. With that one-line change ESLint runs and reports 18 errors, 17 of them pre-existing in 13 other files under rules new in v7 (mostly `react-hooks/static-components`). Fixed separately, stacked on top of this PR, rather than turning CI red here.

---


public/src/components/teacher/CourseLogs.tsx
```diff
@@ -0,0 +1,209 @@
...
+                                <span className="text-base-content/60 shrink-0">
+                                    {entry.time ? timestampDate(entry.time).toLocaleString() : "N/A"}
+                                </span>
+                                <span className={`badge badge-xs ${LEVEL_BADGE_COLOR[entry.level]} shrink-0`}>
+                                    {LEVEL_NAMES[entry.level]}
+                                </span>
+                                {entry.repository && (
+                                    <span className="text-base-content/60 shrink-0">[{entry.repository}]</span>
+                                )}
+                                <span className="break-words">{entry.message}</span>
```
meling, line 265: **[bug]** On-screen rows render only time, level, repository, message, and the per-entry truncated badge. They never show `entry.fields` (or `entry.source`). Almost all useful course-log context lives in `fields`: CI/Docker `output`, `error`, assignment name, commit, submission id, path, score, exit status. `entryText` already formats those fields for copy, download, and the local search filter, and its comment claims it is shared with on-screen rendering — but the visual line does not use it. The teacher docs say this page shows CI and Docker output; that output is invisible unless the teacher copies or downloads. Local search can also match field values that never appear on screen.

**Suggestion:** Render `entry.fields` (and, if useful, `source`) on each row, or render `entryText(entry)` as the line body so the screen, copy, download, and search stay in sync. Add a test entry with `fields` / a multi-line `output` value and assert it is visible.

public/src/components/teacher/CourseLogs.tsx
```diff
@@ -0,0 +1,209 @@
...
+            {!error && loading && <CenteredMessage message="Loading course logs…" />}
+            {!error && !loading && result && filtered.length === 0 && (
+                <CenteredMessage message="No log entries match the current filters" />
+            )}
+            {!error && !loading && result && filtered.length > 0 && (
+                <>
+                    {result.truncated && (
+                        <div className="alert alert-warning">
+                            <span>
+                                Showing the newest {filtered.length} matching {filtered.length === 1 ? "entry" : "entries"}.
```
meling: **[bug]** The truncation banner interpolates `filtered.length`. `result.truncated` means the server cut the unfiltered (server-filtered) result at its limit, not that the local search produced that many lines. After the teacher types in "Filter loaded entries", the banner can say "Showing the newest 2 matching entries" when the server actually returned 2000 and dropped the rest. The warning is also omitted entirely when the local filter reduces the view to zero, even though the underlying result is still truncated.

**Suggestion:** Use `entries.length` (the server page size) in the banner, keep showing it whenever `result.truncated` is set, and phrase it as a server-side cutoff (e.g. "Result limited to the newest N entries; narrow the interval or filters and Refresh").

public/src/components/teacher/CourseLogs.tsx
```diff
@@ -0,0 +1,209 @@
...
+ *  the result by interval, repository, and minimum level, then locally
+ *  filter, copy, or download whatever was loaded. Filters other than the
+ *  free-text one only take effect on Refresh; there is no polling or
+ *  streaming in this version. */
+const CourseLogs = () => {
+    const courseID = useCourseID()
+    const { api } = useGrpc().global
+
+    const [from, setFrom] = useState(() => toLocalDatetimeInput(new Date(Date.now() - 24 * 60 * 60 * 1000)))
+    const [to, setTo] = useState(() => toLocalDatetimeInput(new Date()))
```
meling, line 64: **[suggestion]** `to` is snapshotted to "now" once, on mount, and Refresh always sends that same timestamp. Leaving the page open and clicking Refresh therefore cannot pick up anything logged after load unless the teacher notices and edits To. That is easy to miss because To is pre-filled and the button is labeled Refresh.

**Suggestion:** Treat an untouched/empty To as "now" (omit `to` and let the server default), or set To to the current local time at the start of each `fetchLogs` when the teacher has not customized the range.

public/src/components/teacher/CourseLogs.tsx
```diff
@@ -0,0 +1,209 @@
...
+        }
+        setResult(response.message)
+    }
+
+    useEffect(() => {
+        void fetchLogs()
+        // Fetch once, with the default filters, when the page mounts; later
+        // changes are applied only when the teacher clicks Refresh.
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [])
```
meling: **[suggestion]** The mount effect disables `react-hooks/exhaustive-deps` and never lists `courseID`. React Router keeps `CourseLogs` mounted when only `:id` changes (`/course/1/logs` → `/course/2/logs`), so the page would keep showing the previous course's entries until Refresh. The sidebar currently navigates to the course root first, so this is not the common click path, but `AssignmentFeedbackView` already depends on `activeCourse` for the same reason.

**Suggestion:** Depend on `courseID` (keep filters out of the effect so Refresh stays explicit), or remount with `key={courseID}` on the `/logs` route.

public/src/components/LogOutput.tsx
```diff
@@ -0,0 +1,40 @@
...
+        <div className="card bg-base-200 shadow-xl rounded-2xl overflow-hidden">
+            <div className="card-body p-0">
+                <div className="flex items-center justify-between bg-base-300 px-4 py-3 border-b border-base-content/10">
+                    <h3 className="text-sm font-semibold flex items-center gap-2">
+                        <i className="fas fa-terminal" />
+                        <span>{title}</span>
+                    </h3>
+                    {controls}
+                </div>
+                <div className="overflow-x-auto">
```
meling: **[suggestion]** The log body is only `overflow-x-auto` with no max height. `GetCourseLog` defaults to 2000 entries, and each course-log row is a flex block with badges. A busy course's last 24 hours will produce a very long, expensive page; Lab's build log is typically much shorter, so this is a new cost of reuse.

**Suggestion:** Cap the body (e.g. `max-h-[70vh] overflow-auto`) so Refresh/filters stay reachable and the browser does not lay out thousands of off-screen rows at full page height.

public/src/components/teacher/CourseLogs.tsx
```diff
@@ -0,0 +1,209 @@
...
+// entryText renders one entry as a single line of plain text, shared by the
+// on-screen rendering, copy, and download actions.
+const entryText = (entry: CourseLogEntry): string => {
+    const time = entry.time ? timestampDate(entry.time).toLocaleString() : ""
+    const fields = Object.entries(entry.fields).map(([key, value]) => `${key}=${value}`).join(" ")
+    const repository = entry.repository ? `[${entry.repository}]` : ""
+    return [time, LEVEL_NAMES[entry.level], repository, entry.message, fields].filter(Boolean).join(" ")
+}
+
+/** CourseLogs is the teacher-only "Course Logs" page at /course/:id/logs.
```
meling, line 54: **[suggestion]** Several new comments restate the code or embed design history: this JSDoc narrates "there is no polling or streaming in this version"; `entryText`'s comment claims it is used for on-screen rendering (it is not); and `LogOutput.tsx` documents every prop, including "matching the original submission page usage".

**Suggestion:** Keep only the non-obvious why (e.g. why `toLocalDatetimeInput` cannot use `toISOString`, and why the effect is mount-only). Drop the version narrative and the inaccurate `entryText` claim.

meling (commented):
## Summary

This PR adds a teacher-only Course Logs page that queries `GetCourseLog`, extracts a shared `LogOutput` shell from `Lab.tsx`, and wires the tile, sidebar, route, tests, and docs. Routing, teacher gating (CoursePage + backend `checkTeacher`), React-escaped log text, and the Lab extract look correct, and the tests cover mount, error, empty, truncated, local search, and tile navigation. The page as rendered is still missing the structured payload teachers need (CI/Docker `output`, `error`, assignment, commit), and the truncation banner reports the locally filtered count rather than the server result.

## Issue counts by severity

- bugs: 2
- suggestions: 4
- nits: 0

public/src/components/teacher/CourseLogs.tsx
meling: Fixed in 85eb371f. Rows now render the entry's `fields` and `source` alongside the message, and both the row and `entryText` are built from the same parts (`entryTime`, `entryFields`, `source`), so the screen, the clipboard, the downloaded file, and the free-text filter cannot drift apart. Fields are sorted by key, since a protobuf map has no order of its own and they were otherwise moving between requests. A test asserts a multi-line `output=--- FAIL: TestFoo…` and the `source` are on screen.

Rows also became `<span>`s in the same commit: `LogOutput` renders them inside a `<code>` element, which only admits phrasing content.

public/src/components/teacher/CourseLogs.tsx
meling: Fixed in ceec424c. The banner now reports `entries.length` — the server's own count — is phrased as the server-side cutoff ("Result limited to the newest N entries. Narrow the interval or the filters and click Refresh to see the rest."), and moved out of the non-empty branch so it stands whenever `result.truncated` is set, including when the local filter matches nothing. The test asserts both: the count stays 3 after filtering, and the banner survives a filter that empties the view.

public/src/components/teacher/CourseLogs.tsx
meling: Fixed in 3fa620c2. An unedited To is now left out of the request entirely, so the server bounds the interval by its own clock — this also avoids the minute truncation a `datetime-local` value would impose — and the field is advanced to the current local time so it is not lying about what was queried. Editing To sets a flag and pins it. Tested by asserting `to` is undefined on the mount request and defined after the field is edited.

public/src/components/teacher/CourseLogs.tsx
meling: Fixed in 1c20f07b: the effect depends on `courseID`. Confirmed to be a real regression test — reverting the dependency back to `[]` fails with `expected [ { courseID: 1n } ] to have a length of 2`.

Follow-up in 61335324: the repository selection is cleared on that same transition, since a repository belongs to one course and was otherwise filtering the new course's log by the old course's repository. `fetchLogs` takes the repository explicitly so the clear and the query happen in one pass.

public/src/components/LogOutput.tsx
meling: Fixed in 26ac10c6: the body is `max-h-[70vh] overflow-auto`, matching the `max-h-[90vh] overflow-y-auto` idiom already used in `SubmissionModal`. Worth noting this applies to `Lab`'s build log too — the one deliberate change to the submission page, and an improvement there as well.

public/src/components/teacher/CourseLogs.tsx
meling: Fixed. The per-prop JSDoc in `LogOutput` is gone (26ac10c6), the "no polling or streaming in this version" narrative is gone (14e99fad), and `entryText`'s comment is now accurate: after 85eb371f it genuinely does share its parts with the rendered row, and the comment says why that matters rather than just restating it. The remaining comments are the non-obvious why you called out — why `toLocalDatetimeInput` cannot use `toISOString`, why the effect lists only `courseID`, why an unedited To is omitted, and why a row is a `<span>`.

doc/teacher.md
```diff
@@ -99,6 +99,16 @@ A member of the teaching staff can then copy course `info`, the code template fo
...
+
+By default, the page shows the last 24 hours, ending whenever you last clicked Refresh; leave To alone and each Refresh brings the view up to the present. Use the From/To fields to look further back, up to 14 days: QuickFeed only keeps that much history, so an interval reaching further back is narrowed automatically rather than rejected. A From that falls after To is rejected, since no entry can lie in such an interval. The Repository and Minimum level selectors narrow the result further; the repository list always reflects every repository active in the selected interval, so switching one filter never removes an option from the other, and a repository you have selected stays listed even once it falls quiet. Changing any of these takes effect the next time you click Refresh; the free-text box below it, by contrast, filters what is already on screen immediately, with no additional request. Copy and Download act on whatever is currently on screen, filters included.
```
JosteinLindhom, line 106: This is very dense and heavy on implementation details

JosteinLindhom (approved):

PR #1586, opened 2026-08-23
frontend: restore ESLint and fix what it reports

meling:
`npm run lint` has not been linting anything: it fails while loading its own configuration, with *"A config object has a `plugins` key defined as an array of strings"*. This restores it and fixes everything it then reports.

## The config

eslint-plugin-react-hooks ships every preset twice. The top-level `configs.recommended` and `configs["recommended-latest"]` are the legacy eslintrc shape, whose `plugins` is an array of strings — which flat config rejects outright. The flat equivalents live under `configs.flat`, so `public/eslint.config.js` now uses `reactHooksPlugin.configs.flat["recommended-latest"]`.

With that, ESLint runs and reports 18 errors, all under rules new in react-hooks v7 (the React Compiler set). One commit per rule:

## `react-hooks/static-components` (9)

`CourseCard`, `Lab`, `ManageSubmissionStatus`, `GroupForm`, and `ReviewInfo` each declared a component inside another component's body. That makes a new component *type* on every render, so React unmounts and remounts the whole subtree rather than updating it. All but one are render helpers, not components, and are now called as functions. `InfoRow` is a real component driven only by its props, so it moves to module scope.

## `react-hooks/immutability` (2)

`CourseForm.handleChange` and `ProfileForm.handleChange` mutated the object held in state and then called the setter with that same object. React compares by identity, so the update was dropped — the form's copy and the value a later render reads could disagree. Only the uncontrolled `defaultValue` inputs kept that from being visible. Both now clone, edit the clone, and set that.

## `react-hooks/use-memo` (1)

`SubmissionScores` passed a named function to `useMemo`, so the memo was keyed on a dependency list that never mentioned the function it called. The sort is inlined; `submission`, `sortKey`, and `sortAscending` are now the complete list.

## `react-hooks/set-state-in-effect` (4)

Three are derived state mirrored into `useState` by an effect, which only buys a second render after every change: `Breadcrumbs`' course and assignment names (whose effect additionally ran on *every* render, since `pathnames` is a fresh array each time), `ToggleSwitch`'s teacher/student label, and `Profile`'s edit mode, which is really "forced whenever the user object is incomplete". All three are computed during render now, and three `useState`/`useEffect` pairs go away.

The fourth is `CourseLogs`' fetch effect, which is what an effect is actually for; reporting that a request is in flight is unavoidably a state update, and escaping the rule would take a data-fetching layer this page does not have. It carries a scoped disable and the reason.

## Verification

`npx eslint .` clean, `npx vitest run` 95 passed, `npx tsc --noEmit` clean apart from the pre-existing `src/__tests__/TestHelpers.ts` error. No CI workflow runs the frontend lint today, so this was only reachable locally and through DeepSource.

### DIFF
--- a/public/eslint.config.js
+++ b/public/eslint.config.js
@@ -23,9 +23,11 @@ export default defineConfig([
     js.configs.recommended,
     tsPlugin.configs["flat/eslint-recommended"],
     reactPlugin.configs.flat.recommended,
-    // react-hooks ships two presets: `recommended` is legacy (eslintrc) and
-    // `recommended-latest` is the flat-config equivalent.
-    reactHooksPlugin.configs["recommended-latest"],
+    // react-hooks ships each preset twice: `configs.recommended` and
+    // `configs["recommended-latest"]` are legacy (eslintrc), whose string
+    // `plugins` array flat config rejects outright; `configs.flat` holds the
+    // flat-config equivalents.
+    reactHooksPlugin.configs.flat["recommended-latest"],
 
     // TypeScript wiring and project-specific overrides.
     {
--- a/public/src/components/CourseCard.tsx
+++ b/public/src/components/CourseCard.tsx
@@ -29,7 +29,7 @@ const CourseCard = ({ course, enrollment, unavailable }: CardProps) => {
     const state = useAppState()
 
     const handleEnroll = useCallback(() => actions.enroll(course.ID), [actions, course.ID])
-    const CourseEnrollmentButton = () => {
+    const courseEnrollmentButton = () => {
         // Always show 'Go to Course' if enrolled or pending, even if unavailable
         if (hasNone(status)) {
             // Show enroll if not unavailable, or if user is admin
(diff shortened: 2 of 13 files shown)


meling:
Superseded by #1563, which does the same work coherently.

This branch changed `eslint.config.js` to `reactHooksPlugin.configs.flat["recommended-latest"]` and fixed the 18 errors that surfaced — but it never bumped the dependency. Its lockfile still pins `eslint-plugin-react-hooks@5.2.0`, where there is no `configs.flat` namespace at all and `configs["recommended-latest"]` *is* the flat config. On a clean `npm ci` this branch fails with `TypeError: Cannot read properties of undefined (reading 'recommended-latest')`; it only appeared to work because the local `node_modules` had been left at 7.1.1 from working on #1563. At the declared 5.2.0 the original config lints clean and none of those 18 errors exist, since the React Compiler rules ship in v6/v7.

#1563 bumps `eslint-plugin-react-hooks` to ^7.1.1 — which is what makes the config change and the fixes necessary — applies the same fixes (`ReviewInfo.tsx` is byte-identical; the rest differ only in helper naming, comment wording, and import order), and additionally fixes the `TestHelpers.ts` `tsc` error. It is also better than this branch in two places: it hoists Breadcrumbs' pure resolvers to module scope, and it uses React's documented adjust-state-during-render pattern in Profile rather than changing the edit-mode semantics.

The one thing here that #1563 could not have is the `set-state-in-effect` exemption on the Course Logs fetch effect, since `CourseLogs.tsx` does not exist on master. That moves into #1563 as part of rebasing it onto the stack.

Issue #1596, opened 2026-08-26
epic: detect and surface test environment failures


meling:
Umbrella issue for the work that remains after #1131 is closed.

### What #1131 asked

How does a teacher find out that a test run failed because the *test environment* is wrong — a missing generated file, a compile error, a broken `run.sh` — rather than because the student's code is wrong? Until recently the answer was "read the server log over SSH, if you have access".

### What has landed

The course log stack gives teachers the raw material:

- #1566 — the backend logs through structured, context-aware `slog`
- #1569 — a course-scoped log sink writing one JSONL file per course per day
- #1571 — the `GetCourseLog` RPC, teacher-only, filtered by interval, repository, and level
- #1572 — the `Course Logs` page, with filters, search, copy, and download

A teacher can now see what the server did with their push, without shell access. That is the whole of #1131's simplest-possible resolution, so #1131 is closed and the rest is tracked here.

The robustness work below has since also landed:

- #1439 — removing an entry from `tests.json` or `criteria.json` now removes the corresponding database rows, instead of leaving them stale
- #1408 — the tests repository is validated on push; every problem is collected instead of stopping at the first, and issues reach the teacher through the course log and the frontend
- #1593 — a failed test run (build failure, timeout, no parsable scores, panic) is now recorded with an explicit failure status instead of silently overwriting the previous score with zero
- #1604 — course templates and fixtures now use the single-phase test setup instead of the obsolete `SCORE_INIT` pattern

### What remains

Items marked 🧭 need a planning step before implementation.

**Make the log usable as it happens**

- 🧭 #1592 — stream course log entries to the teacher log view, instead of requiring Refresh
- 🧭 #1591 — log view polish: 24-hour clock, ISO dates, full-height display, column layout

**Catch environment failures before and during a run**

- 🧭 #1594 — canary run against the assignments repository, to catch a broken environment before any student submission does
- #1595 — validate that the tests and assignments repositories are aligned (matching assignment folders, required configuration files present in each)

**Tooling to catch problems locally, before a push**

- 🧭 #1605 — revive and harden the local `qcm` test runner

**Multi-language support**

- 🧭 #1606 — decide where per-language score emitters should live, and what C#/Python support requires (courses use both today; a Java course ran previously)

### Deliberately out of scope

Notification, in any form beyond the log page itself, is tracked in #1197, which already lists "fault with test environment" among the teacher cases. Streaming in general is #1023.

PR #1602, opened 2026-08-29
fix: reconcile removed tests.json entries and validate the tests repository

### LINKED ISSUE #1408: ui: show misconfigured tests repository in frontend
We already detect when processing the `tests` repository whether or not there are any problems, such as malformed json files or missing json files.

Here are some possible rules:
- Always required for each lab assignment: `assignment.json` (top-level folder only)
- Should we require one of: `criteria.json` or `tests.json`?

If a json file is malformed, then we should alert the teaching staff in their frontend.

Similar, if the above rules are not satisfied. Maybe we need to have an exclusion mechanism to avoid unnecessary alerts for cases like the `internal` folder that does not contain an assignment, but extra library functions etc.

### LINKED ISSUE #1439: bug: removing an entry in `tests.json` does not remove it from the database
I added a test case to my `tests.json` by accident. 
However, when removing it from `tests.json` it is not removed from the database.

This causes students to get 0 score on a test that does not exist, lowering their overall score.
This can be fixed by manually removing the entry from the `test_infos` table in the database.

We should ensure that tests removed from `tests.json` are also removed from the `test_infos` table to prevent this.

meling:
Two related fixes for the tests-repository sync path, part of the robustness work under #1596.

## fix: reconcile removed tests.json and criteria.json entries in the database (#1439)

Removing an entry from `tests.json` (or the whole file) left stale rows in
`test_infos`, zero-scoring tests that no longer exist. The same guard bug
left stale grading benchmarks behind when `criteria.json` was removed.

Always reconcile the database to the incoming (possibly empty) lists.

Fixes #1439

## feat: collect and surface tests repository issues (#1408)

Previously `readTestsRepositoryContent` aborted on the first content problem,
skipping the entire course update, and `UpdateFromTestsRepo` swallowed every
error - a teacher pushing a broken json file got a green "Assignments
updated" alert. Which problem was reported first was also nondeterministic
due to map iteration.

Now every content problem becomes a `RepoIssue`: files are visited in sorted
order, an assignment with an unparsable json file is excluded from the
database update (so a typo is not interpreted as removal of its tests or
criteria), and the remaining assignments are still updated.

New validation mirrors the runtime rules of `kit/score.Score.isValid`:
`tests.json` entries need a unique non-empty TestName, MaxScore > 0 and
Weight > 0; invalid entries are dropped with an issue. Empty benchmark
headings and criterion descriptions are reported. An auto-graded assignment
without expected tests gets a warning, since all its submissions would
silently score zero.

Folder rules: build context files are collected from the repository root
and the scripts folder only (a nested go.mod no longer leaks into the build
context); assignment files live at depth one; deeper files (internal
packages, testdata) are ignored, serving as the exclusion mechanism for
non-assignment folders.

Issues are logged at WARN with the course scope, so the webhook path
reaches the teacher through the course log page. The `UpdateAssignments`
RPC now returns `TestsRepositoryIssues` and no longer reports success when
the update failed; the frontend shows a yellow alert listing the issues.

Fixes #1408

## Verification
- `go test ./assignments/... ./database/... ./web/...`
- `make proto` regenerated cleanly
- `cd public && npm test` (104 tests) and `npx tsc --noEmit`
- `golangci-lint run ./...` clean


public/src/overmind/namespaces/global/actions.ts
```diff
@@ -647,6 +647,15 @@ export const updateAssignments = async ({ actions, effects }: Context, courseID:
     if (response.error) {
         return
     }
+    const issues = response.message.issues
```
meling: Is this it? I don't know if it is worth adding the `TestsRepositoryIssues` to the `UpdateAssignments` RPC if we are only using it for counting, and pointing to the course log. It seems more appropriate to just return a count and point to the course log, since we already have the `GetCourseLog` RPC.

qf/quickfeed.proto
```diff
@@ -32,7 +32,7 @@ service QuickFeedService {
     // assignments //
 
     rpc GetAssignments(CourseRequest) returns (Assignments) {}
-    rpc UpdateAssignments(CourseRequest) returns (Void) {}
+    rpc UpdateAssignments(CourseRequest) returns (TestsRepositoryIssues) {}
```
meling: Just return the count instead?

meling (commented):
Now, this looks fine to me.

PR #1603, opened 2026-08-29
feat: record failed test runs instead of zero scores

### LINKED ISSUE #1593: ci: report a failed test run instead of recording a zero score
Extracted from #1131, which asked how to detect that the test environment, rather than the student's code, is what failed.

When a container fails, the student and the teacher both see a zero score, which is indistinguishable from code that compiles and fails every test. `ci/docker.go` logs `container exited` with `exit_status`, and with the course log stack that record now reaches the teacher, but nothing downstream acts on it: the run is recorded as an ordinary scored submission.

### Detect and distinguish

- a non-zero container exit status
- an image build failure
- a run that produced no parsable score output at all
- a timeout — the assignment's `ContainerTimeout` reached
- a panic in the test code

### Report

Record such a run as failed rather than as a scored submission, so that:

- `RecordResults` does not overwrite a previous good score with a zero
- the teacher sees the cause, in the course log and ideally on the results view
- the build log the student sees says the run failed for an environment reason, instead of showing an empty result that reads as "you scored nothing"

### Related

- #539 — rebuild should update the submission status
- #1197 — notifying the teacher of a test environment fault
- #1408 — the same class of problem, caught earlier, at push time

meling:
Part of the robustness work under #1596. Stacked on #1602.

When the test environment failed - an image that would not start, a run
that produced no parsable score output, a timeout, or a panic before any
test results - the run was recorded as an ordinary zero-score submission,
indistinguishable from code that compiles and fails every test. A run that
failed before producing output recorded nothing at all, so the student saw
no change and only the course log knew.

`BuildInfo` now carries a `RunStatus` enum (SUCCESS, BUILD_FAILURE, TIMEOUT,
NO_SCORES, TEST_PANIC). A failed run is recorded with its failure status
and a student-facing explanation in the build log, while the previous
submission's score, scores, grades, and submission date are kept; slip
days are not consumed. The commit hash names the failed commit so that a
rebuild retries it.

A non-zero container exit status alone is not a failure - a course's
run.sh may end with the test command, which exits non-zero on failing
tests. The reliable signal is zero parsed score lines, since even
non-compiling submissions emit the zero-score initialization lines.
The container's exit status is now surfaced via `ContainerExitError`, and
stderr - previously discarded entirely - is included in the output when
the run fails or produces no stdout, so environment failures during
setup commands (`go get`, `go mod tidy`) are finally visible.

Also registers the two failure metrics counters that were incremented but
never exported, removes the unused `Score.weightedScore` left behind by the
task-scoring removal, and shows a failure banner in the submission view.

Fixes #1593

## Verification
- `go test ./ci/... ./kit/...` and `DOCKER_TESTS=1 go test ./ci/...` against the real Docker daemon (includes a new non-zero-exit container test)
- `make proto` regenerated cleanly
- `cd public && npm test` (104 tests) and `npx tsc --noEmit`
- `golangci-lint run ./...` clean in both modules


kit/score/results.go
```diff
@@ -145,6 +147,7 @@ func ExtractResults(out, secret string, execTime time.Duration, zeroScoreTests [
 				errs = append(errs, fmt.Errorf("parsing line '%s': %w", line, err))
 				continue
 			}
+			parsedScores++
```
JosteinLindhom: Does parsedScores > 0 imply success even if the check below fail (score not in expected tests) for all parsed scores? ref: classifyRun in `run_status.go`

JosteinLindhom: I was thinking we could use `len(results.Scores) > 0` instead of `results.ParsedScores > 0`, but I see now that the scores is prepopulated with expected tests (= len will generally be > 0).

PR #1620, opened 2026-09-11
feat(frontend): add Edit Course tile to teacher page

### LINKED ISSUE #1133: ui: add edit course button also to the course page for teachers
Teachers sometimes want to edit the course info such as slip days. We should make the `Edit Course` tile / button available from the course's page as well. Right now it is only available from the admin menu item.

### LINKED ISSUE #1579: Edit course should be available on the teacher's page too
To edit course name, course code, slip days etc requires navigation through the Admin menu from the profile image. This is awkward. We should make a separate button/tile for "Edit Course" on the teacher's main page for the course.

meling:
Closes #1579

## What changed

- **Edit Course tile on the teacher page** — `TeacherPage` gets an "Edit Course" card next to "Update Course Assignments", routed to `/course/:id/edit`. Editing the course name, code, tag, year, and slip days no longer requires going through the profile menu to Admin → Edit Course, which non-admin teachers could not reach at all.
- **Teacher-side `EditCourse` component** — `components/teacher/EditCourse.tsx` resolves the active course from the route and hands it to the existing `CourseForm`, which already navigates back to the course page after saving. A course missing from state renders a short "Course not found." notice.
- **No backend change** — `UpdateCourse` was already permitted for teachers (`web/interceptor/access_control.go`). The admin page is untouched, since it still serves editing any course in the system.

## Testing

New `editCourse.test.tsx` mounts `TeacherPage` under `/course/:id/*` and checks that the tile is present, that clicking it opens the form, that the form is prefilled with the course's values, and that an unknown course id reports "Course not found.".

`npm run lint` and `npm run test` pass (21 files, 142 tests). `npx tsc --noEmit` reports one pre-existing error in `updateAssignments.test.tsx` that is unrelated to this change and also present on master.

### DIFF
--- a/public/src/__tests__/editCourse.test.tsx
+++ b/public/src/__tests__/editCourse.test.tsx
@@ -1,0 +1,68 @@
+import { create } from "@bufbuild/protobuf"
+import { fireEvent, render, screen } from "@testing-library/react"
+import { Provider } from "overmind-react"
+import { MemoryRouter, Route, Routes } from "react-router"
+import { UserSchema } from "../../proto/qf/types_pb"
+import TeacherPage from "../pages/TeacherPage"
+import { MockData } from "./mock_data/mockData"
+import { initializeOvermind } from "./TestHelpers"
+
+// Course 1 of the mocked courses; the teacher below is enrolled in it as a teacher.
+const course = MockData.mockedCourses()[0]
+
+const teacherState = {
+    self: create(UserSchema, { ID: BigInt(1), Name: "Teacher", IsAdmin: false }),
+    activeCourse: course.ID,
+    courses: MockData.mockedCourses(),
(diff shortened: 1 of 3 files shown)


JosteinLindhom (approved):

JosteinLindhom:
Would maybe rename one of the now two components now named `EditCourse`.
The one in `/admin` could be renamed `CourseList` instead, as that's that is really does.

meling:
> Would maybe rename one of the now two components now named `EditCourse`. The one in `/admin` could be renamed `CourseList` instead, as that's that is really does.

I didn't quite understand why `CourseList` is a better name for it. Could you expand on why? I thought this was about editing the course details such as its course code, year, slip days etc.

